PR #5, opened 2023-09-14
Main reconcile logic and conditions

jc-rh:
More unit test for stage transitions will be added in subsequent PRs.


controllers/ibu_controller.go
```diff
@@ -121,16 +123,105 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	} else if reconcileTime == utils.StopReconciling {
 		return
 	}
+
+	if imageBasedUpgrade.Spec.Stage == ranv1alpha1.Stages.Idle {
+		return
+	}
+
+	ibus := &ranv1alpha1.ImageBasedUpgradeList{}
```
imiller0, line 131: Would it make sense for this CR to be a cluster scoped CR with a fixed name? I think bad things will happen if we ever have more than 1 agent CR.

controllers/ibu_controller.go
```diff
@@ -121,16 +123,105 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	} else if reconcileTime == utils.StopReconciling {
 		return
 	}
+
+	if imageBasedUpgrade.Spec.Stage == ranv1alpha1.Stages.Idle {
+		return
```
donpenney, line 128: Wouldn't we potentially have work to do if the stage gets set to Idle?

controllers/ibu_controller.go
```diff
@@ -121,16 +123,105 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			nextReconcile = requeueWithLongInterval()
+			err = r.updateStatus(ctx, imageBasedUpgrade)
+			return
+		}
+	}
+
+	previousSucceededCondition := utils.GetPreviousSucceededCondition(imageBasedUpgrade)
+	previousPendingCondition := utils.GetPreviousPendingCondition(imageBasedUpgrade)
+
+	if imageBasedUpgrade.Spec.Stage == ranv1alpha1.Stages.Rollback {
```
donpenney, line 149: I guess this block is doing a "state transition validation" check? Probably could be in its own function

controllers/ibu_controller.go
```diff
@@ -98,8 +100,8 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	nextReconcile = doNotRequeue()
 
-	ibu := &ranv1alpha1.ImageBasedUpgrade{}
-	err = r.Get(ctx, req.NamespacedName, ibu)
+	imageBasedUpgrade := &ranv1alpha1.ImageBasedUpgrade{}
```
donpenney, line 103: The "ibu" abbreviation is still used in other functions, so it seems inconsistent to change here. Is this change just because "ibu" is now also being used in a for-loop?

controllers/utils/conditions.go
```diff
@@ -0,0 +1,152 @@
...
+}{
+	Completed:  "Completed",
+	Failed:     "Failed",
+	TimedOut:   "TimedOut",
+	InProgress: "InProgress",
+	Blocked:    "Blocked",
+}
+
+// SetStatusCondition is a convenience wrapper for meta.SetStatusCondition that takes in the types defined here and converts them to strings
+func SetStatusCondition(existingConditions *[]metav1.Condition, conditionType ConditionType, conditionReason ConditionReason, conditionStatus metav1.ConditionStatus, message string) {
```
imiller0, line 56: I don't think we currently have any race conditions because there is a uni-directional move through the state machine (prep -> pre -> upgrade -> rollback|commit), at least in the "sunny-day" scenario. However, do we want to set the observedGeneration field in the status on every pass through the reconcile loop (perhaps here)? It is likely easier to add now than to put it in place later.

controllers/ibu_controller.go
jc-rh: I was wondering if there is any kind of validation we could do. Feels like it's better to leave it to later stages where the spec fields are actually consumed.

imiller0 (commented):
Is there meaningful unit testing or CI that can be added along with this? ie quick validation that the progression sets relevant status?

controllers/ibu_controller.go
jc-rh: Yes, I was planning to do that and some unit testing for the function.

controllers/ibu_controller.go
jc-rh: Yes, the full name is used for the one we are reconciling in the main function

controllers/ibu_controller.go
jc-rh: Hmm, that would make it more like a node manager CR (similar to cluster version) instead of a job. There would be more complexity from resetting it in various stages, instead of this little piece of logic.

controllers/utils/conditions.go
jc-rh: We can. We don't need it so much as the policy will be verifying distinct condition types. This does remind me a question. Do we allow any other spec update once the thing is started?

controllers/ibu_controller.go
jc-rh: Updated

controllers/ibu_controller.go
imiller0: A couple thoughts...
1. I would expect any transition to "idle" to stop any in flight processes (eg stop pre-caching)
2. We should identify which transitions we "allow", at least for now, perhaps relaxing/extending later.

controllers/ibu_controller.go
imiller0: My inclination is that it should still be a cluster scoped CR rather than namespaced (not sure what it would mean for a cluster-wide operation to be defined within a namespace). If we allow multiple versions (ie unique names) then we need tight logic around only 1 is allowed to be in any state other than idle.

controllers/utils/conditions.go
imiller0: Certain fields are used by different phases (eg backup crs used by pre phase). It would be user-friendly to allow these fields to be set at the same time as enabling that phase.

controllers/ibu_controller.go
jc-rh: 1. How about deleting the CR instead? The finalizer has to do clean up either way. Putting it back to idle makes sense only if we have a good reason to reuse the CR or whatever behind it (in that case, it would be more like a pause/resume which is not really the same thing as the initial idle state
3. Definitely

controllers/ibu_controller.go
jc-rh: The logic is already implemented in `isIbuActive`. Is it not tight enough?
We can make it cluster scoped. Right now the operator is single namespaced. That probably needs to change too. In theory, many other operators we've seen should only be installed once only. However most of them are single namespaced with a default namespace that customers shouldn't change, like openshift-ptp ....

controllers/utils/conditions.go
jc-rh: Ok, we'll document this for each field then.

jc-rh: `observedGeneration` updated

controllers/ibu_controller.go
```diff
@@ -109,53 +114,167 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	currentPendingStage := utils.GetCurrentPendingStage(ibu)
+	desiredStage := ibu.Spec.Stage
+
+	if currentPendingStage == desiredStage {
+		nextReconcile, err = r.handleStage(ctx, ibu, currentPendingStage)
+	} else if isAborting(ibu) {
+		nextReconcile, err = r.handleAbort(ctx, ibu)
+	} else {
+		if currentPendingStage != "" {
+			nextReconcile, err = r.handleStage(ctx, ibu, currentPendingStage)
```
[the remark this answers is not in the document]
jc-rh: Yes, stage in spec can be set anything and we don't plan to use webhook to reject. There will be conditions to flag invalid stage transition. Meanwhile, the current pending action needs to continue, as the requeued reconcile request with delay gets replaced by the immediate request from the spec update. If we don't reconcile the current action, we won't wake up again until another spec update. In theory, users should correct the spec. But I think it's nicer to keep updating the current stage conditions even if users don't correct the spec.

controllers/ibu_controller.go
```diff
@@ -109,53 +114,167 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		}
+		if err != nil {
```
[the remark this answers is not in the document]
jc-rh: No. Error returned to the reconciler means unexpected error (e.g. API call failure). Controller managed would discard any status update from this reconcile and re-invoke the reconcile a few seconds later. To present an error to users, we should be setting it in the condition for that stage, as part of handler function.

jc-rh:
@eranco74 Thanks for the comments. I forgot to put the PR on hold as we are still refining the stage transition. Your comments are still valid though. So I responded.

controllers/prep_handlers.go
```diff
@@ -0,0 +1,45 @@
...
+		utils.ConditionReasons.Completed,
+		metav1.ConditionTrue,
+		"Prep completed",
+		ibu.Generation)
+	utils.SetStatusCondition(&ibu.Status.Conditions,
+		utils.GetInProgressConditionType(ranv1alpha1.Stages.Prep),
+		utils.ConditionReasons.Completed,
+		metav1.ConditionFalse,
+		"Prep completed",
+		ibu.Generation)
```
Missxiaoguo, line 43: The Idle condition is updated with reason "InProgress" prior to processing each stage. Should it be updated to "Completed" in all completed cases?

controllers/ibu_controller.go
```diff
@@ -109,53 +114,178 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	return
+}
 
-	// Add finalizer for this CR.
-	if !controllerutil.ContainsFinalizer(ibu, utils.CleanupFinalizer) {
-		controllerutil.AddFinalizer(ibu, utils.CleanupFinalizer)
-		err := r.Update(ctx, ibu)
-		if err != nil {
-			return utils.StopReconciling, err
+func isRollbackAllowed(ibu *ranv1alpha1.ImageBasedUpgrade) bool {
```
Missxiaoguo, line 192: Perhaps a description/comment can help people quickly understand when a rollback is allowed. Same as finalize and abort.

controllers/prep_handlers.go
jc-rh: The idle condition is for the whole IBU. It doesn't go back to true until upgrade/rollback/abort is done.

Missxiaoguo: Right, the idle status should still stay in false, but what about the condition reason?  Previously, it was set to Inprogress with desired stage, should it be updated to Completed after the desired stage is completed?

					utils.ConditionTypes.Idle,
					utils.ConditionReasons.InProgress,
					metav1.ConditionFalse,
					string(utils.GetInProgressConditionType(desiredStage)),
					ibu.Generation)

controllers/ibu_controller.go
jc-rh: Updated

PR #15, opened 2023-10-14
Shell script integration

jc-rh:


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ configMapGenerator:
   name: manager-config
 images:
 - name: controller
-  newName: quay.io/openshift-kni/lifecycle-agent-operator
+  newName: quay.io/junchen/lifecycle-agent-operator
```
donpenney, line 15: Should drop this change

controllers/prep_handlers.go
```diff
@@ -18,28 +18,97 @@ package controllers
...
+		var content []byte
+		content, err = os.ReadFile(progressfile)
+		if err != nil {
+			return
+		}
+
+		progress := strings.TrimSpace(string(content))
+		r.Log.Info("Prep progress: " + progress)
+
+		if progress == "Completed" {
```
donpenney, line 74: We can add a check for "Failed" as well

config/manager/kustomization.yaml
jc-rh: Done

controllers/prep_handlers.go
jc-rh: Added

PR #17, opened 2023-10-15
MGMT-15824: Gather networking info

tsorya:
Gathering network configuration files on PreUpgrade
Currently gathering paths:
1. hostname
2. system-connections
3. ovn chosen interface in order to know which interface is the default


controllers/upgrade_handlers.go
```diff
@@ -30,7 +30,11 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *ra
 	// pre upgrade steps
 	if err := r.ClusterConfig.FetchClusterConfig(ctx); err != nil {
```
jc-rh, line 31: @donpenney Do you think the prep handler can make use of these or we really only need them for the upgrade stage?

browsell: Generally we want to collect the data as close to the upgrade as possible, therefore we collect during the upgrade stage. The prep conceivably could be days before the actual upgrade

romfreiman:
What's the process? who approves it?

donpenney (approved):

PR #21, opened 2023-10-23
Integrate initial Prep handler scripts

donpenney:
This update breaks the initial upgrade prep handler script into three phases:
- Get the seed image and mount it
- Import data from the seed image to create the new ostree stateroot and deployment, and collect certs (which will move to the Upgrade stage)
- Precache images

The Prep handler runs each script in a chroot environment, in order.

Additionally, a HACKING.md file is introduced to provide developer commands.

This update also fixes a couple of stale ClusterGroupUpgrade references from initial code source.

/cc @jc-rh


HACKING.md
```diff
@@ -0,0 +1,112 @@
...
+  seedImageRef:
+    version: 4.13.9
+    image: quay.io/dpenney/upgbackup:lca-test-seed-v1
+EOF
+
+# Set the stage to Prep
+oc patch imagebasedupgrades.ran.openshift.io upgrade -p='{"spec": {"stage": "Prep"}}' --type=merge
+
+# Set the stage back to Idle
+oc patch imagebasedupgrades.ran.openshift.io upgrade -p='{"spec": {"stage": "IDle"}}' --type=merge
```
sudomakeinstall2, line 73: Idle case typo
I am not sure if it matters though

donpenney: Thanks!

HACKING.md
```diff
@@ -0,0 +1,112 @@
...
+# Set the stage to Prep
+oc patch imagebasedupgrades.ran.openshift.io upgrade -p='{"spec": {"stage": "Prep"}}' --type=merge
+
+# Set the stage back to Idle
+oc patch imagebasedupgrades.ran.openshift.io upgrade -p='{"spec": {"stage": "Idle"}}' --type=merge
+```
+
+## Cleanup
+```console
+# Delete LCA resources
```
jc-rh, line 78: `make undeploy` does most of the cleanup if not all

donpenney: Updated

controllers/prep_handlers.go
```diff
@@ -58,68 +149,69 @@ func (r *ImageBasedUpgradeReconciler) handlePrep(ctx context.Context, ibu *ranv1
...
 				ibu.Generation)
 			utils.SetStatusCondition(&ibu.Status.Conditions,
 				utils.GetInProgressConditionType(ranv1alpha1.Stages.Prep),
 				utils.ConditionReasons.Completed,
 				metav1.ConditionFalse,
-				"Prep completed",
+				"Prep failed",
 				ibu.Generation)
-		} else if progress == "Failed" {
+		} else if progress == "completed-seed-image-pull" {
```
jc-rh, line 178: Seems to make sense if we also update the in progress condition message with whatever in the progress file if it's not failed or completed case. We can leave it to the next PR though.

jc-rh (approved):

PR #22, opened 2023-10-25
Enable linters

jc-rh:


jc-rh:
@eranco74 I had to unexport some of your vars. Feel free to export them again if needed, with a code comment so linter is happy.

donpenney (approved):

PR #23, opened 2023-10-26
Add IBU Imager to LCA

donpenney:


donpenney:
@jc-rh @leo8a For now, I've updated hack/lint.sh to skip ibu-imager. We'll need to take a look at resolving warnings, and make sure any other pieces function properly for ibu-imager living in a subdirectory

leo8a:
This is just to confirm that I've tried the new packaging of the Imager tool (as a new entrypoint in the lifecycle-agent-operator container image) in my env, and it works as expected.

PR #25, opened 2023-10-26
Add handle abort

sudomakeinstall2:
- undeploy and remove stateroot after a failed or successful Prep

### DIFF
--- a/controllers/idle_handlers.go
+++ b/controllers/idle_handlers.go
@@ -18,17 +18,75 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"path/filepath"
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
+	rpmostreeclient "github.com/openshift-kni/lifecycle-agent/ibu-imager/ostreeclient"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+func cleanupUnbootedStateroot(stateroot string) error {
+	status, err := rpmostreeclient.QueryStatusChroot(utils.Host)
+	if err != nil {
+		return err
+	}
+
+	undeployIndices := make([]int, 0)
+	for index, deployment := range status.Deployments {
+		if deployment.OSName != stateroot {
+			continue
+		}
+		// make sure none of the deployments in the stateroot is booted
+		if deployment.Booted {
+			return fmt.Errorf("failed abort: deployment %d in stateroot %s is booted", index, stateroot)
+		}
+		undeployIndices = append(undeployIndices, index)
+	}
+
+	// since undeploy shifts the order, undeploy in the reverse order
+	for i := len(undeployIndices) - 1; i >= 0; i-- {
+		ostreeUndeployCommand := fmt.Sprintf("ostree admin undeploy %d", i)
+		err := utils.ExecuteChrootCmd(utils.Host, ostreeUndeployCommand)
+		if err != nil {
+			return fmt.Errorf("ostree undeploy %d failed: %w", i, err)
+		}
+	}
+
+	// remove stateroot
+	sysrootPath := fmt.Sprintf("/sysroot/ostree/deploy/%s", stateroot)
+	if _, err := os.Stat(pathOutsideChroot(sysrootPath)); err == nil {
+		removeSysrootCommand := fmt.Sprintf("unshare -m /bin/sh -c \"mount -o remount,rw /sysroot && rm -rf %s\"", sysrootPath)
+		err := utils.ExecuteChrootCmd(utils.Host, removeSysrootCommand)
+		if err != nil {
+			return fmt.Errorf("removing stateroot failed: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func cleanupIBUFiles() error {
+	if _, err := os.Stat(pathOutsideChroot(utils.IBUWorkspacePath)); err == nil {
+		if err := os.RemoveAll(filepath.Join(utils.Host, utils.IBUWorkspacePath)); err != nil {
+			return fmt.Errorf("removing %s failed: %w", utils.IBUWorkspacePath, err)
+		}
+	}
+	return nil
+}
+
 //nolint:unparam
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
-
-	// TODO actual steps
-	// If succeeds, return doNotRequeue
-	return doNotRequeue(), nil
+	stateroot := getStaterootName(ibu.Spec.SeedImageRef.Version)
+	r.Log.Info("Cleanup stateroot", "stateroot", stateroot)
+	err := cleanupUnbootedStateroot(stateroot)
+	if err != nil {
+		return doNotRequeue(), err
+	}
+	err = cleanupIBUFiles()
+	return doNotRequeue(), err
 }
 
 //nolint:unparam
--- a/HACKING.md
+++ b/HACKING.md
@@ -38,7 +38,7 @@ For example:
 # Update bindata, then build and push the image
 make IMAGE_TAG_BASE=quay.io/${MY_REPO_ID}/lifecycle-agent-operator VERSION=latest ENGINE=podman \
     update-bindata \
-    docker-build
+    docker-build \
     docker-push
 
 # Deploy the operator to your SNO (with KUBECONFIG set appropriately)
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main/main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run /main/main.go
+	go run ./main/main.go
 
 debug: manifests generate fmt vet ## Run a controller from your host that accepts remote attachment.
 	dlv debug --headless --listen 127.0.0.1:2345 --api-version 2 --accept-multiclient ./main.go
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -366,6 +366,6 @@ func (r *ImageBasedUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func getOSName(seedImageVersion string) string {
+func getStaterootName(seedImageVersion string) string {
 	return fmt.Sprintf("rhcos_%s", seedImageVersion)
 }
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -41,7 +41,7 @@ func (r *ImageBasedUpgradeReconciler) launchGetSeedImage(
 	}
 
 	// Write the script
-	scriptname := filepath.Join(utils.Path, utils.PrepGetSeedImage)
+	scriptname := filepath.Join(utils.IBUWorkspacePath, utils.PrepGetSeedImage)
 	scriptcontent, _ := generated.Asset(utils.PrepGetSeedImage)
 	err = os.WriteFile(pathOutsideChroot(scriptname), scriptcontent, 0o700)
 
@@ -65,7 +65,7 @@ func (r *ImageBasedUpgradeReconciler) launchPullImages(
 	}
 
 	// Write the script
-	scriptname := filepath.Join(utils.Path, utils.PrepPullImages)
+	scriptname := filepath.Join(utils.IBUWorkspacePath, utils.PrepPullImages)
 	scriptcontent, _ := generated.Asset(utils.PrepPullImages)
 	err = os.WriteFile(pathOutsideChroot(scriptname), scriptcontent, 0o700)
 
@@ -89,7 +89,7 @@ func (r *ImageBasedUpgradeReconciler) launchSetupStateroot(
 	}
 
 	// Write the script
-	scriptname := filepath.Join(utils.Path, utils.PrepSetupStateroot)
+	scriptname := filepath.Join(utils.IBUWorkspacePath, utils.PrepSetupStateroot)
 	scriptcontent, _ := generated.Asset(utils.PrepSetupStateroot)
 	err = os.WriteFile(pathOutsideChroot(scriptname), scriptcontent, 0o700)
 
@@ -101,7 +101,7 @@ func (r *ImageBasedUpgradeReconciler) launchSetupStateroot(
 
 	cmd := fmt.Sprintf("%s --seed-image %s --progress-file %s --os-version %s --os-name %s",
 		scriptname, ibu.Spec.SeedImageRef.Image, progressfile,
-		ibu.Spec.SeedImageRef.Version, getOSName(ibu.Spec.SeedImageRef.Version))
+		ibu.Spec.SeedImageRef.Version, getStaterootName(ibu.Spec.SeedImageRef.Version))
 	go utils.ExecuteChrootCmd(utils.Host, cmd)
 	result = requeueWithShortInterval()
 
@@ -115,7 +115,7 @@ func (r *ImageBasedUpgradeReconciler) runCleanup(
 	}
 
 	// Write the script
-	scriptname := filepath.Join(utils.Path, utils.PrepCleanup)
+	scriptname := filepath.Join(utils.IBUWorkspacePath, utils.PrepCleanup)
 	scriptcontent, _ := generated.Asset(utils.PrepCleanup)
 	err := os.WriteFile(pathOutsideChroot(scriptname), scriptcontent, 0o700)
 
@@ -142,15 +142,15 @@ func (r *ImageBasedUpgradeReconciler) handlePrep(ctx context.Context, ibu *lcav1
 		return
 	}
 
-	if _, err = os.Stat(pathOutsideChroot(utils.Path)); os.IsNotExist(err) {
-		err = os.Mkdir(pathOutsideChroot(utils.Path), 0o700)
+	if _, err = os.Stat(pathOutsideChroot(utils.IBUWorkspacePath)); os.IsNotExist(err) {
+		err = os.Mkdir(pathOutsideChroot(utils.IBUWorkspacePath), 0o700)
 	}
 
 	if err != nil {
 		return
 	}
 
-	progressfile := filepath.Join(utils.Path, "prep-progress")
+	progressfile := filepath.Join(utils.IBUWorkspacePath, "prep-progress")
 
 	_, err = os.Stat(pathOutsideChroot(progressfile))
 
--- a/controllers/utils/cmd.go
+++ b/controllers/utils/cmd.go
@@ -25,7 +25,7 @@ import (
 
 // Constants for file and directory names
 const (
-	Path               string = "/var/ibu"
+	IBUWorkspacePath   string = "/var/ibu"
 	Host               string = "/host"
 	PrepGetSeedImage   string = "prepGetSeedImage.sh"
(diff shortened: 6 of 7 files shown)


controllers/idle_handlers.go
```diff
@@ -18,15 +18,19 @@ package controllers
...
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *ranv1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
-
-	// TODO actual steps
-	// If succeeds, return doNotRequeue
+	ostreeUndeployCommand := fmt.Sprintf("ostree admin undeploy rhcos_%s", ibu.Spec.SeedImageRef.Version)
+	utils.ExecuteChrootCmd(utils.Host, ostreeUndeployCommand)
```
jc-rh, line 30: The execute functions should return err now. I didn't do it initially as we were using it in background only and error/progress checking was based on the progress file.

controllers/idle_handlers.go
```diff
@@ -18,15 +18,19 @@ package controllers
...
 	ranv1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *ranv1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
-
-	// TODO actual steps
-	// If succeeds, return doNotRequeue
+	ostreeUndeployCommand := fmt.Sprintf("ostree admin undeploy rhcos_%s", ibu.Spec.SeedImageRef.Version)
```
jc-rh, line 29: The os name can be generated with the same function and passed in to the prep script as an arg.

sudomakeinstall2: currently the prep script is getting the version directly from seed image, which could be different from what is specified in the spec. What are we going to do about this?

donpenney, line 29: The undeploy command takes the deployment index, not the stateroot

controllers/idle_handlers.go
```diff
@@ -18,15 +18,19 @@ package controllers
...
+	utils.ExecuteChrootCmd(utils.Host, ostreeUndeployCommand)
+	removeSysrootFilesCommand := fmt.Sprintf("unshare -m /bin/sh -c \"mount -o remount,rw /sysroot && rm -rf /sysroot/ostree/deploy/rhcos_%s\"", ibu.Spec.SeedImageRef.Version)
+	utils.ExecuteChrootCmd(utils.Host, removeSysrootFilesCommand)
+	utils.ExecuteChrootCmd(utils.Host, "rm -rf /var/ibu/")
```
donpenney, line 33: There is a constant in controllers/utils/cmd.go for this path /var/ibu, named Path (which should really be renamed)

controllers/idle_handlers.go
donpenney: We probably could use utility functions for determining:
- stateroot for either side of the upgrade
- deployment index (again, for either side)
- maybe the deployment id, but I'm not sure if the LCA itself will need that

sudomakeinstall2: I have tested this on 4.10.15 and it works with the name. Maybe it has changed in newer versions.

sudomakeinstall2: I think it is an undocumented feature
`ostree admin undeploy rhcos_4.10.15
Bootloader updated; bootconfig swap: yes; deployment count change: -1
Deleted deployment 3f573c36a90b1c2412487be2401c2c0442c75da829c26182a59029494279739c.0
`

donpenney: I think we'd should confirm with @cgwalters or someone else from the coreos team if we want to use an "undocumented feature"

donpenney: I think we should also be confirming the existence of the stateroot and deployment first to determine whether there's any cleanup required.

jc-rh: @javipolo is the ostree plugin going to provide this function? will it be based on index or name?

jc-rh: @sudomakeinstall2 What Don mentioned in his last comment is essentially what I was thinking for CNF-10134. This kind of data seems to be generally useful for multiple stage handlers. Therefore it makes sense to implement it as a part of main reconcile logic before entering stage specific parts.

javipolo: > @javipolo is the ostree plugin going to provide this function? will it be based on index or name?

what do you mean by the ostree plugin?
I think we'll need someone from RHCOS to clarify ostree functionality here ... maybe @jmarrero can provide some insights?

controllers/idle_handlers.go
sudomakeinstall2: updated. changed to IBUWorkspacePath.

controllers/idle_handlers.go
sudomakeinstall2: there is `ibu-imager/internal/ostreeclient` but it is internal to IBU.

controllers/idle_handlers.go
sudomakeinstall2: do you mean we should change all the uses of execute functions in this PR?

controllers/idle_handlers.go
donpenney: > > @javipolo is the ostree plugin going to provide this function? will it be based on index or name?
> 
> what do you mean by the ostree plugin? I think we'll need someone from RHCOS to clarify ostree functionality here ... maybe @jmarrero can provide some insights?

@javipolo : @jc-rh is referring to the plugin for setting up and managing the new stateroot and deployment in CNF-9797

sudomakeinstall2: updated

sudomakeinstall2: updated

controllers/idle_handlers.go
```diff
@@ -18,16 +18,49 @@ package controllers
...
 	ranv1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+func getOSName(seedImageVersion string) string {
+	return fmt.Sprintf("rhcos_%s", seedImageVersion)
+}
+
+func (r *ImageBasedUpgradeReconciler) removeDeployment(osName string) {
```
jc-rh, line 33: Should this be a member function of the ostree client instead?

internal/rpmostreeclient/rpmostreeclient.go
```diff
@@ -0,0 +1,91 @@
...
+	Booted                  bool     `json:"booted"`
+	Staged                  bool     `json:"staged"`
+	LiveReplaced            string   `json:"live-replaced,omitempty"`
+	Origin                  string   `json:"origin"`
+	CustomOrigin            []string `json:"custom-origin"`
+	ContainerImageReference string   `json:"container-image-reference"`
+	RequestedPackages       []string `json:"requested-packages"`
+	RequestedBaseRemovals   []string `json:"requested-base-removals"`
+	Unlocked                *string  `json:"unlocked"`
+}
```
jc-rh, line 38: @donpenney which of these fields are good for the stateroot list in IBU status? I am guessing id, osname, booted at minimum?

controllers/idle_handlers.go
sudomakeinstall2: The client is for `rpm-ostree` not `os-tree` . If you prefer it can be moved there though.

jc-rh: Ah, ok. I think we will be using various `ostree` commands in multiple handlers. It's good to encapsulate them into another client type wrapper in internal. We can do it in another PR later though.

sudomakeinstall2: yes, i agree, but I thought this was part of the ostree plugin you mentioned.

controllers/idle_handlers.go
```diff
@@ -18,16 +18,49 @@ package controllers
...
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *ranv1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	OSName := getOSName(ibu.Spec.SeedImageRef.Version)
+	r.removeDeployment(OSName)
 
-	// TODO actual steps
-	// If succeeds, return doNotRequeue
+	sysrootDeployPath := fmt.Sprintf("/sysroot/ostree/deploy/%s", OSName)
+	if _, err := os.Stat(sysrootDeployPath); err != nil {
+		removeSysrootFilesCommand := fmt.Sprintf("unshare -m /bin/sh -c \"mount -o remount,rw /sysroot && rm -rf %s\"", sysrootDeployPath)
+		utils.ExecuteChrootCmd(utils.Host, removeSysrootFilesCommand)
```
jc-rh, line 57: As mentioned in my previous comment, the execute function needs to return err and the handler should decide what to do in different cases, either retry by itself or let it reconcile again by returning err or declare abort failure etc.

internal/rpmostreeclient/rpmostreeclient.go
```diff
@@ -0,0 +1,91 @@
+package rpmostreeclient
```
jc-rh, line 1: I would name the package `ostree`. We might add other ostree related commands here.

controllers/idle_handlers.go
sudomakeinstall2: do you mean change the execute functions in this PR?

controllers/idle_handlers.go
jc-rh: Actually, should we be using `rpm-ostree cleanup` instead? @donpenney

sudomakeinstall2: I don't think `rpm-ostree cleanup` is equivalent to `ostree undeploy index`

controllers/idle_handlers.go
jc-rh: Yes. I made the execute function for the prep handler demo and didn't think through other use cases.

controllers/idle_handlers.go
jc-rh: Ok, so we will be using both commands then. That's probably why it's better to have an ostree plugin that encapsulate them.

controllers/idle_handlers.go
jc-rh: We can leave the actual error handling logic for later. But we should correct what the function returns before it's getting used everywhere.

controllers/idle_handlers.go
sudomakeinstall2: updated

controllers/idle_handlers.go
sudomakeinstall2: updated

internal/rpmostreeclient/rpmostreeclient.go
sudomakeinstall2: updated

controllers/idle_handlers.go
```diff
@@ -18,16 +18,55 @@ package controllers
...
 	"context"
+	"fmt"
+	"os"
 
 	ranv1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+func getOSName(seedImageVersion string) string {
```
jc-rh, line 29: Point of this is to also use it in the prep handler.

controllers/idle_handlers.go
```diff
@@ -18,16 +18,55 @@ package controllers
...
 
-	// TODO actual steps
-	// If succeeds, return doNotRequeue
+	sysrootDeployPath := fmt.Sprintf("/sysroot/ostree/deploy/%s", OSName)
+	if _, err := os.Stat(sysrootDeployPath); err != nil {
+		removeSysrootFilesCommand := fmt.Sprintf("unshare -m /bin/sh -c \"mount -o remount,rw /sysroot && rm -rf %s\"", sysrootDeployPath)
+		if err := utils.ExecuteChrootCmd(utils.Host, removeSysrootFilesCommand); err != nil {
+			r.Log.Error(err, "Failed to remove sysroot files")
+		}
+	}
```
jc-rh, line 62: if the previous command fails, does this leave the node in a bad state?

donpenney: @jc-rh yes, very possibly. And if we proceeded to wiping the stateroot, it would be in a worse state. We also need to add a check here that there are no other deployments in the stateroot before deleting it

controllers/idle_handlers.go
sudomakeinstall2: can you point to where in prepHandler? osname is only used in the bash scripts and it is extracted from the seed image as I said in the previous comment. If we want to change that it is out of scope of this PR.

jc-rh:
```
new_osname="rhcos_${up_ver}"
```
Make this an arg passed in by the prep handler?

controllers/idle_handlers.go
sudomakeinstall2: @donpenney sorry I don't understand the second part, if we want to remove `rhcos_4.13.9` there should be only on directory in `/sysroot/ostree/deploy/`?

controllers/idle_handlers.go
sudomakeinstall2: That value comes from the seed image
```
up_ver=$(jq -r '.status.desired.version' "${img_mnt}/clusterversion.json")
if [ -z "${up_ver}" ]; then
    fatal "Failed to identify version from seed image"
fi
```
we need to handle the case when that value is not equal to the version specified in the CR

donpenney: The current prep handler scripts are just the initial implementation. The final implementation will presumably be go and would be able to make use of such functions

controllers/idle_handlers.go
jmarrero: I think you just stumbled on part of the code that needs some input sanitization.
I don't see any test where we are testing passing anything other than a number and giving the code a glance [here](https://github.com/ostreedev/ostree/blob/main/src/ostree/ot-admin-builtin-undeploy.c#L53C28-L53C44), I think we are just taking whatever the input it and trying to convert it to int.

Trying really quick to replicate:
```
jmarrero@fedora ~> sudo ostree admin undeploy trash
[sudo] password for jmarrero: 
error: Cannot undeploy currently booted deployment 0
```

I bet that deployment that is being undeployed index is 0?

sudomakeinstall2: yes I did realize that after I looked at the code, and changed the implementation.

controllers/idle_handlers.go
```diff
@@ -18,16 +18,57 @@ package controllers
...
+			}
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
 //nolint:unparam
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	OSName := getOSName(ibu.Spec.SeedImageRef.Version)
```
donpenney, line 50: This function should also be renamed to getStateroot. Currently, this function seems geared toward just the seed image version? We'd also need to clean up the original stateroot, which could be rhcos or rhcos_$version, after the upgrade is completed. I assume the handleAbort is only called from the A-side?

controllers/idle_handlers.go
```diff
@@ -18,16 +18,57 @@ package controllers
...
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+// removeDeployment removes the deployment with the specified name
+// in case there is no deployment with the specified name returns false
+func (r *ImageBasedUpgradeReconciler) removeDeployment(osName string) (bool, error) {
+	status, err := r.RpmOSTreeClient.QueryStatus()
```
donpenney, line 32: Can we also add a safety check that the booted deployment is not the same stateroot that's being cleaned up? That would be a very bad thing to have happen

controllers/idle_handlers.go
```diff
@@ -18,16 +18,57 @@ package controllers
...
+	if err != nil {
+		return false, err
+	}
+	for index, deployment := range status.Deployments {
+		if deployment.OSName == osName {
+			ostreeUndeployCommand := fmt.Sprintf("ostree admin undeploy %d", index)
```
donpenney, line 38: Because the undeploy will shift the indexes, you'll need to either get a new status, or do the undeploy commands in reverse order

controllers/idle_handlers.go
```diff
@@ -18,17 +18,76 @@ package controllers
...
+		}
+		// make sure none of the deployments in the stateroot is booted
+		if deployment.Booted {
+			return fmt.Errorf("failed abort: deployment %d in stateroot %s is booted", index, stateroot)
+		}
+		undeployIndices = append(undeployIndices, index)
+	}
+
+	// since undeploy shifts the order, undeploy in the reverse order
+	for i := len(undeployIndices) - 1; i >= 0; i-- {
```
pixelsoccupied, line 49: Since you're collect the index in `undeployIndices` and skipping when it's new stateroot (L45)...it be should be safe undeploy regardless of the direction of the loop?

controllers/idle_handlers.go
```diff
@@ -18,17 +18,76 @@ package controllers
...
+			return fmt.Errorf("removing stateroot failed: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func cleanupIBUFiles() error {
+	if _, err := os.Stat(pathOutsideChroot(utils.IBUWorkspacePath)); err == nil {
+		removeIBUFilesCommand := fmt.Sprintf("rm -rf %s", utils.IBUWorkspacePath)
```
pixelsoccupied, line 72: Can we do `os.RemoveAll("/tmp/")` instead of `rm` ?

internal/ostree/ostree.go
```diff
@@ -0,0 +1,50 @@
...
+	cmd.Stdout = &stdoutBytes
+	cmd.Stderr = &stderrBytes
+	cmd.Dir = "/"
+	err := cmd.Run()
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", stderrBytes.String(), err)
+	}
+	var q rpmostreeclient.Status
+	if err := json.Unmarshal(stdoutBytes.Bytes(), &q); err != nil {
+		return nil, fmt.Errorf("failed to parse `rpm-ostree status --json` output: %w", err)
```
pixelsoccupied, line 47: most likely you have checked...but are we intentionally skipping https://github.com/coreos/rpmostree-client-go ?

controllers/idle_handlers.go
sudomakeinstall2: it needs `chroot`.

internal/rpmostreeclient/rpmostreeclient.go
donpenney: @jc-rh Depends on what the intent is. The stateroot itself, ie. rhcos_$version, is the OSName here. The ID would be the stateroot + the deployment commit id and iterator. If it's for determining which release is active (A or B side), then we'd just need the OSName and Booted. The ID likely doesn't provide any use for LCA directly

controllers/idle_handlers.go
sudomakeinstall2: updated

internal/ostree/ostree.go
sudomakeinstall2: updated

PR #29, opened 2023-10-27
Add unit tests for prep handlers

sudomakeinstall2:
- Add fake cmd and os for unit tests


sudomakeinstall2:
If we plan to convert the scripts to go, then I think is not necessary and we can close this.

PR #30, opened 2023-10-30
ibu-imager: Use full refspec for the remote repository

javipolo:
Remove hardcoded image tag `oneimage` and force the command line parameter to include the full refspec.

Example:
`ibu-imager create -r quay.io/someone/ibu-seed-image:mytag`

### DIFF
--- a/ibu-imager/README.md
+++ b/ibu-imager/README.md
@@ -2,42 +2,43 @@
 
 [![CI](https://github.com/leo8a/ibu-imager/actions/workflows/pull_request_workflow.yml/badge.svg)](https://github.com/leo8a/ibu-imager/actions/workflows/pull_request_workflow.yml)
 
-This application will assist users to easily create an OCI seed image for the Image-Based Upgrade (IBU) workflow, using 
+This application will assist users to easily create an OCI seed image for the Image-Based Upgrade (IBU) workflow, using
 a simple CLI.
 
 ## Motivation
 
 One of the most important / critical day2 operations for Telecommunications cloud-native deployments is upgrading their
-Container-as-a-Service (CaaS) platforms as quick (and secure!) as possible while minimizing the disruption time of 
+Container-as-a-Service (CaaS) platforms as quick (and secure!) as possible while minimizing the disruption time of
 active workloads.
 
 A novel method to approach this problem can be derived based on the
-[CoreOS Layering](https://github.com/coreos/enhancements/blob/main/os/coreos-layering.md) concepts, which proposes a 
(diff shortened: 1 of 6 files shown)


ibu-imager/internal/seedcreator/seedcreator.go
```diff
@@ -32,21 +32,19 @@ type SeedCreator struct {
...
 func NewSeedCreator(log *logrus.Logger, ops ops.Ops, ostreeClient *ostree.Client, backupDir,
-	kubeconfig, containerRegistry, backupTag, authFile string) *SeedCreator {
+	kubeconfig, containerRegistry, authFile string) *SeedCreator {
 	return &SeedCreator{
 		log:               log,
 		ops:               ops,
 		ostreeClient:      ostreeClient,
 		backupDir:         backupDir,
 		kubeconfig:        kubeconfig,
 		containerRegistry: containerRegistry,
```
donpenney, line 47: Can we rename this field and argument as well? ie. --image instead of --registry? Or maybe add --image and (temporarily) keep --registry so dev can use both for the meantime?

javipolo: I would directly rename it, since we're so early in the process, and that would force people also to include the tag .... delaying it will only make the same trauma of changing in a while :D

WDYT?

donpenney: I'm completely fine with that!

javipolo: Done :)

ibu-imager/README.md
```diff
@@ -153,17 +154,17 @@ To create an IBU seed image out of your Single Node OpenShift (SNO), run the fol
  				-v /var:/var \
  				-v /var/run:/var/run \
  				-v /run/systemd/journal/socket:/run/systemd/journal/socket \
- 				quay.io/lochoa/ibu-imager:4.14.0 create --authfile /var/lib/kubelet/config.json --registry ${LOCAL_USER_REGISTRY}
+ 				quay.io/lochoa/ibu-imager:4.14.0 create --authfile /var/lib/kubelet/config.json --image quay.io/somewhere/ibu-seed-images:4.14.0-du
```
leo8a, line 157: could we also update the HACKING.md docs about this change? thanks!

https://github.com/openshift-kni/lifecycle-agent/blob/main/HACKING.md#running-the-ibu-imager

javipolo: done! :)

donpenney (approved):

PR #33, opened 2023-10-30
Fix check condition for crio down

javipolo:
If crio is down, `systemctl is-active` ends with status code 3, so we need to capture that exit code and continue execution
/cc @leo8a 
/cc @tsorya

### DIFF
--- a/ibu-imager/seedcreator/seedcreator.go
+++ b/ibu-imager/seedcreator/seedcreator.go
@@ -1,9 +1,11 @@
 package seedcreator
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"path"
 	"strings"
 
@@ -155,7 +157,9 @@ func (s *SeedCreator) stopServices() error {
 
 	s.log.Println("Stopping containers and CRI-O runtime.")
 	crioSystemdStatus, err := s.ops.SystemctlAction("is-active", "crio")
-	if err != nil {
+	var exitErr *exec.ExitError
+	// If ExitCode is 3, the command succeeded and told us that crio is down
+	if err != nil && errors.As(err, &exitErr) && exitErr.ExitCode() != 3 {
 		return err
 	}
 	s.log.Println("crio status is", crioSystemdStatus)


ibu-imager/internal/seedcreator/seedcreator.go
```diff
@@ -156,7 +157,13 @@ func (s *SeedCreator) stopServices() error {
 	s.log.Println("Stopping containers and CRI-O runtime.")
 	crioSystemdStatus, err := s.ops.SystemctlAction("is-active", "crio")
 	if err != nil {
-		return err
+		exitError, ok := err.(*exec.ExitError)
```
mresvanis, line 160: Nice @javipolo! Perhaps we could simplify the nested `if`s with something like this:

```go
var exitErr *exec.ExitError
if err != nil && errors.As(err, &exitErr) && exitErr.ExitCode() != 3 {
    return err
}
```

WDYT?

javipolo: I love it @mresvanis ! 
The nested ifs were something I wanted to get rid of, but I couldnt make them go in a nice way ....
Change added and tested! Thanks!

leo8a (approved):

donpenney (approved):

ibu-imager/internal/seedcreator/seedcreator.go
```diff
@@ -155,7 +156,9 @@ func (s *SeedCreator) stopServices() error {
 
 	s.log.Println("Stopping containers and CRI-O runtime.")
 	crioSystemdStatus, err := s.ops.SystemctlAction("is-active", "crio")
-	if err != nil {
+	var exitErr *exec.ExitError
+	// If ExitCode is 3, the command succeeded and told us that crio is down
+	if err != nil && errors.As(err, &exitErr) && exitErr.ExitCode() != 3 {
```
donpenney, line 161: This PR needs to be rebased for https://github.com/openshift-kni/lifecycle-agent/pull/38

donpenney (approved):

PR #35, opened 2023-10-30
[imager] Add recert force-expire towards seed cluster 

leo8a:
This validation ensures that the seed SNO cluster can be re-certified during the IBU workflow later on without errors. Additionally, it forces the expiration of the certificates in the seed SNO cluster.

### DIFF
--- a/ibu-imager/cmd/create.go
+++ b/ibu-imager/cmd/create.go
@@ -45,6 +45,7 @@ var (
 
 	// recertContainerImage is the container image for the recert tool
 	recertContainerImage string
+	recertSkipValidation bool
 )
 
 func init() {
@@ -67,11 +68,13 @@ func init() {
 	// Add create command
 	rootCmd.AddCommand(createCmd)
 
-	// Add flags related to container registry
+	// Add flags to imager command
 	createCmd.Flags().StringVarP(&authFile, "authfile", "a", imageRegistryAuthFile, "The path to the authentication file of the container registry.")
 	createCmd.Flags().StringVarP(&containerRegistry, "image", "i", "", "The full image name with the container registry to push the OCI image.")
 	createCmd.Flags().StringVarP(&recertContainerImage, "recert-image", "e", defaultRecertImage, "The full image name for the recert container tool.")
+	createCmd.Flags().BoolVarP(&recertSkipValidation, "skip-recert-validation", "", false, "Skips the validations performed by the recert tool.")
 
+	// Mark flags as required
 	createCmd.MarkFlagRequired("image")
 }
 
@@ -99,7 +102,7 @@ func create() {
 	}
 
 	seedCreator := seed.NewSeedCreator(client, log, op, rpmOstreeClient, backupDir, kubeconfigFile,
-		containerRegistry, authFile, recertContainerImage)
+		containerRegistry, authFile, recertContainerImage, recertSkipValidation)
 	err = seedCreator.CreateSeedImage()
 	if err != nil {
 		log.Fatal(err)
--- a/ibu-imager/cmd/create.go
+++ b/ibu-imager/cmd/create.go
@@ -73,7 +73,6 @@ func init() {
 	createCmd.Flags().StringVarP(&recertContainerImage, "recert-image", "e", defaultRecertImage, "The full image name for the recert container tool.")
 
 	createCmd.MarkFlagRequired("image")
-	createCmd.MarkFlagRequired("recert-image")
 }
 
 func create() {
--- a/ibu-imager/cmd/create.go
+++ b/ibu-imager/cmd/create.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cmd
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 
@@ -37,10 +36,15 @@ import (
 
 var (
 	scheme = runtime.NewScheme()
+
 	// authFile is the path to the registry credentials used to push the OCI image
 	authFile string
+
 	// containerRegistry is the registry to push the OCI image
 	containerRegistry string
+
+	// recertContainerImage is the container image for the recert tool
+	recertContainerImage string
 )
 
 func init() {
@@ -48,9 +52,6 @@ func init() {
 	utilruntime.Must(v1.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
 }
-
-// recertContainerImage is the container image for the recert tool
-var recertContainerImage string
 
 // createCmd represents the create command
 var createCmd = &cobra.Command{
@@ -70,19 +71,15 @@ func init() {
 	createCmd.Flags().StringVarP(&authFile, "authfile", "a", imageRegistryAuthFile, "The path to the authentication file of the container registry.")
 	createCmd.Flags().StringVarP(&containerRegistry, "image", "i", "", "The full image name with the container registry to push the OCI image.")
 	createCmd.Flags().StringVarP(&recertContainerImage, "recert-image", "e", defaultRecertImage, "The full image name for the recert container tool.")
+
+	createCmd.MarkFlagRequired("image")
+	createCmd.MarkFlagRequired("recert-image")
 }
 
 func create() {
 
 	var err error
 	log.Printf("OCI image creation has started")
-
-	// Check if containerRegistry was provided by the user
-	if containerRegistry == "" {
-		fmt.Printf(" *** Please provide a valid container registry to store the created OCI images *** \n")
-		log.Info("Skipping OCI image creation.")
-		return
-	}
 
 	op := ops.NewOps(log, ops.NewExecutor(log, true))
 	rpmOstreeClient := ostree.NewClient("ibu-imager", op)
@@ -94,12 +91,12 @@ func create() {
 
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigFile)
 	if err != nil {
-		log.Fatal("failed top create k8s config", err)
+		log.Fatal("Failed to create k8s config", err)
 	}
 
 	client, err := runtimeClient.New(config, runtimeClient.Options{Scheme: scheme})
 	if err != nil {
-		log.Fatal("failed to create runtime client", err)
+		log.Fatal("Failed to create runtime client", err)
 	}
 
 	seedCreator := seed.NewSeedCreator(client, log, op, rpmOstreeClient, backupDir, kubeconfigFile,
@@ -127,9 +124,6 @@ func copyConfigurationScripts() error {
 	log.Infof("Copying installation_configuration_files/scripts to local/bin")
 	return cp.Copy("installation_configuration_files/scripts", "/var/usrlocal/bin", cp.Options{AddPermission: os.FileMode(0o777)})
 }
-
-// copyEnvFile reads and copy the env file from the source directory to the destination directory
-// while performing variable substitution for each var.
 
 func handleServices(ops ops.Ops) error {
 	dir := "installation_configuration_files/services"
--- a/ibu-imager/ops/ops.go
+++ b/ibu-imager/ops/ops.go
@@ -2,9 +2,18 @@ package ops
 
 import (
 	"fmt"
+	"net/http"
+	"os"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	// Default location for etcdStaticPodFile
+	etcdStaticPodFile = "/etc/kubernetes/manifests/etcd-pod.yaml"
 )
 
 // Ops is an interface for executing commands and actions in the host namespace
@@ -14,6 +23,8 @@ type Ops interface {
 	SystemctlAction(action string, args ...string) (string, error)
 	RunInHostNamespace(command string, args ...string) (string, error)
 	RunBashInHostNamespace(command string, args ...string) (string, error)
+	WaitForEtcd(endpoint string) error
+	GetImageFromPodDefinition(containerImage string) (string, error)
 }
 
 type ops struct {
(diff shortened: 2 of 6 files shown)


ibu-imager/internal/seedcreator/seedcreator.go
```diff
@@ -183,6 +193,72 @@ func (s *SeedCreator) stopServices() error {
 	return nil
 }
 
+func (s *SeedCreator) runRecert() error {
```
tsorya, line 196: This is validation or actual recert?

ibu-imager/cmd/create.go
```diff
@@ -61,9 +64,14 @@ func create() {
...
 	if containerRegistry == "" {
-		fmt.Printf(" *** Please provide a valid container registry to store the created OCI images *** \n")
-		log.Info("Skipping OCI image creation.")
-		return
+		log.Printf(" *** Please provide a valid container registry to store the created OCI images *** ")
+		log.Fatal("Skipping OCI image creation.")
+	}
+
+	// Check if recertContainerImage was provided by the user
+	if recertContainerImage == "" {
```
tsorya, line 72: In case we have this new param we should change 
https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/installation_configuration_files/scripts/installation-configuration.sh#L32
make it template and set image to it before copying this file to the seed image

ibu-imager/internal/seedcreator/seedcreator.go
```diff
@@ -183,6 +193,72 @@ func (s *SeedCreator) stopServices() error {
...
+		_, err = s.ops.RunInHostNamespace(
+			"podman", []string{"kill", "recert_etcd"}...)
+		if err != nil {
+			s.log.Errorf("Failed to kill recert_etcd container: %v", err)
+		}
+	}()
+
+	// TODO: wait for etcd server programmatically
+	s.log.Info("Wait 10 secs for unauthenticated etcd start serving")
+	time.Sleep(10 * time.Second)
```
tsorya, line 235: Lets have smarted validation and verify etcd container is running

ibu-imager/internal/seedcreator/seedcreator.go
```diff
@@ -356,3 +432,35 @@ func (s *SeedCreator) backupOstreeOrigin(statusRpmOstree *ostree.Status) error {
 	log.Println("Backup of .origin created successfully.")
 	return nil
 }
+
+// getEtcdImageFromStaticDefinition reads the static definition of the etcd pod
+// and returns the image used by this pod.
+func getEtcdImageFromStaticDefinition() (string, error) {
+
+	// Read the YAML file
+	yamlData, err := os.ReadFile(etcdStaticPodFile)
```
tsorya, line 441: Maybe we can read file into pod struct?

ibu-imager/internal/seedcreator/seedcreator.go
leo8a: Added a WaitForEtcd function to ops in [f25aa67](https://github.com/openshift-kni/lifecycle-agent/pull/35/commits/f25aa67b2014b3e84dc53fc8a579a69da52ac120).

ibu-imager/internal/seedcreator/seedcreator.go
leo8a: Updated `GetEtcdImageFromStaticDefinition` function in [7ee04ba](https://github.com/openshift-kni/lifecycle-agent/pull/35/commits/7ee04ba680d0c631239e2545efb79221fea30058).

donpenney:
Would there be an added step for the cleanup to bring the SNO back up, or would the seed SNO certificates automatically update after?

leo8a:
> Would there be an added step for the cleanup to bring the SNO back up, or would the seed SNO certificates automatically update after?

Yea, an extra cleanup step is required to bring the SNO back up. Not only to rejuvenate via recert --extend-expiration, but also to disable reconfiguration systemd units and enable kubelet and crio.

donpenney:
> > Would there be an added step for the cleanup to bring the SNO back up, or would the seed SNO certificates automatically update after?
> 
> Yea, an extra cleanup step is required to bring the SNO back up. Not only to rejuvenate via recert --extend-expiration, but also to disable reconfiguration systemd units and enable kubelet and crio.

We have a cleanup script:
https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/hack/clean.sh

Can you add the command for fixing the certs? Maybe we can also add the cleanup to the imager to automatically bring the system back up once the seed image is pushed

leo8a:
> > > Would there be an added step for the cleanup to bring the SNO back up, or would the seed SNO certificates automatically update after?
> > 
> > 
> > Yea, an extra cleanup step is required to bring the SNO back up. Not only to rejuvenate via recert --extend-expiration, but also to disable reconfiguration systemd units and enable kubelet and crio.
> 
> We have a cleanup script: https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/hack/clean.sh
> 
> Can you add the command for fixing the certs? Maybe we can also add the cleanup to the imager to automatically bring the system back up once the seed image is pushed

Sure, will send a PR with the updates required.

After running the cleanup script the system will back up to running, however, the only thing missing would be a re-registration into ACM or similar, if that is desired at all.

ibu-imager/cmd/create.go
leo8a: This is being addressed in #47.

ibu-imager/cmd/create.go
```diff
@@ -79,9 +80,14 @@ func create() {
 
 	// Check if containerRegistry was provided by the user
 	if containerRegistry == "" {
```
omertuc, line 82: I believe we can just use https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired

ibu-imager/cmd/create.go
```diff
@@ -79,9 +80,14 @@ func create() {
...
-		fmt.Printf(" *** Please provide a valid container registry to store the created OCI images *** \n")
-		log.Info("Skipping OCI image creation.")
-		return
+		log.Printf(" *** Please provide a valid container registry to store the created OCI images *** ")
+		log.Fatal("Skipping OCI image creation.")
+	}
+
+	// Check if recertContainerImage was provided by the user
+	if recertContainerImage == "" {
```
omertuc, line 88: I believe we can just use https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired

ibu-imager/ops/ops.go
```diff
@@ -67,3 +79,85 @@ func (o *ops) RunBashInHostNamespace(command string, args ...string) (string, er
...
+	}
+	defer etcdClient.Close()
+
+	// Create a context with a timeout to limit the waiting duration
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Continuously attempt to retrieve a specific key from etcd
+	for {
+		_, err = etcdClient.Get(ctx, "some_key")
```
omertuc, line 106: etcd has a healthz endpoint, no need for checking for keys, which also means we can drop the etcd client dependency

tsorya: +1

ibu-imager/ops/ops.go
```diff
@@ -67,3 +79,85 @@ func (o *ops) RunBashInHostNamespace(command string, args ...string) (string, er
...
+	return nil
+}
+
+// GetImageFromPodDefinition reads a YAML file containing pod configuration data
+// and extracts the image name for a given container named.
+// It returns the image name if found, or an error if not found or encountered any issues.
+func (o *ops) GetImageFromPodDefinition(containerImage string) (string, error) {
+
+	// Define a struct to match the structure of the YAML data
+	type PodConfig struct {
```
tsorya, line 126: why not corev1.Pod ?

omertuc: huge dependency

leo8a: I agree with @omertuc on this, I'd advocate keeping it lighter for now, if that makes sense

ibu-imager/ops/ops.go
leo8a: +1

ibu-imager/ops/ops.go
```diff
@@ -67,3 +78,78 @@ func (o *ops) RunBashInHostNamespace(command string, args ...string) (string, er
...
+
+		// If the server responds with HTTP status 200 (OK), it's up and running
+		if resp.StatusCode == http.StatusOK {
+			return nil
+		}
+
+		// Wait for a short interval before retrying
+		time.Sleep(1 * time.Second)
+	}
+}
```
omertuc, line 110: * This is a rather simple, clear function that doesn't do anything worth documenting IMO
* Use `http.Get` because it automatically uses the default client
* Logging only err and not status code seems inconsistent to me. Also the log should contain some context to make the error less scary

```suggestion
func (o *ops) WaitForEtcd(healthzEndpoint string) error {
	for {
		resp, err := http.Get(healthzEndpoint)
		if err != nil {
			o.log.Infof("Waiting for etcd: %s", err)
			time.Sleep(1 * time.Second)
			continue
		}
		defer resp.Body.Close()
		
		if resp.StatusCode != http.StatusOK {
			o.log.Infof("Waiting for etcd, status: %d", resp.StatusCode)
			time.sleep(1 * time.Second)
			continue
		}
		
		return nil
	}
}
```

ibu-imager/ops/ops.go
```diff
@@ -67,3 +78,78 @@ func (o *ops) RunBashInHostNamespace(command string, args ...string) (string, er
...
+// It returns the image name if found, or an error if not found or encountered any issues.
+func (o *ops) GetImageFromPodDefinition(containerImage string) (string, error) {
+
+	// Define a struct to match the structure of the YAML data
+	type PodConfig struct {
+		APIVersion string `yaml:"apiVersion"`
+		Kind       string `yaml:"kind"`
+		Metadata   struct {
+			Name string `yaml:"name"`
+		} `yaml:"metadata"`
```
omertuc, line 123:
```suggestion
```

The YAML library ignores unknown fields, so we don't need to maintain all this

ibu-imager/ops/ops.go
```diff
@@ -67,3 +78,78 @@ func (o *ops) RunBashInHostNamespace(command string, args ...string) (string, er
...
+		Spec struct {
+			Containers []struct {
+				Name  string `yaml:"name"`
+				Image string `yaml:"image"`
+			} `yaml:"containers"`
+		} `yaml:"spec"`
+	}
+
+	// Read the YAML file
```
omertuc, line 132: Many unnecessary comments like this. Comments are good but only when something complex or non-obvious is happening

ibu-imager/ops/ops.go
```diff
@@ -67,3 +78,78 @@ func (o *ops) RunBashInHostNamespace(command string, args ...string) (string, er
...
+	yamlData, err := os.ReadFile(etcdStaticPodFile)
```
omertuc, line 133: The function has a generic name, so it should receive `etcdStaticPodFile` as a parameter and not hardcode it.

This would make the call-site look like `GetImageFromPodDefinition(etcdStaticPodFile, "etcd")` which I feel is much easier to follow.

Also create a `var etcdStaticPodEtcdContainer string = "etcd"` global var so the call looks even clearer:

`GetImageFromPodDefinition(etcdStaticPodFile, etcdStaticPodEtcdContainer)`

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +226,76 @@ func (s *SeedCreator) stopServices() error {
...
+			"-v", "/var/lib/etcd:/store",
+			etcdImage,
+			"--name", "editor",
+			"--data-dir", "/store"}...)
+	if err != nil {
+		return fmt.Errorf("failed to run recert_etcd container: %w", err)
+	}
+
+	// Ensure that the unauthenticated etcd server is killed before the runRecertValidation
+	// function returns, even if an error occurs.
```
omertuc, line 256: Comment feels a bit unnecessary,  cleanup regardless of error is what `defer` was created for, doesn't need to be explained

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +226,76 @@ func (s *SeedCreator) stopServices() error {
...
+	defer func() {
+		s.log.Info("Kill the unauthenticated etcd server")
```
omertuc, line 258: nit

```suggestion
		s.log.Info("Killing the unauthenticated etcd server")
```

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +226,76 @@ func (s *SeedCreator) stopServices() error {
...
+		if err != nil {
+			s.log.Errorf("Failed to kill recert_etcd container: %v", err)
+		}
+	}()
+
+	// Wait for unauthenticated etcd server
+	s.log.Info("Wait for unauthenticated etcd start serving for recert tool")
+	err = s.ops.WaitForEtcd("http://localhost:2379/health")
+	if err != nil {
+		return err
```
omertuc, line 270: Wrap the error

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +226,76 @@ func (s *SeedCreator) stopServices() error {
...
+			"-v", "/etc/machine-config-daemon:/machine-config-daemon",
+			s.recertContainerImage,
+			"--etcd-endpoint", "localhost:2379",
+			"--static-dir", "/kubernetes",
+			"--static-dir", "/kubelet",
+			"--static-dir", "/machine-config-daemon",
+			"--summary-file-clean", "/kubernetes/recert-summary.yaml",
+			"--force-expire"}...)
+	if err != nil {
+		return err
```
omertuc, line 292: Wrap the error

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -33,6 +35,8 @@ FROM scratch
 COPY . /
 `
 
+var etcdStaticPodContainer string = "etcd"
```
omertuc, line 38: Let's move it to be a const and be right next to etcdStaticPodFile

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +230,73 @@ func (s *SeedCreator) stopServices() error {
 	return nil
 }
 
+func (s *SeedCreator) runRecertValidation() error {
+	s.log.Println("Running recert --force-expire tool and save a summary without sensitive data.")
```
omertuc, line 234:
```suggestion
	s.log.Info("Running recert --force-expire tool and saving a summary without sensitive data.")
```

omertuc (changes requested):
Looks good, Just a couple more nits

javipolo (approved):

ibu-imager/cmd/create.go
```diff
@@ -70,20 +71,16 @@ func init() {
 	createCmd.Flags().StringVarP(&authFile, "authfile", "a", imageRegistryAuthFile, "The path to the authentication file of the container registry.")
 	createCmd.Flags().StringVarP(&containerRegistry, "image", "i", "", "The full image name with the container registry to push the OCI image.")
 	createCmd.Flags().StringVarP(&recertContainerImage, "recert-image", "e", defaultRecertImage, "The full image name for the recert container tool.")
+
+	createCmd.MarkFlagRequired("image")
+	createCmd.MarkFlagRequired("recert-image")
```
donpenney, line 76: If we have a default recert image value, we don't need to mark the flag as "required", making it transparent to existing workflows

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +229,73 @@ func (s *SeedCreator) stopServices() error {
...
+			"--rm", "--network=host", "--privileged", "--replace", "--authfile", s.authFile,
+			"-v", "/etc/kubernetes:/kubernetes",
+			"-v", "/var/lib/kubelet:/kubelet",
+			"-v", "/etc/machine-config-daemon:/machine-config-daemon",
+			s.recertContainerImage,
+			"--etcd-endpoint", "localhost:2379",
+			"--static-dir", "/kubernetes",
+			"--static-dir", "/kubelet",
+			"--static-dir", "/machine-config-daemon",
+			"--summary-file-clean", "/kubernetes/recert-summary.yaml",
```
donpenney, line 289: Does this lead to having the /etc/kubernetes/recert-summary.yaml file in the seed image? If we want it included, maybe a different name for the file would be good, to make it clear this isn't the post-upgrade recert summary file

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -222,6 +229,73 @@ func (s *SeedCreator) stopServices() error {
 	return nil
 }
 
+func (s *SeedCreator) runRecertValidation() error {
```
tsorya, line 232: I think we need a way to disable this validation, we can have default "run=true" but will be nice to have a way to run without it as i am not sure why i always need it in dev process

omertuc: As a workaround you can pass in a dummy image that doesn't do anything

ibu-imager/seedcreator/seedcreator.go
leo8a: > Does this lead to having the /etc/kubernetes/recert-summary.yaml file in the seed image?

yup, agree with your point also. Will change it to `/etc/kubernetes/recert-seed-summary.yaml` to make the distinction.

ibu-imager/seedcreator/seedcreator.go
tsorya: why not to add flag for skip?

omertuc: Not opposed, I said workaround

leo8a: k, added `--skip-recert-validation` flag, feel free to give it a spin

PR #36, opened 2023-10-30
Update api group names

jc-rh:


main.go
```diff
@@ -79,13 +78,10 @@ func main() {
 		Scheme:                 scheme,
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
-		LeaderElectionID:       "lca.ran.openshift.io",
+		LeaderElectionID:       "lca.openshift.io",
```
browsell, line 81: Can go in another PR but need to update for the correct values for SNO

jc-rh: There are rules for this on SNO? The ptp operator uses `ptp.openshift.io`

browsell: An example here,  https://github.com/openshift/ptp-operator/blob/release-4.13/main.go#L57

PR #38, opened 2023-10-31
Remove redundant github.com/pkg/errors pkg

mresvanis:
This PR removes the `github.com/pkg/errors` package in favor of the standard [errors](https://pkg.go.dev/errors) package both in LCA and in imager. 

> We can wrap any error using `fmt.Errorf("... %w ...", err)` instead of the previous `github.com/pkg/errors` API `errors.Wrap(err, ...)` or `errors.Wrapf(err, ...)`.


ibu-imager/internal/ops/ops.go
```diff
@@ -30,7 +30,7 @@ func (o *ops) SystemctlAction(action string, args ...string) (string, error) {
 	o.log.Infof("Running systemctl %s %s", action, args)
 	output, err := o.RunInHostNamespace("systemctl", append([]string{action}, args...)...)
 	if err != nil {
-		err = errors.Wrapf(err, "Failed executing systemctl %s %s", action, args)
+		err = fmt.Errorf("Failed executing systemctl %s %s: %w", action, args, err)
```
omertuc, line 33:
```suggestion
		err = fmt.Errorf("failed executing systemctl %s %s: %w", action, args, err)
```

mresvanis: Nice catch, updated :+1:

ibu-imager/internal/ops/execute.go
```diff
@@ -33,5 +33,5 @@ func (e *executor) Execute(command string, args ...string) (string, error) {
 	cmd.Stdout = &stdoutBytes
 	cmd.Stderr = &stderrBytes
 	err := cmd.Run()
-	return strings.TrimSpace(stdoutBytes.String()), errors.Wrap(err, stderrBytes.String())
```
omertuc, line 36: Unlike `errors.Wrap`, `fmt.Errorf` will not return `nil` if the only instance of `%w` in its string is nil

ibu-imager/internal/ops/execute.go
```diff
@@ -33,5 +33,5 @@ func (e *executor) Execute(command string, args ...string) (string, error) {
...
+	return strings.TrimSpace(stdoutBytes.String()), fmt.Errorf("%s: %w", stderrBytes.String(), err)
```
omertuc, line 36:
```suggestion
    stdoutBytesTrimmed = strings.TrimSpace(stdoutBytes.String())
    stderrBytesTrimmed = strings.TrimSpace(stderrBytes.String())
    if err != nil {
        return stdoutBytesTrimmed, fmt.Errorf("%s: %w", stderrBytes.String(), err)
    }

    return stdoutBytesTrimmed, nil
```

omertuc: (see [comment](https://github.com/openshift-kni/lifecycle-agent/pull/38/files#r1377851143))

mresvanis: Absolutely, great catch! Updated.

PR #39, opened 2023-10-31
Use RH go-toolset instead of dockerhub golang base image

mresvanis:
This PR replaces Dockerhub's [golang](https://registry.hub.docker.com/layers/library/golang/1.19/images/sha256-1611d4f97c5ab666d3c123d72b3bb646dd12bbf7577dd1388fdb191d54cdf440?context=explore) base image with the Red Hat's [go-toolset](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=653020421d14c242647051fc&container-tabs=overview) one. 

The latter is a bit bigger `449.3 MB` vs `361.55 MB`, but it's scanned for vulnerabilities and it's actively being maintained by RH.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.hub.docker.com/library/golang:1.19 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.19 as builder
 
 WORKDIR /workspace
 
@@ -18,7 +18,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a
 
 #####################################################################################################
 # Build the imager binary
-FROM registry.hub.docker.com/library/golang:1.19 as imager
+FROM registry.access.redhat.com/ubi9/go-toolset:1.19 as imager
 
 ENV CRIO_VERSION="v1.28.0"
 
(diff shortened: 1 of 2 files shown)


jc-rh:
Just noticed both go.mod files are on 1.20. Is the go toolset image available for 1.20?

mresvanis:
> Just noticed both go.mod files are on 1.20. Is the go toolset image available for 1.20?

Not yet, but it should be sometime in November 2023, as far as I can tell (because it's never officially scheduled beforehand). I noticed the difference of the `Dockerfile` go version and the `go.mod` too, but I thought we would just wait for the go-toolset 1.20 to be released to change the Dockerfile` go version to `1.20`. This change just replaces dockerhub with the RH registry, but we can close the PR and come back to it when the go-toolset 1.20 is out. WDYT?

donpenney:
It looks like we now need 1.20, as another PR is getting this error on the build test:

# ibu-imager/internal/seedcreator
internal/seedcreator/seedcreator.go:161:19: undefined: errors
note: module requires Go 1.20

mresvanis:
> It looks like we now need 1.20, as another PR is getting this error on the build test:
> 
> # ibu-imager/internal/seedcreator
> internal/seedcreator/seedcreator.go:161:19: undefined: errors note: module requires Go 1.20

https://github.com/openshift-kni/lifecycle-agent/pull/43

PR #41, opened 2023-11-02
Moving ibu imager to be part of LCA code base

tsorya:
Moving ibu imager to be part of LCA code base.
1. Removed go.mod from imager
2. Moving main to be part of main package
3. Handled Dockerfile build and removed crictl (didn't find any usage of it)
 4. Changed Dockerfile to copy full project in order to stop adding folder by folder

From now we can import ibu imager packages into LCA and vice a versa


donpenney:
crictl is being used in seedcreator.go when collecting the list of container images

Dockerfile
```diff
@@ -51,10 +35,10 @@ RUN yum -y install jq && \
...
 WORKDIR /
 
-COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/build/manager .
 
-COPY --from=imager /workspace/ibu-imager .
-COPY --from=imager /workspace/crictl /usr/bin/
+COPY --from=imager /workspace/build/ibu-imager .
 COPY ibu-imager/installation_configuration_files/ installation_configuration_files/
+RUN ls -l /
```
donpenney, line 42: I'm guessing this is a debugging leftover?

ibu-imager/Dockerfile
```diff
@@ -13,7 +13,7 @@ COPY vendor/ vendor/
 # Copy the Go source installation_configuration_files
 COPY main.go main.go
 COPY cmd/ cmd/
-COPY internal/ internal/
+COPY pkg/ internal/
```
donpenney, line 16: I assume we could just drop this dockerfile?

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -305,7 +305,7 @@ func (s *SeedCreator) createAndPushSeedImage() error {
 	if err != nil {
 		return fmt.Errorf("failed to query ostree status: %w", err)
 	}
-	if err = s.backupOstreeOrigin(statusRpmOstree); err != nil {
+	if err := s.backupOstreeOrigin(statusRpmOstree); err != nil {
```
donpenney, line 308: I didn't think the : was necessary here?

ibu-imager/Dockerfile
tsorya: yep, though to do in the next pr

ibu-imager/seedcreator/seedcreator.go
tsorya: lint in ci-job is failing on it

tsorya: It sort of right if you do in as oneliner than this err var is sort of new and exists only in context of this "if"

donpenney: Makes sense. Thanks!

leo8a:
/lgtm

(I like the idea, this would also simplify the code by allowing the reutilization of code from both parts - no rewrite stuff)

donpenney (approved):

PR #45, opened 2023-11-02
[imager] Refactor cleanup and add cmd to rejuvenate seed cluster

leo8a:


ibu-imager/hack/clean.sh
```diff
@@ -1,13 +1,27 @@
...
+sudo rm -rf /var/tmp/container_list.done /var/tmp/backup
+
+# 2) cleanup preparation scripts and services
+sudo rm -f /usr/local/bin/prepare-installation-configuration.sh /usr/local/bin/installation-configuration.sh
+sudo systemctl disable installation-configuration.service
+sudo systemctl disable prepare-installation-configuration.service
+rm -f /etc/systemd/system/installation-configuration.service /etc/systemd/system/prepare-installation-configuration.service
+
+# cleanup container images
+sudo podman rmi quay.io/alosadag/ibu-seed-sno0:oneimage --force
```
donpenney, line 13: This shouldn't be needed here. We could have the ibu-imager delete the image after the push, but at a minimum, we'd need to drop the hardcoded name

ibu-imager/hack/clean.sh
```diff
@@ -1,13 +1,27 @@
...
-    /usr/local/bin/installation-configuration.sh && \
-sudo systemctl disable installation-configuration.service && \
-sudo systemctl disable prepare-installation-configuration.service && \
-rm -f /etc/systemd/system/installation-configuration.service \
-    /etc/systemd/system/prepare-installation-configuration.service && \
-sudo podman rmi quay.io/alosadag/ibu-seed-sno0:oneimage --force && \
-sudo systemctl enable --now kubelet && \
-    sudo systemctl enable --now crio
+# 1) cleanup backup dir
+sudo rm -rf /var/tmp/container_list.done /var/tmp/backup
```
donpenney, line 4: The whole script would likely run as root, so sudo wouldn't be needed

ibu-imager/hack/clean.sh
```diff
@@ -1,13 +1,27 @@
...
+
+podman run --name recert_etcd --detach --rm --network=host --privileged --replace --authfile /var/lib/kubelet/config.json --entrypoint etcd -v /var/lib/etcd:/store ${ETCD_IMAGE} --name editor --data-dir /store
+podman run --name recert --rm --network=host --privileged --replace --authfile /var/lib/kubelet/config.json -v /etc/kubernetes:/kubernetes -v /var/lib/kubelet:/kubelet -v /etc/machine-config-daemon:/machine-config-daemon ${RECERT_IMAGE} --etcd-endpoint localhost:2379 --static-dir /kubernetes --static-dir /kubelet --static-dir /machine-config-daemon --extend-expiration
+
+podman rm -f recert_etcd
+podman rm -f recert
+
+# restore cluster services
+sudo systemctl enable --now kubelet
+sudo systemctl enable --now crio
```
donpenney, line 27: crio.service isn't actually enabled, so we don't need this last command. It gets started automatically by kubelet.service, I believe

ibu-imager/hack/clean.sh
leo8a: Agree, I removed it.

ibu-imager/hack/clean.sh
leo8a: Gotcha, removed all `sudo` prefixes.

donpenney (approved):

PR #47, opened 2023-11-02
[imager] Add envFile to dynamically configure vars in systemd units

leo8a:
This file can be used to configure on-the-fly the variables used / needed in the scripts called by the systemd units like  installation-configuration.service.

### DIFF
--- a/ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
+++ b/ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
@@ -32,9 +32,9 @@ function recert {
     OLD_IP=$(cat /etc/default/seed-ip)
 
     ETCD_IMAGE="$(jq -r '.spec.containers[] | select(.name == "etcd") | .image' </etc/kubernetes/manifests/etcd-pod.yaml)"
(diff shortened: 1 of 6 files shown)


ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -32,9 +32,9 @@ function recert {
     OLD_IP=$(cat /etc/default/seed-ip)
 
     ETCD_IMAGE="$(jq -r '.spec.containers[] | select(.name == "etcd") | .image' </etc/kubernetes/manifests/etcd-pod.yaml)"
-    RECERT_IMAGE="quay.io/edge-infrastructure/recert:latest"
+    local recert_tool_image=${RECERT_IMAGE}
```
javipolo, line 35: Maybe add a default value, so the script will be usable without environment file?

javipolo (changes requested):

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
leo8a: Agree. 

Updated the script to use `quay.io/edge-infrastructure/recert:latest` image as the default value.

PR #49, opened 2023-11-02
Create cluster info object that will be provided to installation script

tsorya:
Create cluster info object that will be provided to installation script in future.
Creating manifest.json on seed image that should be used in preparation phase
Adding additional cluster data for recert


leo8a:
Needs to rebase for #33.

PR #50, opened 2023-11-03
MGMT-15882: Cluster and domain renaming

tsorya:
Providing to recert new cluster name and domain
Adding dnsmasq custom.conf in order to support new domain.

### DIFF
--- a/ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
+++ b/ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
@@ -25,6 +25,25 @@ fi
 echo "${RELOCATION_CONFIG_PATH} has been found"
 # Replace this with a function that loads values from yaml file
 set +o allexport
+NEW_CLUSTER_NAME=$(jq -r '.cluster_name' "${RELOCATION_CONFIG_PATH}"/clusterinfo/manifest.json)
+NEW_CLUSTER_BASE_DOMAIN=$(jq -r '.domain' "${RELOCATION_CONFIG_PATH}"/clusterinfo/manifest.json)
+NEW_CLUSTER_FULL_DOMAIN="${NEW_CLUSTER_NAME}.${NEW_CLUSTER_BASE_DOMAIN}"
+
+function reconfigure_dnsmasq {
+    if [ -z $1 ]; then
+        echo "domain not defined"
+    else
+        echo "Updating dnsmasq with new domain"
+        # shellcheck disable=SC1009
+        cat << EOF > /etc/dnsmasq.d/customer-domain.conf
+address=/apps.$1/$2
+address=/api-int.$1/$2
+address=/api.$1/$2
+EOF
+        systemctl restart dnsmasq --no-block
+    fi
+}
+
 
 # Recertify
 function recert {
@@ -42,6 +61,12 @@ function recert {
     # remove this condition when we will support only single ip
     if [[ -n "${NODE_IP}" ]]; then
         recert_cmd="${recert_cmd} ""--cn-san-replace ${OLD_IP},${NODE_IP}"
+        if [[ -n "${SEED_FULL_DOMAIN}" ]]; then
+            recert_cmd="${recert_cmd} ""--cn-san-replace api.${SEED_FULL_DOMAIN},api.${NEW_CLUSTER_FULL_DOMAIN}"
+            recert_cmd="${recert_cmd} ""--cn-san-replace api-int.${SEED_FULL_DOMAIN},api-int.${NEW_CLUSTER_FULL_DOMAIN}"
+            recert_cmd="${recert_cmd} ""--cn-san-replace *.apps.${SEED_FULL_DOMAIN},*.apps.${NEW_CLUSTER_FULL_DOMAIN}"
+            recert_cmd="${recert_cmd} ""--cluster-rename ${NEW_CLUSTER_NAME}:${NEW_CLUSTER_BASE_DOMAIN}"
+        fi
         if [[ ${NODE_IP} =~ .*:.* ]]; then
             ETCD_NEW_IP="[${NODE_IP}]"
         else
@@ -51,6 +76,7 @@ function recert {
         sudo podman exec -it recert_etcd bash -c "/usr/bin/etcdctl member list | cut -d',' -f1 | xargs -i etcdctl member update "{}" --peer-urls=http://${ETCD_NEW_IP}:2380"
         sudo podman exec -it recert_etcd bash -c "/usr/bin/etcdctl del /kubernetes.io/configmaps/openshift-etcd/etcd-endpoints"
         find /etc/kubernetes/ -type f -print0 | xargs -0 sed -i "s/${OLD_IP}/${NODE_IP}/g"
+        reconfigure_dnsmasq ${NEW_CLUSTER_FULL_DOMAIN} ${NODE_IP}
     fi
 
     # Use previous cluster certs if directory is present
@@ -169,27 +195,8 @@ until openssl x509 -in ${KUBELET_CLIENT_CERTIFICATE} -checkend 30 &> /dev/null; 
 done
 echo "${KUBELET_CLIENT_CERTIFICATE} is valid."
 
-# Reconfigure DNS
-node_ip=$(oc get nodes -o jsonpath='{.items[0].status.addresses[?(@.type == "InternalIP")].address}')
-domain="$(oc apply -f "${RELOCATION_CONFIG_PATH}" --dry-run=client -o jsonpath='{.items[?(@.kind=="ClusterRelocation")].spec.domain}')"
-
-if [ -z ${domain+x} ]; then
-    echo "domain not defined"
-else
-    echo "Updating dnsmasq with new domain"
-    cat << EOF > /etc/dnsmasq.d/customer-domain.conf
-address=/apps.${domain}/${node_ip}
-address=/api-int.${domain}/${node_ip}
-address=/api.${domain}/${node_ip}
-EOF
-    systemctl restart dnsmasq
-fi
-
 echo "Applying cluster configuration"
 oc apply -f "${RELOCATION_CONFIG_PATH}"
-echo "Waiting for cluster relocation status"
-oc wait --timeout=1h clusterrelocation cluster --for condition=Reconciled=true &> /dev/null
-echo "Cluster configuration updated"
 
 if [ -d ${EXTRA_MANIFESTS_PATH} ]; then
     echo "Applying extra-manifests"
--- a/ibu-imager/seedcreator/seedcreator.go
+++ b/ibu-imager/seedcreator/seedcreator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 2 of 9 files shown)


mresvanis (approved):

utils/utils.go
```diff
@@ -13,3 +16,23 @@ func WriteToFile(data interface{}, filePath string) error {
 	}
 	return os.WriteFile(filePath, marshaled, 0o644)
 }
+
+// RenderTemplateFile render template file
+func RenderTemplateFile(srcTemplate string, params map[string]interface{}, dest string, perm os.FileMode) error {
+	templateData, err := os.ReadFile(srcTemplate)
+	if err != nil {
+		return fmt.Errorf("error occurred while trying to read %s : %e", srcTemplate, err)
```
sudomakeinstall2, line 24: not very important but you can use `errors.Wrap` or `errors.Wrapf`:
`return errors.Wrapf(err, "error occured while trying to read %s", srcTemplate)`

tsorya: I know but there was PR that removed errors from go.mod so i assumed we are using fmt for errors

omertuc: >not very important but you can use errors.Wrap or errors.Wrapf:

Why? we intentionally [got rid](https://github.com/openshift-kni/lifecycle-agent/pull/38) of the `errors` library, which is non-standard and mostly obsolete since Go 1.13 introduced a standard way to wrap errors with `%w` in `fmt.Errorf`

Having said that, the line you commented on has a typo, it should say `%w` and not `%e`

sudomakeinstall2: Oh, I did not know that.

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -398,3 +402,16 @@ func (s *SeedCreator) backupOstreeOrigin(statusRpmOstree *ostree.Status) error {
 	log.Println("Backup of .origin created successfully.")
 	return nil
 }
+
+func (s *SeedCreator) renderInstallationEnvFile(recertContainerImage, seedDomain string) error {
+	// Define source and destination file paths
+	srcFile := "installation_configuration_files/conf/installation-configuration.env"
+	destFile := "/etc/systemd/system/installation-configuration.env"
+	// Prepare variable substitutions for template files
+	substitutions := map[string]interface{}{
```
omertuc, line 411:
```suggestion
	substitutions := map[string]any{
```

utils/utils.go
```diff
@@ -13,3 +16,23 @@ func WriteToFile(data interface{}, filePath string) error {
 	}
 	return os.WriteFile(filePath, marshaled, 0o644)
 }
+
+// RenderTemplateFile render template file
+func RenderTemplateFile(srcTemplate string, params map[string]interface{}, dest string, perm os.FileMode) error {
```
omertuc, line 21:
```suggestion
func RenderTemplateFile(srcTemplate string, params map[string]any, dest string, perm os.FileMode) error {
```

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -25,6 +25,26 @@ fi
 echo "${RELOCATION_CONFIG_PATH} has been found"
```
omertuc, line 25: Unrelated: why is this logic in bash`? we should change it to Go, this is way too much logic to rely on error-prone bash

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -25,6 +25,26 @@ fi
...
 # Replace this with a function that loads values from yaml file
 set +o allexport
+CLUSTER_NAME=$(jq -r '.cluster_name' "${RELOCATION_CONFIG_PATH}"/clusterinfo/manifest.json)
+DOMAIN=$(jq -r '.domain' "${RELOCATION_CONFIG_PATH}"/clusterinfo/manifest.json)
+NEW_DOMAIN="${CLUSTER_NAME}.${DOMAIN}"
```
omertuc, line 30: The entire "original SNO" vs "seed SNO" vs "new sno" trinity is confusing enough, let's make sure we use very explicit names here to avoid confusion, e.g.:

```suggestion
NEW_CLUSTER_NAME=$(jq -r '.cluster_name' "${RELOCATION_CONFIG_PATH}"/clusterinfo/manifest.json)
NEW_CLUSTER_BASE_DOMAIN=$(jq -r '.domain' "${RELOCATION_CONFIG_PATH}"/clusterinfo/manifest.json)
NEW_CLUSTER_FULL_DOMAIN="${NEW_CLUSTER_NAME}.${NEW_CLUSTER_BASE_DOMAIN}"
```

omertuc: Also `SEED_DOMAIN` in other parts of this PR should be `SEED_FULL_DOMAIN` for consistency

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
donpenney: This is the systemd service that gets injected into the seed image. Bash seems appropriate here, doesn't it?

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -169,27 +196,11 @@ until openssl x509 -in ${KUBELET_CLIENT_CERTIFICATE} -checkend 30 &> /dev/null;
...
-fi
-
 echo "Applying cluster configuration"
 oc apply -f "${RELOCATION_CONFIG_PATH}"
-echo "Waiting for cluster relocation status"
-oc wait --timeout=1h clusterrelocation cluster --for condition=Reconciled=true &> /dev/null
-echo "Cluster configuration updated"
+#echo "Waiting for cluster relocation status"
+#oc wait --timeout=1h clusterrelocation cluster --for condition=Reconciled=true &> /dev/null
+#echo "Cluster configuration updated"
```
omertuc, line 203: Should these be just deleted?

utils/utils.go
```diff
@@ -13,3 +16,23 @@ func WriteToFile(data interface{}, filePath string) error {
...
+	templateData, err := os.ReadFile(srcTemplate)
+	if err != nil {
+		return fmt.Errorf("error occurred while trying to read %s : %e", srcTemplate, err)
+	}
+
+	tmpl := template.New("template")
+	tmpl = template.Must(tmpl.Parse(string(templateData)))
+	var buf bytes.Buffer
+	if err = tmpl.Execute(&buf, params); err != nil {
+		return fmt.Errorf("failed to render controller template: %e", err)
```
omertuc, line 31:
```suggestion
		return fmt.Errorf("failed to render controller template: %w", err)
```

utils/utils.go
omertuc, line 24:
```suggestion
		return fmt.Errorf("error occurred while trying to read %s: %w", srcTemplate, err)
```

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
omertuc: Not to me, I prefer we run a go binary in a container

utils/utils.go
donpenney: I wonder if we can add a linter test or something else to ensure we don't inadvertently add the lib back in the future ... @jc-rh ?

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
omertuc: Bash is a maintenance nightmare. Lots of pitfalls, horrible error checking (Even with `set -e`, `pipefail` it completely ignores errors in subshells like `$(do something)`)

omertuc: No logging facilities, no type checking, etc etc. Not suitable for anything other than running a few commands IMO

donpenney: I will agree with you that there are pitfalls, and many poorly-written bash scripts

romfreiman: I think we agreed that we gonna write everything in go

browsell: I do not recall that agreement. We said we would do go where is made sense.

omertuc: Much of the "productization" overhead of using Go instead of Bash, which includes:

* needing to compile
* needing to build an image
* needing to have podman available in the execution environment
* needing to have the image available and mirrored in the execution environment

is already things **we have covered**. We already compile go in this repo. We already build an image and make it available. We already have podman installed in the environment where we run. 

So running this logic in a new go binary from the systemd service, instead of running this bash script, is only a matter of writing `podman run --rm --replace -v some:mounts --entrypoint blabla <lifecycle_image_pullspec>` instead of `./blabla.sh`, which I would say "makes sense".

I don't see the benefit of bash here, the only one being "we already wrote the script", but it's a hastily written script from a poc repo, we're writing a real product here, I don't think we should keep it

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -169,27 +196,11 @@ until openssl x509 -in ${KUBELET_CLIENT_CERTIFICATE} -checkend 30 &> /dev/null;
...
```
donpenney, line 204: We wipe /opt/openshift at the end, after successful completion. Should we also delete the env file (and other injected files) as part of cleanup? Or this could be done as part of the post-reboot completion of the Upgrade stage in LCA.
```
/etc/systemd/system/installation-configuration.env
/etc/systemd/system/installation-configuration.service
/etc/systemd/system/prepare-installation-configuration.service
/usr/local/bin/installation-configuration.sh
/usr/local/bin/prepare-installation-configuration.sh
```

tsorya: We can do it as a next step

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
tsorya: To say the truth the option to run bash without need to pull image is a great plus as for me. Most of the code of this script is taking care of running recert. @omertuc there is an option we could provide params to recert as json file? It that case we could do most of this logic in LCA and just provide file with list of params without need to have large bash script

omertuc: >One day we will move it to go, i really don't think we should change everything together.

I started this thread saying "Unrelated", I didn't mean to imply it should be done as part of this PR.

>there is an option we could provide params to recert as json file

I could add that as an option, sure. Created [ticket](https://issues.redhat.com/browse/MGMT-16150)

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
tsorya: Will open task for cleanup

donpenney (approved):

PR #56, opened 2023-11-06
Add helper script for generating dnsmasq workaround site-policy

donpenney:
Added helper script to generate a section that can be added to an SNO site-policy to override the 50-master-dnsmasq-configuration from the assisted installer, until the necessary changes are available in ACM.

/cc @tsorya @alosadagrande

### DIFF
--- a/hack/generate-dnsmasq-site-policy-section.sh
+++ b/hack/generate-dnsmasq-site-policy-section.sh
@@ -5,13 +5,15 @@ PROG=$(basename "$0")
 
 declare CLUSTER_NAME_DOMAIN=
 declare CLUSTER_IP=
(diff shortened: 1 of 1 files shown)


alosadagrande:
Used the policy successfully so lgtm. I would also create a generate_machineconfig function so you can choose between configure the dnsmasq via ACM policy or just apply the machineconfig as day2 operation.

donpenney:
> Used the policy successfully so lgtm. I would also create a generate_machineconfig function so you can choose between configure the dnsmasq via ACM policy or just apply the machineconfig as day2 operation.

@alosadagrande Added --mc option to generate machine-config instead

jc-rh (approved):

PR #58, opened 2023-11-07
CNF-9461: Precaching Plugin

sakhoury:
This PR comprises of the precaching plugin for the Lifecycle Agent (LCA) operator. 
Note:
- The plugin replaces the PoC pull images script. 
- Precaching is performed via a Kubernetes Job. 
- The list of precaching images are obtained via the IBU (image-based upgrade) seed image.

/cc @browsell @donpenney @jc-rh


internal/precache/constants.go
```diff
@@ -0,0 +1,72 @@
...
+	BackoffLimit    int32               = 1 // Allow 1 automatic retry for failed pre-caching job
+	DefaultMode     int32               = 420
+	RunAsUser       int64               = 0 // Run as root user
+	Privileged      bool                = true
+	HostDirPathType corev1.HostPathType = corev1.HostPathDirectory
+)
+
+// Resource Limits for Job
+const (
+	RequestResourceCPU    string = "1"
```
browsell, line 62: lets make this 10m to start

internal/bindata/prepGetSeedImage.sh
```diff
@@ -5,12 +5,14 @@
 
 declare SEED_IMAGE=
 declare PROGRESS_FILE=
+declare IMAGE_LIST_FILE=
```
browsell, line 8: Do we plan on leaving this as a bash script ?

internal/precache/constants.go
sakhoury: Done

internal/precache/precache-workload/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM registry.hub.docker.com/library/golang:1.20 as builder
```
jc-rh, line 1: There is a red hat one. See the PR from Michail

donpenney, line 1: Please roll this up into the LCA image as a secondary binary with a single vendor folder, rather than having a separate image, as has been done with the ibu-imager.

controllers/prep_handlers.go
```diff
@@ -185,20 +259,38 @@ func (r *ImageBasedUpgradeReconciler) handlePrep(ctx context.Context, ibu *lcav1
...
 
+			// Fetch final precaching job report summary
+			conditionMessage := "Prep completed"
+			status, err := precache.QueryJobStatus(ctx, r.Client)
+			if err == nil && status.Message != "" {
+				conditionMessage += fmt.Sprintf(": %s", status.Message)
+			}
+
+			// Cleanup precaching resources
+			if err = precache.Cleanup(ctx, r.Client); err != nil {
```
donpenney, line 284: I would prefer to see the cleanup done in the handleAbort, so that the logs stick around if they are needed

donpenney (commented):
Please update HACKING.md with info for devs on the new job and how to access logs, and maybe add an overview somewhere for the precaching job setup/model

internal/precache/precache-workload/Dockerfile
sakhoury: Changed to `registry.access.redhat.com/ubi9/go-toolset:1.20`

sakhoury: Done

controllers/prep_handlers.go
sakhoury: Pre-caching cleanup moved to handleAbort

donpenney:
This is a very large list of new vendor packages being pulled in to use the podman library. The added maintenance cost here seems to greatly outweigh the benefit, vs just doing an os.exec call

internal/precache/workload/pullImages.go
```diff
@@ -0,0 +1,200 @@
...
+// podmanImgPull pulls the image via podman
+func podmanImgPull(image, authFile string) error {
+	args := []string{"pull", image}
+	if authFile != "" {
+		args = append(args, []string{"--authfile", authFile}...)
+	}
+	return executeCmd(false, "podman", args...)
+}
+
+// pullImage attempts to pull an image using the Podman API
```
donpenney, line 90: nit: Should probably remove comment references to Podman API

internal/precache/workload/pullImages.go
```diff
@@ -0,0 +1,200 @@
...
+const MaxRetries int = 5
+
+// Podman auth-file related constants
+const (
+	EnvAuthFile     string = "PULL_SECRET_PATH"
+	DefaultAuthFile string = "/var/lib/kubelet/config.json"
+)
+
+// executeCmd execute shell commands
+func executeCmd(suppressCmdOutput bool, cmd string, args ...string) (err error) {
```
donpenney, line 39: We have similar util commands, which have been recently refactored and moved to:
https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/ops/execute.go

Can this be used instead? I guess the "suppression" option would need to be added

internal/precache/workload/pullImages.go
sakhoury: Changed from API -> CLI

internal/precache/workload/pullImages.go
sakhoury: It can be used, however, refactoring the function `func (e *executor) execute(liveLogger io.Writer, root, command string, args ...string) (string, error) ` in `ibu-imager/ops/execute.go` to include command output suppression should probably be done in a separate PR under its own Jira. 

Given that output suppression is not used elsewhere, it may be an unnecessary attempt resulting in extraneous parameter bloating for other use-cases. Since the `executeCmd` function is a simple wrapper to `exec/os.{Command,Run}`, I think it is okay to simply leave it here as is. If the command output suppression is beneficial to other scenarios, then a future refactoring may suffice.

internal/precache/precache.go
```diff
@@ -0,0 +1,203 @@
...
+	"strings"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;delete
+
+var log = ctrl.Log.WithName("precache")
```
pixelsoccupied, line 33: I have noticed that we are including loggers in the main https://github.com/openshift-kni/lifecycle-agent/blob/main/main/main.go#L133-L134

Would that be easy to pattern to convert to? Up to you

internal/precache/precache.go
```diff
@@ -0,0 +1,203 @@
...
+// Supported configuration options in args:
+//   - "NumConcurrentPulls" (int): Number of concurrent pulls for pre-caching.
+//   - "NicePriority" (int): Nice priority for pre-caching.
+//   - "IoNiceClass" (int): I/O nice class for pre-caching.
+//   - "IoNicePriority" (int): I/O nice priority for pre-caching.
+//
+// Example usage:
+//
+//	config := NewConfig(imageList, "NumConcurrentPulls", 10, "NicePriority", 5)
+func NewConfig(imageList []string, args ...interface{}) *Config {
```
pixelsoccupied, line 58: Curious why you're allowing arbitrary args and not something like a struct which would just be type checked etc? 

Generally the pattern i have noticed `func NewConfig(options *PrecacheOptions) *PrecacheConfig`

internal/precache/workload/pullImages.go
```diff
@@ -0,0 +1,200 @@
...
+func pullImage(image, authFile string, progress *precache.Progress) error {
+
+	var err error
+	for i := 0; i < MaxRetries; i++ {
+		err = podmanImgPull(image, authFile)
+		if err == nil {
+			log.Infof("Successfully pulled image: %s", image)
+			break
+		} else {
+			log.Infof("Attempt %d/%d: Failed to pull %s: %v", i+1, MaxRetries, image, err)
```
pixelsoccupied, line 100: Can we benefit from a tiny sleep when there needs to be retry? Apparently each pull is 3x retry by default 
https://github.com/containers/podman/issues/14359 
https://github.com/containers/common/blob/main/libimage/copier.go#L297-L300

internal/precache/workload/pullImages.go
```diff
@@ -0,0 +1,200 @@
...
+	var wg sync.WaitGroup
+	numThreads, err := strconv.Atoi(os.Getenv(precache.EnvMaxPullThreads))
+	if err != nil {
+		numThreads = precache.DefaultMaxConcurrentPulls
+	}
+	threads := make(chan struct{}, numThreads)
+	log.Infof("Configured precaching job to concurrently pull %d images.", numThreads)
+
+	// Start pulling images
+	for _, image := range pullSpec {
```
pixelsoccupied, line 172: nit: maybe pullSpecs (plural)?

controllers/idle_handlers.go
```diff
@@ -67,6 +69,13 @@ func (r *ImageBasedUpgradeReconciler) cleanupUnbootedStateroot(stateroot string)
 
 //nolint:unparam
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	// Cleanup precaching resources
+	r.Log.Info("Cleanup precaching resources")
+	if err := precache.Cleanup(ctx, r.Client); err != nil {
+		r.Log.Error(err, "Failed to cleanup precaching resources")
+		return doNotRequeue(), err
```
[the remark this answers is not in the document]
sakhoury: Shouldn't the function `requeueWithError` be called `doNotRequeueWithError` since an empty `ctrl.Result{}` is returned?

controllers/prep_handlers.go
```diff
@@ -45,39 +46,114 @@ func (r *ImageBasedUpgradeReconciler) launchGetSeedImage(
...
+	switch {
+	case status.Status == "Active":
+		logMsg = "Precaching in-progress"
+		result = requeueWithMediumInterval()
+	case status.Status == "Succeeded":
+		progressContent = []byte("completed-precache")
+		logMsg = "Precaching completed"
+	case status.Status == "Failed":
+		progressContent = []byte("Failed")
+		logMsg = "Precaching failed"
```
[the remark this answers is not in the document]
sakhoury: I had the struct called `PrecacheStatus` before, however, the linter complained about it since the package name is `precache`.

internal/precache/precache.go
sakhoury: Yes! Will follow suit

internal/precache/precache.go
sakhoury: Implemented this as a factory function for creating a new Config instance by allowing the user to specify a subset of arguments without having to specify the entire parameter list. This then allows for default values to be assigned for non-specified parameters.

internal/precache/workload/pullImages.go
sakhoury: I can add a sleep, however, with 15 (5*3) retries, if the pull failed, clearly there is an issue (either with the image spec not being available) or the registry. The reason I avoided sleep was to fail fast as there can be easily over 100 images that need to be precached.

internal/precache/workload/pullImages.go
sakhoury: Spec refers to a specification object, in this context, the specification object is a list of images. The relationship is 1 pullSpec : 1 list of images. So plural would imply multiple lists of images.

donpenney (approved):

PR #59, opened 2023-11-07
CNF-9894: Add extra manifest plugin for exporting and applying manifests

Missxiaoguo:


controllers/upgrade_handlers.go
```diff
@@ -41,6 +48,13 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *ra
 		return ctrl.Result{}, err
 	}
 
+	// After reboot to the new stateroot
```
browsell, line 51: Let's add a TODO here

internal/extramanifest/extramanifest_test.go
```diff
@@ -0,0 +1,184 @@
...
+		Kind:    "SriovNetworkNodePolicyList",
+	})
+
+	err = fakeClient.List(context.Background(), sriovnodepolicies)
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
+
+	if len(sriovnodepolicies.Items) != 2 {
+		t.Errorf("Expected 2 SriovNetworkNodePolicies, got %d", len(sriovnodepolicies.Items))
```
sudomakeinstall2, line 182: not very important but we already have added `testify` package. You can rely on assert for `if` statement and the expect and got messages.

Missxiaoguo: Yeah, right. Somehow I don't know why I didn't use that. Perhaps I thought I only have a few checks.

PR #60, opened 2023-11-07
CNF-10235: [Imager] Add targets for CI job in Makefile and clean up unused scripts

leo8a:
This PR includes changes for:

- Adds a help target to LCA's Makefile and sets it as the default
- Add imager's targets for the `ci-job` in Makefile
- Clean up unused scripts and enhance the existing ones

/cc @donpenney @jc-rh


Makefile
```diff
@@ -114,20 +119,20 @@ unittest:
...
 
 .PHONY: bashate
-bashate: ## Run bashate
+bashate: ## Run bashate.
 	@echo "Running bashate"
 	hack/bashate.sh
 
 .PHONY: ci-job	
-ci-job: common-deps-update generate fmt vet lint golangci-lint unittest shellcheck bashate update-bindata bundle-check #test
+ci-job: common-deps-update generate fmt vet lint golangci-lint unittest shellcheck bashate bundle-check imager-unittest
```
jc-rh, line 135: why is update-bindata removed?

donpenney: The update-bindata is required here as part of checking that generated code has been updated - at least until (and if) we get rid of all the bindata bits and implement them fully in go, so it would need to be restored (and done ahead of "bundle-check")

Makefile
```diff
@@ -263,3 +265,40 @@ catalog-build: opm ## Build a catalog image.
...
+			 go install gotest.tools/gotestsum@latest && \
+			 popd'
+	$(MAKE) imager_test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./ibu-imager/...))"
+
+.PHONY: imager-build
+imager-build: common-deps-update fmt vet ## Build imager binary.
+	go build -o bin/ibu-imager main/ibu-imager/main.go
+
+.PHONY: imager-build-container
+imager-build-container: ## Build container image with the tool.
```
donpenney, line 295: Is the separate image for test purposes?

Dockerfile.imager
```diff
@@ -0,0 +1,30 @@
+########## Builder ##########
+# Build the imager binary
```
tsorya, line 2: I thought we wanted to build only one image?

leo8a: Indeed, we want to / will productize only one image, comprising both LCA and Imager.

The `imager-build-container` target is intended for folks developing / testing **only** the Imager tool, without including the LCA bits in a lighter image.

mresvanis: Another idea would be to have another stage in the lifecycle-agent Dockerfile, e.g. :

```dockerfile
FROM registry.access.redhat.com/ubi9/ubi:latest as imager-dev

RUN dnf -y install jq && \
    dnf clean all && \
    rm -rf /var/cache/dnf

COPY --from=imager /opt/app-root/src/build/ibu-imager /usr/local/bin/ibu-imager
COPY ibu-imager/installation_configuration_files/ /usr/local/bin/installation_configuration_files/

ENTRYPOINT ["/usr/local/bin/ibu-imager"]
```

and build the imager-dev image with something like this:

```shell
podman build . -t quay.io/mresvani/ibu-imager -f Dockerfile --target imager-dev
```

This way, we're only building the imager part and we have a dev image, which is small enough (I guess we could go smaller by using ubi-minimal or perhaps ubi-micro?). WDYT?

leo8a: tbh, I don't have a strong preference here, as long as we get a smaller image for testing / developing Imager, I guess both approaches are OK to me.

(just created a separate Dockerfile cause it seemed kind of a different / independent piece of software)

/deferring to @jc-rh @donpenney on this

jc-rh: I like Michail's idea. Less chance for us to forget to update the standalone file. And yes, we typically use ubi-minimal for pure go app. Not sure if the imager requires anything additional.

donpenney: The LCA was originally using gcr.io/distroless/static:nonroot, and updated to ubi:latest because that's what the ibu-imager was using. So presumably we just need to ensure any dependencies required by the imager are met. Smaller is definitely better, and less chances of being impacted by CVEs, etc

donpenney: If you're looking to make a lighter image, using a smaller base FROM image would be the way to go, I think. The LCA manager binary isn't much bigger than the ibu-imager binary

leo8a: yea, originally I also tried to use `gcr.io/distroless/static:nonroot` in Imager but had some issues with `nsenter` missing in that base. Let's do Michail's idea, thanks y'all.

leo8a:
```
-> podman run --rm gcr.io/distroless/static:nonroot nsenter --version
Error: crun: executable file `nsenter` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found

-> podman run --rm -ti registry.access.redhat.com/ubi9/ubi-micro:latest nsenter --version
Error: crun: executable file `nsenter` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found

-> podman run --rm -ti registry.access.redhat.com/ubi9/ubi-minimal:latest nsenter --version
Error: crun: executable file `nsenter` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found

-> podman run --rm -ti registry.access.redhat.com/ubi9/ubi:latest nsenter --version
nsenter from util-linux 2.37.4
```

donpenney: If you feel you need a separate image, ok. But the only difference between the LCA image and this ibu-imager image is the exclusion of the manager binary, which is currently 53M. So I don't understand what the benefit would be here.

hack/shellcheck.sh
```diff
@@ -28,6 +28,11 @@ else
...
 fi
 
+# Find and check shell script files with ShellCheck
 find . -name '*.sh' -not -path './vendor/*' -not -path './*/vendor/*' -not -path './git/*' \
     -not -path './bin/*' -not -path './testbin/*' -print0 \
     | xargs -0 --no-run-if-empty ${shellcheck} -x
+
+# All checks passed successfully
+echo "All checks passed successfully"
+exit 0
```
donpenney, line 38: You've discarded the exit status of the shellcheck command completely, so it will now always succeed.

donpenney (changes requested):

donpenney (approved):

PR #63, opened 2023-11-07
init upgradehandler (pre-pivot) and cleanup

pixelsoccupied:
- init code for upgrade prepivot
  - Call and backup and extra-manifest as needed and moved some of the functions to upgradeHandler
  - No requeue-ing when an error is caught
- an error msg for when ibu name is not what's expected
- CRs to quickly deploy minIO and other support CRs+doc to work with backup in HACKING.md
- include RBAC for deletebackuprequests

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -64,7 +64,11 @@ apiVersion: lca.openshift.io/v1alpha1
 kind: ImageBasedUpgrade
 metadata:
   name: upgrade
+  namespace: openshift-lifecycle-agent
 spec:
+  oadpContent:
+    - name: oadp-cm
+      namespace: openshift-adp
   stage: Idle
   seedImageRef:
     version: 4.13.9
@@ -145,4 +149,140 @@ perform some checks in the seed SNO cluster.
 
 ```shell
 skopeo copy docker://quay.io/edge-infrastructure/recert docker://${LOCAL_REGISTRY}/edge-infrastructure/recert
+````
+
+## Setup dev backup steps
+
+### minio + oadp oprator
+
+Consider using podman if you have a HV 
+```shell
+podman run --name minio -d -p 9000:9000 -p 9001:9001 quay.io/minio/minio server /data --console-address ":9001" 
+```
+
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: minio-dev
+  labels:
+    name: minio-dev
+
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    app: minio
+  name: minio
+  namespace: minio-dev
+spec:
+  containers:
+  - name: minio
+    image: quay.io/minio/minio:latest
+    command:
+    - /bin/bash
+    - -c
+    args:
+    - minio server /data --console-address :9090
+    volumeMounts:
+    - mountPath: /var/local/data
+      name: localvolume
+  volumes:
+  - name: localvolume
+    hostPath:
+      path: /var/local/data
+      type: DirectoryOrCreate
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: minio
+  namespace: minio-dev
+spec:
+  selector:
+    app: minio
+  ports:
+    - name: two
+      protocol: TCP
+      port: 9090
+    - name: one
+      protocol: TCP
+      port: 9000
+# kubectl port-forward svc/minio 9000 9090 -n minio-dev
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-adp
+  labels:
+    name: openshift-adp
+
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: openshift-adp
+  namespace: openshift-adp
+spec:
+  targetNamespaces:
+  - openshift-adp
+
+---
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: openshift-adp
+  namespace: openshift-adp
+spec:
+  channel: stable-1.2
+  installPlanApproval: Automatic
+  approved: true
+  name: redhat-oadp-operator
+  source: redhat-operators-disconnected
+  sourceNamespace: openshift-marketplace
+```
+
+### setup DataProtectionApplication CR
+```yaml
+# before applying this setup creds for the bucket access + create bucket
+# create bucket: aws s3api create-bucket --bucket test-backups --profile minio
+# create a new called `credentials-velero`
+#
+# [default]
+# aws_access_key_id=testkey
+# aws_secret_access_key=testaccesskey
+#
+# apply: oc create secret generic cloud-credentials -n openshift-adp --from-file cloud=credentials-velero
+
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataProtectionApplication
+metadata:
+  name: dataprotectionapplication
+  namespace: openshift-adp
+spec:
+  configuration:
+    velero:
+      defaultPlugins:
+        - aws
+        - openshift
+      resourceTimeout: 10m
+  backupLocations:
+    - velero:
+        config:
+          profile: "default" # todo find out why it didnt work with a different profile name
+          region: minio
+          s3Url: "http://minio.minio-dev.svc.cluster.local:9000"
+          insecureSkipTLSVerify: "true"
+          s3ForcePathStyle: "true"
+        provider: aws
+        default: true
+        credential:
+          key: cloud
+          name: cloud-credentials
+        objectStorage:
+          bucket: test-backups
+          prefix: velero # a required value if no image provided
 ```
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
+
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -38,7 +40,6 @@ import (
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	"github.com/openshift-kni/lifecycle-agent/ibu-imager/ops"
-	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
 	"github.com/openshift-kni/lifecycle-agent/internal/clusterconfig"
 	"github.com/openshift-kni/lifecycle-agent/internal/extramanifest"
 
@@ -61,6 +62,11 @@ type ImageBasedUpgradeReconciler struct {
 
 func doNotRequeue() ctrl.Result {
 	return ctrl.Result{}
+}
+
+func requeueWithError(err error) (ctrl.Result, error) {
+	// can not be fixed by user during reconcile
+	return ctrl.Result{}, err
 }
 
 //nolint:unused
@@ -117,6 +123,7 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	nextReconcile = doNotRequeue()
 
 	if req.Name != utils.IBUName {
+		r.Log.Error(fmt.Errorf("ibu CR must be named, %s", utils.IBUName), "")
 		return
 	}
 
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -18,39 +18,133 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"strings"
+	"time"
+
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+// simple custom error types to handle backup cases
+type (
+	BackupFailedError     string
+	BackupValidationError string
+)
+
+func (b BackupFailedError) Error() string {
+	return string(b)
+}
+
+func (b BackupValidationError) Error() string {
+	return string(b)
+}
+
+// handleUpgrade orchestrate main upgrade steps and update status as needed
 func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
 
-	r.Executor.Execute("mount", "/sysroot", "-o", "remount,rw")
+	r.Log.Info("Starting handleUpgrade")
+
+	// backup with OADP
+	r.Log.Info("Backup with Oadp operator")
+	reqBackup, backupCRs, err := r.BackupRestore.CheckIfBackupRequested(ibu.Spec.OADPContent, ctx)
+	if err != nil {
+		if k8serrors.IsInvalid(err) {
+			utils.SetStatusCondition(&ibu.Status.Conditions,
+				utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+				utils.ConditionReasons.InProgress,
+				metav1.ConditionTrue,
+				fmt.Sprintf("Invalid backup resource detected in configMap, please update"),
+				ibu.Generation)
+			return requeueWithShortInterval(), nil
+		}
+		r.Log.Error(err, "")
+		return requeueWithError(err)
+	}
+	if reqBackup {
+		r.Log.Info("Backup requested")
+		ctrlResult, err := r.startBackup(ctx, backupCRs)
+		if err != nil {
+			var (
+				backupFailedError     *BackupFailedError
+				backupValidationError *BackupValidationError
+			)
+			switch {
+			case errors.As(err, &backupFailedError):
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+					utils.ConditionReasons.Failed,
+					metav1.ConditionFalse,
+					backupFailedError.Error(),
+					ibu.Generation)
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.GetCompletedConditionType(lcav1alpha1.Stages.Upgrade),
+					utils.ConditionReasons.Failed,
+					metav1.ConditionFalse,
+					"Upgrade failed",
+					ibu.Generation)
+				return ctrlResult, nil
+			case errors.As(err, &backupValidationError):
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+					utils.ConditionReasons.InProgress,
+					metav1.ConditionTrue,
+					backupValidationError.Error(),
+					ibu.Generation)
+				return ctrlResult, nil
+			default:
+				return requeueWithError(err)
+			}
+		}
+		if !ctrlResult.IsZero() {
+			return ctrlResult, nil
+		}
+	}
+
+	if _, err := r.Executor.Execute("mount", "/sysroot", "-o", "remount,rw"); err != nil {
+		return requeueWithError(err)
+	}
 	stateRootRepo := fmt.Sprintf("/host/ostree/deploy/rhcos_%s/var", ibu.Spec.SeedImageRef.Version)
 
-	// TODO: Pre-pivot steps
-	//
-	// TODO: Call r.BackupRestore.ExportRestoresToDir to write restore CRs to the new stateroot.
-	//       If there is invalid type of error returned, we should update the ibu status and
-	//       requeue with short interval.
-	//
-	err := r.ExtraManifest.ExportExtraManifestToDir(ctx, ibu.Spec.ExtraManifests, stateRootRepo)
-	if err != nil {
-		r.Log.Error(err, "Failed to export extra manifests")
-		return ctrl.Result{}, err
+	r.Log.Info("Writing OadpConfiguratio CRs into new stateroot")
+	if err := r.BackupRestore.ExportOadpConfigurationToDir(ctx, stateRootRepo, backuprestore.OadpNs); err != nil {
+		return requeueWithError(err)
 	}
 
+	r.Log.Info("Writing Restore CRs into new stateroot")
+	if err := r.BackupRestore.ExportRestoresToDir(ctx, ibu.Spec.OADPContent, stateRootRepo); err != nil {
+		if k8serrors.IsInvalid(err) {
+			utils.SetStatusCondition(&ibu.Status.Conditions,
+				utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+				utils.ConditionReasons.InProgress,
+				metav1.ConditionTrue,
+				fmt.Sprintf("Invalid restore resource detected in configMap, please update"),
+				ibu.Generation)
+			return requeueWithShortInterval(), nil
+		}
+		return requeueWithError(err)
+	}
+
+	r.Log.Info("Writing extra-manifests into new stateroot")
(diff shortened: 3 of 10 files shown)


pixelsoccupied:
/hold 

plugging in ExportRestoresToDir

HACKING.md
```diff
@@ -140,4 +144,133 @@ perform some checks in the seed SNO cluster.
 
 ```shell
 skopeo copy docker://quay.io/edge-infrastructure/recert docker://${LOCAL_REGISTRY}/edge-infrastructure/recert
+````
+
+## Setup dev backup steps
+
+### minio + oadp oprator
```
Missxiaoguo, line 151: This can be easy as running this command. Perhaps provide this option as well to bring up minio in a container.
`podman run --name minio -d -p 9000:9000 -p 9001:9001 quay.io/minio/minio server /data --console-address ":9001"
`

controllers/upgrade_handlers.go
```diff
@@ -21,36 +21,101 @@ import (
...
+// handleUpgrade orchestrate main upgrade steps
 func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	// log and update status to in-progress
+	r.Log.Info("Starting handleUpgrade")
+	utils.SetStatusCondition(&ibu.Status.Conditions,
+		utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+		utils.ConditionReasons.InProgress,
+		metav1.ConditionTrue,
+		"Upgrade in-progress",
+		ibu.Generation)
```
Missxiaoguo, line 40: This will override the old status everytime it's re-queued and perform uncessary steps which were previously completed. We should check the current IBU status to determine if move to the next step.

controllers/upgrade_handlers.go
```diff
@@ -21,36 +21,101 @@ import (
...
-		return ctrl.Result{}, err
+		msg := "failed to export Restore CRs"
+		r.Log.Error(err, msg)
+		utils.SetStatusCondition(&ibu.Status.Conditions,
+			utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+			utils.ConditionReasons.Failed,
+			metav1.ConditionFalse,
+			fmt.Sprintf("%s: %v", msg, err),
+			ibu.Generation)
+		return requeueWithError(err)
```
Missxiaoguo, line 92: We want to invoke ExportOadpConfigurationToDir() to export the DataProtectionApplication CR and secrets as well.

controllers/upgrade_handlers.go
```diff
@@ -21,36 +21,101 @@ import (
...
 
-	// TODO: Pre-pivot steps
-	//
-	// TODO: Call r.BackupRestore.ExportRestoresToDir to write restore CRs to the new stateroot.
-	//       If there is invalid type of error returned, we should update the ibu status and
-	//       requeue with short interval.
-	//
-	err := r.ExtraManifest.ExportExtraManifestToDir(ctx, ibu.Spec.ExtraManifests, stateRootRepo)
+	r.Log.Info("Writing extra-manifests into new stateroot")
+	if err = r.ExtraManifest.ExportExtraManifestToDir(ctx, ibu.Spec.ExtraManifests, stateRootRepo); err != nil {
```
Missxiaoguo, line 69: As we have done the backups previously, could we then export the restore CRs and OADP configurations prior to exporting extramanifests

controllers/ibu_controller.go
```diff
@@ -99,40 +103,38 @@ func requeueWithCustomInterval(interval time.Duration) ctrl.Result {
...
-	}()
 
-	nextReconcile = doNotRequeue()
+	// default to not requeue
+	nextReconcile := doNotRequeue()
 
 	if req.Name != utils.IBUName {
-		return
+		msg := fmt.Errorf("invalid IBU name, it must be named '%s'", utils.IBUName)
+		return doNotRequeue(), msg
```
jc-rh, line 114: If you return msg as an error, k8s will retry by itself.

controllers/ibu_controller.go
```diff
@@ -99,40 +103,38 @@ func requeueWithCustomInterval(interval time.Duration) ctrl.Result {
...
 	r.Log.Info("Loaded IBU", "name", req.NamespacedName, "version", ibu.GetResourceVersion(), "desired stage", ibu.Spec.Stage)
 
+	// main
 	currentInProgressStage := utils.GetCurrentInProgressStage(ibu)
+
 	if currentInProgressStage != "" {
 		nextReconcile, err = r.handleStage(ctx, ibu, currentInProgressStage)
 		if err != nil {
-			return
+			return requeueWithError(err)
```
jc-rh, line 137: This is just to make it more descriptive? Personally I prefer more concise code with named return vars.

controllers/upgrade_handlers.go
jc-rh: Yes, the stage handlers are not one shot. It can get called repeatedly if waiting for long running background jobs. The initial in progress condition is already set by the main controller when stage transition happens. The handlers can update the reason/message along the way until it reaches the end (status false with either success or failure)

pixelsoccupied: Removed it!

controllers/upgrade_handlers.go
pixelsoccupied: Added ExportOadpConfigurationToDir PTAL

controllers/ibu_controller.go
pixelsoccupied: Updated to return nil

controllers/ibu_controller.go
pixelsoccupied: Yeah agreed named func parameter with no return values are cleaner. But it's only really good for small functions and large functions like our reconcile it was getting pretty tricky to follow without being explicit. 

Let me know if you like me revert it no worries!

HACKING.md
pixelsoccupied: Added doc

controllers/upgrade_handlers.go
pixelsoccupied: Updated

controllers/ibu_controller.go
jc-rh: I see that you also removed the defer block. That was the main reason why I did that way for TALM. There were too many returns with all sorts of requeueAfter values and I didn't want to duplicate the logging lines so many times. This one is not too bad yet. Hope it stays that way :)

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,125 @@ package controllers
...
+
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
+	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+// handleUpgrade orchestrate main upgrade steps
 func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	// log and update status to in-progress
```
Missxiaoguo, line 34: The overall flow is not that correct.  Comments for the high level, 
- The handler could be repeatedly called when waiting for long running process or if error happens. It shouldn't start from beginning everytime it's being called. 
- We shouldn't record every single error in the IBU status. Most of them could be just API call issues due to unstable network.

controllers/ibu_controller.go
```diff
@@ -191,6 +199,12 @@ func (r *ImageBasedUpgradeReconciler) handleStage(ctx context.Context, ibu *lcav
 	case lcav1alpha1.Stages.Rollback:
 		nextReconcile, err = r.handleRollback(ctx, ibu)
 	}
+
```
jc-rh, line 202: If there is any hander error, it gets returned to k8s and any update made here gets discarded. K8s basically calls the reconcile again with the original version of the CR.

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,125 @@ package controllers
...
-	//
-	err := r.ExtraManifest.ExportExtraManifestToDir(ctx, ibu.Spec.ExtraManifests, stateRootRepo)
+	r.Log.Info("Writing OadpConfiguratio CRs into new stateroot")
+	err = r.BackupRestore.ExportOadpConfigurationToDir(ctx, stateRootRepo, backuprestore.OadpNs)
 	if err != nil {
-		r.Log.Error(err, "Failed to export extra manifests")
-		return ctrl.Result{}, err
+		msg := "failed to export OadpConfiguratio CRs"
+		r.Log.Error(err, msg)
+		utils.SetStatusCondition(&ibu.Status.Conditions,
```
jc-rh, line 69: If we want to declare upgrade failure in the conditions, don't return the err back to k8s.
If it's an unexpected/transient error, don't set the conditions and let k8s retry.

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,125 @@ package controllers
...
+// handleUpgrade orchestrate main upgrade steps
 func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	// log and update status to in-progress
+	r.Log.Info("Starting handleUpgrade")
+
+	// Pre-pivot steps
+
+	// backup with OADP
+	r.Log.Info("Starting backup")
+	backupRequeue, backupStatus, err := r.BackupRestore.ReconcileBackup(ctx, ibu.Spec.OADPContent)
```
jc-rh, line 41: I don't see when we will get completed status here unless the spec is empty. @Missxiaoguo should triggerBackup return something if all backup jobs found completed?

controllers/upgrade_handlers.go
jc-rh: Also, if we decide we are done with this stage (no more retry or requeue), both conditions need to be set, whether success or failure.

pixelsoccupied:
/hold 

making some adjustments

controllers/upgrade_handlers.go
Missxiaoguo: Right, the triggerBackup returns BackupCompleted status if all backups are completed.

internal/backuprestore/backup.go
```diff
@@ -423,6 +424,11 @@ func (h *BRHandler) ExportRestoresToDir(ctx context.Context, configMaps []lcav1a
 
 	restores, err := extractRestoreFromConfigmaps(ctx, h.Client, configmaps)
 	if err != nil {
+		// detects validation error with dryrun
+		if !k8serrors.IsNotFound(err) {
+			ErrInvalidCR = err
+			return ErrInvalidCR
+		}
```
Missxiaoguo, line 431:
```suggestion
	if err != nil {
	         return err
         }
```
There are other error types other than notfound and invalid.  For particular invalid error type returned by the API call should be k8serrors.IsInvalid(err).  And in the main handlerUpgrade function, you should check the if it's k8serrors.IsInvalid(err)

internal/backuprestore/backup.go
```diff
@@ -41,27 +41,27 @@ import (
...
+		result = ctrl.Result{}
+	)
 	if len(oadpContent) == 0 {
-		status.Status = BackupFailedValidation
-		status.Message = "No oadp configmap is provided."
-		result.RequeueAfter = 1 * time.Minute
-		return
+		status.Status = BackupCompleted
+		status.Message = "No oadp configmap is provided. Skipping backup"
+		return result, status, nil
```
Missxiaoguo, line 52: Add an info log here to indicate no oadp configmap is provided.  For this case, the upgrade will continue and we don't need to update the IBU status.

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,125 @@ package controllers
...
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("error msg from backup: %s, will requeue after: %s", backupStatus.Message, backupRequeue.RequeueAfter))
+		return backupRequeue, err
+	}
+	switch backupStatus.Status {
+	case backuprestore.BackupCompleted:
+		// only proceed when BackupCompleted
+		r.Log.Info(fmt.Sprintf("Backup completed with msg %s", backupStatus.Message))
+	default:
+		// backup in progress
```
Missxiaoguo, line 51: Don't forget to include the cases BackupFailed(failed upgrade, no requeue), BackupFailedValidation(requeue)

controllers/upgrade_handlers.go
pixelsoccupied: Updated like we discussed offline

controllers/ibu_controller.go
pixelsoccupied: Reverted!

controllers/ibu_controller.go
pixelsoccupied: Reverted

internal/backuprestore/backup.go
pixelsoccupied: Updated to use Invalid

controllers/upgrade_handlers.go
pixelsoccupied: We are not requeue-ing on any errors and i brought back the statues that were previously there

internal/backuprestore/backup.go
pixelsoccupied: no more logging as discussed

controllers/upgrade_handlers.go
```diff
@@ -19,38 +19,134 @@ package controllers
...
+	if err != nil {
+		r.Log.Error(err, "")
+		return doNotRequeue(), nil
+	}
+	if reqBackup {
+		// validate if backup CR is not malformed
+		err := backuprestore.ValidateBackupCRs(ctx, r.BackupRestore, backupCRs)
+		if err != nil {
+			r.Log.Error(err, "")
+			return doNotRequeue(), nil
```
jc-rh, line 87: Are these errors that can be corrected by user? If they are, we would want to requeue with an interval.

pixelsoccupied: Yeah! But yesterday during we return regardless if can be fixed or not?

controllers/upgrade_handlers.go
```diff
@@ -76,5 +172,167 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 		metav1.ConditionFalse,
 		"Upgrade completed",
 		ibu.Generation)
+
 	return doNotRequeue(), nil
 }
+
+func checkIfBackupRequested(restore *backuprestore.BRHandler, content []lcav1alpha1.ConfigMapRef, ctx context.Context) (bool, []*velerov1.Backup, error) {
```
jc-rh, line 179: Does it make more sense to have these functions as part of the backup plugin code?

pixelsoccupied: Yeah! Moved to backup.go

controllers/upgrade_handlers.go
```diff
@@ -62,7 +158,7 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 	// 	 r.Log.Error(err, "Failed to apply extra manifests")
 	//	 return ctrl.Result{}, err
 	// }
-
+	r.Log.Info("Done handleUpgrade")
```
Missxiaoguo, line 161: The IBU status condition for UpgradeCompleted and UpgradeInProgress should be updated to False which indicates the upgrade fails.

controllers/upgrade_handlers.go
```diff
@@ -76,5 +172,146 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
...
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+
+	return nil
+}
+func (r *ImageBasedUpgradeReconciler) CreateNewBackupCr(ctx context.Context, backup *velerov1.Backup) error {
```
Missxiaoguo, line 201: As we talked, this shouldn't be splitted out of the triggerBackup function. If the backup requires ordering, the previous group should be completed prior to applying the second group.

Missxiaoguo: In addition to that, if the upgrade handler is an one-shot, you will need to periodically check for the resulf of the backup CR.  If it's not an one-shot, you will return requeue interval and check the result in the next intervel.

controllers/upgrade_handlers.go
jc-rh: See Angie's comment too. We either
- keep this upgrade alive by requeuing and maybe updating in progress condition message/reason only, or
- give it up by doNotRequeue and updating the conditions to false.

The only exception is an error that can only be corrected by updating IBU CR spec. In that case, we can return doNotRequeue while keeping the upgrade still in progress.

controllers/upgrade_handlers.go
```diff
@@ -76,5 +172,146 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
...
+				return err
+			}
+
+			if existingBackup == nil {
+				// Backup has not been applied. Applying backup
+				err := r.CreateNewBackupCr(ctx, backup)
+				if err != nil {
+					return err
+				}
+			}
```
pixelsoccupied, line 195: call GetBackupCtrlResult from here

internal/backuprestore/backup.go
```diff
@@ -295,17 +121,10 @@ func extractBackupFromConfigmaps(ctx context.Context, c client.Client, configmap
...
 				continue
 			}
 
-			// Create the backup CR in dry-run mode to detect any validation errors in the CR
-			// i.e., missing required fields
-			err = c.Create(ctx, &resource, &client.CreateOptions{DryRun: []string{metav1.DryRunAll}})
-			if err != nil {
-				return nil, err
-			}
-
```
Missxiaoguo, line 308: Forget to mention. You need to keep this dryrun here for the "unstructured" resource type. With the CR marshalled into velerov1.backup type, the missing required fields will not be catched as they will be filled with empty string.

internal/backuprestore/backup.go
```diff
@@ -35,218 +34,45 @@ import (
...
-	return
-}
-
-func equivalentBackups(backup1, backup2 *velerov1.Backup) bool {
+func EquivalentBackups(backup1, backup2 *velerov1.Backup) bool {
 	// Compare the specs only
 	return equality.Semantic.DeepEqual(backup1.Spec, backup2.Spec)
 }
 
-func (h *BRHandler) checkVeleroBackupProcessStatus(backup *velerov1.Backup) BackupPhase {
```
Missxiaoguo, line 216: Prefer you keep this function as-is. Have the log added in this function

controllers/upgrade_handlers.go
pixelsoccupied: Updated

internal/backuprestore/backup.go
pixelsoccupied: Updated

internal/backuprestore/backup.go
pixelsoccupied: Reverted

controllers/upgrade_handlers.go
pixelsoccupied: Reverted + updated as discussed

controllers/upgrade_handlers.go
pixelsoccupied: Updated as discussed. Following the more traditional path

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,104 @@ package controllers
...
 func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	r.Log.Info("Starting handleUpgrade")
+
+	// backup with OADP
+	r.Log.Info("Backup with Oadp operator")
+	reqBackup, backupCRs, err := r.BackupRestore.CheckIfBackupRequested(ibu.Spec.OADPContent, ctx)
+	if err != nil {
+		if k8serrors.IsInvalid(err) {
+			utils.SetStatusCondition(&ibu.Status.Conditions,
+				utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
```
Missxiaoguo, line 52: I think the reason should be InProgress

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,104 @@ package controllers
...
+	if reqBackup {
+		r.Log.Info("Backup requested")
+		ctrlResult, err := r.handleBackup(ctx, backupCRs)
+		if err != nil {
+			switch {
+			case errors.Is(err, BackupFailed):
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+					utils.ConditionReasons.Failed,
+					metav1.ConditionFalse,
```
Missxiaoguo, line 70: The status should be True. And we also need to update the UpgradeCompleted condition.

controllers/upgrade_handlers.go
```diff
@@ -18,39 +18,104 @@ package controllers
...
+					"Upgrade failed",
+					ibu.Generation)
+				return doNotRequeue(), nil
+			case errors.Is(err, BackupValidationFailedShortRequeue):
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
+					utils.ConditionReasons.Failed,
```
Missxiaoguo, line 77: I think the reason should be InProgress

internal/backuprestore/backup.go
```diff
@@ -35,185 +34,125 @@ import (
...
+// Constants for backup phase
+const (
+	BackupPending          BackupPhase = "BackupPending"
+	BackupFailedValidation BackupPhase = "BackupFailedValidation"
+	BackupFailed           BackupPhase = "BackupFailed"
+	BackupCompleted        BackupPhase = "BackupCompleted"
+	BackupInProgress       BackupPhase = "BackupInProgress"
+	BackupSkipped          BackupPhase = "BackupSkipped"
+	BackupRequested        BackupPhase = "BackupRequested"
+)
```
Missxiaoguo, line 61: As we are not returning a status, could these be removed?

controllers/upgrade_handlers.go
```diff
@@ -76,5 +141,49 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
...
+			msg := fmt.Sprintf("Failed backups: %s", strings.Join(backupTracker.FailedBackups, ","))
+			BackupFailed = fmt.Errorf("%s: %s", BackupFailed.Error(), msg)
+			r.Log.Error(BackupFailed, "")
+			return doNotRequeue(), BackupFailed
+		} else if len(backupTracker.ProgressingBackups) != 0 {
+			r.Log.Info(fmt.Sprintf("Inprogress backups: %s", strings.Join(backupTracker.ProgressingBackups, ",")))
+			return requeueWithCustomInterval(2 * time.Second), nil
+		} else if len(backupTracker.FailedValidationBackups) != 0 {
+			msg := fmt.Sprintf("Failed validation backups reported by Oadp: %s. %s", strings.Join(backupTracker.FailedValidationBackups, ","), "Please update the invalid backups.")
+			BackupValidationFailedShortRequeue = fmt.Errorf(msg)
```
Missxiaoguo, line 176: This is defined as gobal variable on the top. Will that be overriden?

controllers/upgrade_handlers.go
Missxiaoguo: Please ignore this comment. The status should be reflected to the condition type.

controllers/upgrade_handlers.go
Missxiaoguo: Please ignore this comment. The status should be reflected to the condition type so it should be False

controllers/upgrade_handlers.go
pixelsoccupied: Updated

controllers/upgrade_handlers.go
pixelsoccupied: Updated

internal/backuprestore/backup.go
pixelsoccupied: Removed the custom status and instead using velvero status.phase

controllers/upgrade_handlers.go
pixelsoccupied: Updated implementation to make it clearer

pixelsoccupied:
We will need some tests for this as there some edge cases which we might miss in case minor/major refactoring. Maybe kuttle/e2e test will benefit from this? I'll do that with another PR though. wdyt @jc-rh @Missxiaoguo

jc-rh:
> We will need some tests for this as there some edge cases which we might miss in case minor/major refactoring. Maybe kuttle/e2e test will benefit from this? I'll do that with another PR though. wdyt @jc-rh @Missxiaoguo

kuttl would be a bit difficult. Similar to TALM ones where we use scripts to update ACM CR status, we would need to fake OADP behavior. Might want to try UT first.

PR #65, opened 2023-11-08
Move to RH go-toolset 1.20 base images

mresvanis:
This PR replaces the dockerhub Go 1.20 base images with the [RH go-toolset 1.20](https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690?architecture=amd64&image=654387c17ea05ac1dd92a58a). It also adds source code layers explicitly, in order to leverage caching where possible.

Related PRs: 
- https://github.com/openshift-kni/lifecycle-agent/pull/39
- https://github.com/openshift-kni/lifecycle-agent/pull/43

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,45 @@
 # Build the manager binary
-FROM registry.hub.docker.com/library/golang:1.20 as builder
-
-WORKDIR /workspace/
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20 as builder
 
 # Bring in the go dependencies before anything else so we can take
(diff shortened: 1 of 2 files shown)


mresvanis:
/hold
Perhaps we need to check whether the issue with podman builds and user permissions needs to be address before merging this.

mresvanis:
> /hold Perhaps we need to check whether the issue with podman builds and user permissions needs to be addressed before merging this.

/unhold I have adjusted the LCA Dockerfile to the go-toolset / ubi base image, so there should be no permissions issues.

PR #68, opened 2023-11-08
Update golanci-lint and shellcheck scripts to check tool versions

donpenney:


pixelsoccupied:
/lgtm

btw do we have any plans for removing hack/lint.sh? The project has been abandoned for a while now https://github.com/golang/lint. IMO we can enable some stuff in golanci-lint if there's anything missing from golint

donpenney:
> /lgtm
> 
> btw do we have any plans for removing hack/lint.sh? The project has been abandoned for a while now https://github.com/golang/lint. IMO we can enable some stuff in golanci-lint if there's anything missing from golint

Sure, that's fine. We need to upversion the linters, as well, and it would probably be a valuable effort to see what golangci-lint plugins we'd want to enable

PR #69, opened 2023-11-09
MGMT-16155: Copy  cluster-wide proxy from original SNO to upgraded one

mresvanis:
This change is part of the cluster reconfiguration, which takes place after spinning up the seed image onto the target SNO. Cluster reconfiguration supports a predefined set of the original SNO cluster's configuration, e.g. IDMS, pull-secret, network configuration.

Specifically, this change addresses fetcing the [OCP cluster-wide proxy ](https://docs.openshift.com/container-platform/4.14/networking/enable-cluster-wide-proxy.html)  of the original SNO and creating the respective JSON manifest to be [applied](https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/installation_configuration_files/scripts/installation-configuration.sh#L199) alongside the rest of the cluster reconfiguration manifests.
    
> The canonical name for the Proxy object is `cluster`. As per the respective API [doc](https://github.com/openshift/api/blob/master/config/v1/types_proxy.go#L11).

### DIFF
--- a/internal/clusterconfig/clusterconfig.go
+++ b/internal/clusterconfig/clusterconfig.go
@@ -22,13 +22,16 @@ import (
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=imagedigestmirrorsets,verbs=get;list;watch
+// +kubebuilder:rbac:groups=config.openshift.io,resources=proxies,verbs=get;list;watch
 
 const (
+	proxyName                     = "cluster"
 	pullSecretName                = "pull-secret"
 	configNamespace               = "openshift-config"
 	upgradeConfigurationNamespace = "upgrade-configuration"
 	clusterConfigDir              = "/opt/openshift/cluster-configuration"
 	manifestDir                   = "manifests"
+	proxyFileName                 = "proxy.json"
 	clusterIDFileName             = "cluster-id-override.json"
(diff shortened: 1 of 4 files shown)


omertuc:
What's the flow of information here? Is the proxy copied from the original SNO?

mresvanis:
> What's the flow of information here? Is the proxy copied from the original SNO?

Yes, the idea is to bring the proxy configuration of the original SNO to the upgraded one.

donpenney (approved):

mresvanis:
@tsorya thanks for LGTM-ing! I had to rebase the PR branch, could you check it out and LGTM again?

internal/clusterconfig/clusterconfig.go
```diff
@@ -145,6 +160,17 @@ func (r *UpgradeClusterConfigGather) writeSecretToFile(secret *corev1.Secret, fi
 	return utils.WriteToFile(secret, filePath)
 }
 
+func (r *UpgradeClusterConfigGather) writeProxyToFile(proxy *v1.Proxy, filePath string) error {
+	r.Log.Info("Writing proxy to file", "path", filePath)
+	p := &v1.Proxy{
```
tsorya, line 165: why not to remove metadata from original proxy with typeMetaForObject?

mresvanis: I didn't go with the `typeMetaForObject` method, because I wanted to leave out the [ObjectMeta](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L111) of the original object, while IIUC the former replaces the object's [TypeMeta](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L42). WDYT?

PR #73, opened 2023-11-10
Recertify multus and ovn static dirs

mresvanis:
This PR adds the `multus` and `OVN` directories for certificate regeneration when running recert. Without the regeneration of these certs, an OpenShift 4.14 cluster will be stuck trying to create containers because networking will be broken (due to OVN and multus authentication errors). 

The respective directories added for recertification are:

- `/var/lib/ovn-ic`
- `/etc/cni/multus`

In OpenShift 4.14, `multus` and `OVN` have added per-node certificates. For more info please check [here](https://github.com/openshift/multus-cni/commit/30558b5e68d3fd316fefa35fa7d258b91773f8a0) and [here](https://github.com/openshift/ovn-kubernetes/commit/7dc48048a346f175772fe5c2839b083f7a20c597).


mresvanis:
We don't need to recertify those 2 components as we just delete the previous node object.

PR #75, opened 2023-11-10
Fixes for 4.14

tsorya:
Fixes for 4.14:
1. Removing node while creating seed image
2. Filtering multus and ovn certs from seed imag

### DIFF
--- a/ibu-imager/seedcreator/seedcreator.go
+++ b/ibu-imager/seedcreator/seedcreator.go
@@ -14,6 +14,7 @@ import (
 	v1 "github.com/openshift/api/config/v1"
 	cp "github.com/otiai10/copy"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


ibu-imager/seedcreator/seedcreator.go
```diff
@@ -502,3 +511,35 @@ func (s *SeedCreator) renderInstallationEnvFile(recertContainerImage, seedFullDo
 
 	return utils.RenderTemplateFile(srcFile, substitutions, destFile, 0o644)
 }
+
+func (s *SeedCreator) deleteNode(ctx context.Context) error {
+	doneFile := "/var/tmp/node_deletion.done"
```
omertuc, line 516: This is not needed, the function is already idempotent without it

tsorya: Not fully, there can be a case where we already brought down kube-api and something else failed, example 
"delete all containers" and in this case this one will fail

donpenney (approved):

PR #77, opened 2023-11-14
Add function to check if stateroot is booted

sudomakeinstall2:
- Complete the `isRollbackAllowed` logic
- Add interface for RPMOSTreeClient in order to be able to mock it in tests
- Add uber-go/mock package
- Add genmock to Makefile to generate mock files


sudomakeinstall2:
/hold
wait till ostree query is fixed

controllers/ibu_controller.go
```diff
@@ -219,14 +225,34 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 	return
 }
 
-func isRollbackAllowed(ibu *lcav1alpha1.ImageBasedUpgrade) bool {
+var isAfterPivot = func(stateroot string) (bool, error) {
```
donpenney, line 228: Suggest renaming to something like runningUpgradedRelease, with no parameters. There should be no need to pass in the stateroot here, as it could use the ```ibu.Spec.SeedImageRef.Version``` internally. It's either running that release or it's not.
Alternatively, if you wanted to be able to pass in different stateroots for some reason, rename to ```isStaterootBooted```, since that's really what this is saying here.

donpenney: Why the "var" syntax here? This function would be a useful utility function elsewhere, for the Upgrade handler for example. I'm not sure what the purpose of the var here is

controllers/ibu_controller.go
```diff
@@ -142,7 +142,12 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	if isTransitionRequested(ibu) {
-		if validateStageTransition(ibu) {
+		var isAfterPivot bool
+		isAfterPivot, err = r.RPMOstreeClient.IsStaterootBooted(getStaterootName(ibu.Spec.SeedImageRef.Version))
```
sudomakeinstall2, line 146: @donpenney I renamed the function and moved it to `rpmOstreeClient`. Do you think the variable name is okay?

sudomakeinstall2:
rebased
@tsorya @donpenney @jc-rh

ibu-imager/ostreeclient/rpmostreeclient.go
```diff
@@ -112,3 +119,37 @@ func (c *Client) QueryStatus() (*Status, error) {
...
+
+	for _, deployment := range status.Deployments {
+		if deployment.Booted && deployment.OSName == stateroot {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+type FakeClient struct {
```
tsorya, line 138: Can we move to use mocks please? Ostree client should implement interface and interface can be mocked easily with https://github.com/golang/mock. It will be much more generic than creating fake struct per each object.

tsorya: As here https://github.com/leo8a/ibu-imager/pull/12/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R51

controllers/ibu_controller.go
```diff
@@ -142,7 +142,12 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		if err != nil {
+			return
+		}
```
pixelsoccupied, line 149: What do you think moving this check into the `isRollbackAllowed`? Looks like a precondition of that func and reads better instead of passing the bool all the down there?

ibu-imager/ostreeclient/rpmostreeclient.go
pixelsoccupied: Agreed..also If possible consider updating the mod file to https://github.com/uber-go/mock since what we have currently is archived

controllers/ibu_controller.go
sudomakeinstall2: It was like that at first. the problem is `validateStageTransition` and `isRollbackAllowed` don't have access `RPMOStreeClient` or `ImageBasedUpgradeReconciler`. So we can pass this bool or we can pass down the `RPMOstreeClient` or we can make these methods of `ImageBasedUpgradeReconciler`. I thought passing down bool is simpler. Also still open to suggestion.

sudomakeinstall2: other con of moving this `isRollbackAllowed` is we have to propagate the error back to here. Currently `ValidateStageTransition` and `isRollbackAllowed` don't return any errors.

sudomakeinstall2: one other con is it makes the unit tests for these two function more complicated, since we have to create a Reconciler to test them, which they don't actually need it.

pixelsoccupied: Ah yeah too much refactoring. 

What about adding it in `handleRollback`....which has access to client? Seems like the handlers are meant to check the state of the system (unlike controller funcs...which are really just checking the .status.condtion)....wdyt?

sudomakeinstall2: if we add it in `handleRollabck` we probably have to go RollbackFailed condition if it is not after pivot.
If we keep it here we go to `InvalidTransition`.
I don't have a preference regarding this.
What do you think @jc-rh ?

ibu-imager/ostreeclient/rpmostreeclient.go
```diff
@@ -56,6 +56,13 @@ type Deployment struct {
 	Unlocked                *string  `json:"unlocked"`
 }
 
+type IClient interface {
```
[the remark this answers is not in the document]
sudomakeinstall2: AFAIK It was first added for imager. And it is probably temporary until the original rpm-ostreeclient package is usable with chroot.

controllers/ibu_controller.go
pixelsoccupied: Gotcha. Totally your call and I wont hold up the PR for sure as this works well! 

Basically I'm just not too sure if makes sense to call `IsStaterootBooted` regardless of which state we are since it's only relevant for rollback.

sudomakeinstall2: I agree. It is an extra query whenever there is a transition. Although transitions happens two or three times for every upgrade. So time efficiency will not be an issue.
It is a trade off between giving those functions internal access to the `reconciler` and two or three extra queries for every upgrade.
I am still not sure which is better.

jc-rh (approved):

PR #83, opened 2023-11-14
Add hostPID to LCA pod

sudomakeinstall2:
setting `hostPID: true` in the pod allows us to use `nsenter` within the pod


donpenney:
Suggest for the commit message, it would be clearer to state something like:
```
Setting hostPID=true allows access to the pid namespace for nsenter, as well as systemd queries and operations (which are required for rpm-ostree status query).
```

jc-rh (approved):

PR #85, opened 2023-11-14
Save IBU CR to the new installation

jc-rh:


ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -225,5 +225,10 @@ until openssl x509 -in ${KUBELET_SERVING_CERTIFICATE} -checkend 30 &> /dev/null;
 done
 echo "${KUBELET_SERVING_CERTIFICATE} is valid."
 
+IBU_FILE=/var/ibu.yaml
+if oc apply -f ${IBU_FILE}; then
```
jc-rh, line 229: @omertuc I need to replace this with a curl post. Do we already have the cacert and token for accessing api server in this script?

omertuc: You can grab all you need from the kubeconfig

omertuc: But it's fine keeping it `oc` since that's guaranteed to be available in this context and used by many other OCP scripts

tsorya: Why we should have ibu to be different than any other manifest?
Please add this logic to clusterconfig.go as it is part of it

controllers/upgrade_handlers.go
```diff
@@ -53,6 +58,10 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 		return ctrl.Result{}, err
 	}
 
+	// Save the IBU CR to the new state root before pivot
+	if err := saveIBU(ibu, stateRootRepo); err != nil {
```
tsorya, line 62: Why we should have ibu to be different than any other manifest?
Please add this logic to clusterconfig.go as it is part of it

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
tsorya: We already apply all manifests from cluster-configuration folder, there is no need to create new way for the same logic

jc-rh: I'd like to preserve the CR status. Do you have an example somewhere that extracts things from kubeconfig?

omertuc:
```
API_PATH=api/v1/pods
curl \
               $(oc config view --raw -ojson | jq '.clusters[0].cluster.server' -r)/$API_PATH \
    --cacert   <(oc config view --raw -ojson | jq '.clusters[0].cluster."certificate-authority-data"' -r | base64 -d) \
    --cert     <(oc config view --raw -ojson | jq '.users[0].user."client-certificate-data"' -r | base64 -d) \
    --key      <(oc config view --raw -ojson | jq '.users[0].user."client-key-data"' -r | base64 -d)
```

omertuc: This is just an example, we should be more robust with `.current-context` etc

tsorya: Wondering if we actually need status? Seems so many code, i assume i would like to understand why status is actually required as it will have really small amount of data and "verify after pivot" state can be done in code and i think there is already PR for it

tsorya: I mean this code doesn't worth this status or i miss something?
` conditions:
    - lastTransitionTime: "2023-11-15T14:52:50Z"
      message: In progress
      observedGeneration: 10
      reason: InProgress
      status: "False"
      type: Idle
    - lastTransitionTime: "2023-11-15T14:54:21Z"
      message: Prep completed
      observedGeneration: 10
      reason: Completed
      status: "True"
      type: PrepCompleted
    - lastTransitionTime: "2023-11-15T14:54:21Z"
      message: Prep completed
      observedGeneration: 10
      reason: Completed
      status: "False"
      type: PrepInProgress
    - lastTransitionTime: "2023-11-15T14:56:30Z"
      message: Upgrade completed
      observedGeneration: 11
      reason: Completed
      status: "True"
      type: UpgradeCompleted
    - lastTransitionTime: "2023-11-15T14:56:30Z"
      message: Upgrade completed
      observedGeneration: 11
      reason: Completed
      status: "False"
      type: UpgradeInProgress
    observedGeneration: 11
`

omertuc: This is only a lot of code because we're still using bash, this should be Go, then it's a matter of creating a simple kubernetes client which automatically loads the kubeconfig

tsorya: Yeah but this will require moving to go and this is actually i really hoped we will not need. 
Trying to understand why we need this status at all.
I got status from upgrade cr and this is exactly the one we want to preserve. Trying to understand why it is needed

tsorya: If we need this status we should move to golang cause this code is not nice

controllers/upgrade_handlers.go
jc-rh: clusterconfig is called when we enter the upgrade stage. This one is meant to be done right before the pivot.

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
jc-rh: Feel free to create a story for converting the whole thing to go.

romfreiman: Go indeed. Lets create a jira ticket

controllers/upgrade_handlers.go
jc-rh: Also, we are saving the CR that the current reconcile call is working on. There may be status changes comparing to what you would get from a get API call.

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
jc-rh: https://issues.redhat.com/browse/CNF-10403

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -114,4 +114,36 @@ if [ -d ${EXTRA_MANIFESTS_PATH} ]; then
...
+
+    # Cluster CA certs
+    K8S_API_CA_CERTS=$(jq -r '."certificate-authority-data" | @base64d' <<< $K8S_KUBECONFIG_CLUSTER)
+
+    # User credentials
+    K8S_API_CLIENT_CERT=$(jq -r '."client-certificate-data" | @base64d' <<< $K8S_KUBECONFIG_USER)
+    K8S_API_CLIENT_CERT_KEY=$(jq -r '."client-key-data" | @base64d' <<< $K8S_KUBECONFIG_USER)
+
+    if oc apply -f ${IBU_FILE}; then
+        curl -X PATCH "$K8S_API_ENDPOINT"/apis/lca.openshift.io/v1alpha1/namespaces/openshift-lifecycle-agent/imagebasedupgrades/upgrade/status -H "Content-Type: application/merge-patch+json" -d @${IBU_FILE} --cacert <(cat <<< "$K8S_API_CA_CERTS") --cert <(cat <<< "$K8S_API_CLIENT_CERT") --key <(cat <<< "$K8S_API_CLIENT_CERT_KEY")
```
omertuc, line 134: Wouldn't that only update the status? Shouldn't we apply the entire CR?

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -114,4 +114,36 @@ if [ -d ${EXTRA_MANIFESTS_PATH} ]; then
...
+
+    # Extract the server
+    K8S_API_ENDPOINT="$(jq -r '.server' <<< $K8S_KUBECONFIG_CLUSTER)"
+
+    # Cluster CA certs
+    K8S_API_CA_CERTS=$(jq -r '."certificate-authority-data" | @base64d' <<< $K8S_KUBECONFIG_CLUSTER)
+
+    # User credentials
+    K8S_API_CLIENT_CERT=$(jq -r '."client-certificate-data" | @base64d' <<< $K8S_KUBECONFIG_USER)
+    K8S_API_CLIENT_CERT_KEY=$(jq -r '."client-key-data" | @base64d' <<< $K8S_KUBECONFIG_USER)
```
omertuc, line 131: Let's move this to a load_kubeconfig function because it's a lot of noise for `restore_IBU`

ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
jc-rh: `oc apply` creates the CR. The curl command restores the status.

omertuc:
Now that we have #101 we should rewrite this PR

jc-rh:
> Now that we'll have #101 we should rewrite this PR

Do you know a way in go to do what the curl+patch does? Normal status update through API regenerates the transition timestamps.

omertuc:
> > Now that we'll have #101 we should rewrite this PR
> 
> @omertuc Do you know a way in go to do what the curl+patch does? Normal status update through API regenerates the transition timestamps in conditions.

In Go we should do whatever the operator is doing for status updates

controllers/utils/constants.go
```diff
@@ -8,5 +8,6 @@ const (
 	PrepPullImages     string = "prepPullImages.sh"
 	PrepSetupStateroot string = "prepSetupStateroot.sh"
 	PrepCleanup        string = "prepCleanup.sh"
-	IBUName                   = "upgrade"
+	IBUName            string = "upgrade"
+	IBUFilePath        string = "/opt/ibu.json"
```
donpenney, line 12: nit: This doesn't seem like the right path for this, writing the file to the top-level /opt directory. Why not reuse /var/ibu directory, ie. the IBUWorkspacePath constant?

jc-rh: This is for the B side. Do we have anything to carry forward to the B side under /var/ibu?

PR #87, opened 2023-11-15
Create recert configuration file in LCA 

tsorya:
Create recert configuration file in LCA: 
1. Adding host  ip to clusterInfo in order to use it for recert config
2. Moving recert configuration creation to LCA in order to remove this logic from installation script that will only run recert without decision making
3. LCA will create recert_config.json that will be provided to recert command
4. Installation script cleanup and impromevents
5. Moving get node ip to kube command

### DIFF
--- a/bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
+++ b/bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
@@ -82,6 +82,14 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - nodes
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
+          resources:
           - secrets
           verbs:
           - create
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -32,6 +32,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - create
--- a/ibu-imager/clusterinfo/clusterinfo.go
+++ b/ibu-imager/clusterinfo/clusterinfo.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/openshift-kni/lifecycle-agent/utils"
 
 	v1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -25,6 +26,7 @@ type ClusterInfo struct {
 	Domain      string `json:"domain,omitempty"`
 	ClusterName string `json:"cluster_name,omitempty"`
 	ClusterID   string `json:"cluster_id,omitempty"`
+	MasterIP    string `json:"master_ip,omitempty"`
 }
 
 type installConfigMetadata struct {
@@ -54,16 +56,38 @@ func (m *InfoClient) CreateClusterInfo(ctx context.Context) (*ClusterInfo, error
 	if err := m.client.Get(ctx, types.NamespacedName{Name: "version"}, clusterVersion); err != nil {
 		return nil, err
 	}
+
 	installConfig, err := m.getInstallConfig(ctx)
 	if err != nil {
 		return nil, err
 	}
+
+	ip, err := m.getNodeInternalIP(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	return &ClusterInfo{
 		ClusterName: installConfig.Metadata.Name,
 		Domain:      installConfig.BaseDomain,
 		Version:     clusterVersion.Status.Desired.Version,
 		ClusterID:   string(clusterVersion.Spec.ClusterID),
+		MasterIP:    ip,
 	}, nil
+}
+
+func (m *InfoClient) getNodeInternalIP(ctx context.Context) (string, error) {
+	node, err := utils.GetSNOMasterNode(ctx, m.client)
+	if err != nil {
+		return "", err
+	}
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == corev1.NodeInternalIP {
+			return addr.Address, nil
+		}
+	}
+
+	return "", fmt.Errorf("failed to find node internal ip address")
 }
 
 func (m *InfoClient) getInstallConfig(ctx context.Context) (*basicInstallConfig, error) {
--- a/ibu-imager/clusterinfo/clusterinfo.go
+++ b/ibu-imager/clusterinfo/clusterinfo.go
@@ -77,6 +77,7 @@ func (m *InfoClient) CreateClusterInfo(ctx context.Context) (*ClusterInfo, error
 	}, nil
 }
 
+// TODO: add dual stuck support
 func (m *InfoClient) getNodeInternalIP(ctx context.Context) (string, error) {
 	node, err := utils.GetSNOMasterNode(ctx, m.client)
 	if err != nil {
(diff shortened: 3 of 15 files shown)


bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
```diff
@@ -79,6 +79,14 @@ spec:
           - namespaces
           verbs:
           - delete
+        - apiGroups:
```
omertuc, line 82: .gitattributes

config/rbac/role.yaml
```diff
@@ -29,6 +29,14 @@ rules:
   - namespaces
   verbs:
   - delete
+- apiGroups:
```
omertuc, line 32: .gitattributes

ibu-imager/clusterinfo/clusterinfo.go
```diff
@@ -54,18 +57,40 @@ func (m *InfoClient) CreateClusterInfo(ctx context.Context) (*ClusterInfo, error
...
 	}
+
 	installConfig, err := m.getInstallConfig(ctx)
 	if err != nil {
 		return nil, err
 	}
+
+	ip, err := m.getNodeInternalIP(ctx)
+	if err != nil {
+		return nil, err
```
omertuc, line 68: Errors should be wrapped

tsorya: Hmm, you think we should wrap each error? with some better string? 
The question is where better to wrap it inside caller or callee

ibu-imager/clusterinfo/clusterinfo.go
```diff
@@ -54,18 +57,40 @@ func (m *InfoClient) CreateClusterInfo(ctx context.Context) (*ClusterInfo, error
...
+		MasterIP:    ip,
 	}, nil
 }
 
+func (m *InfoClient) getNodeInternalIP(ctx context.Context) (string, error) {
+	node, err := utils.GetSNOMasterNode(ctx, m.client)
+	if err != nil {
+		return "", err
+	}
+	for _, addr := range node.Status.Addresses {
```
omertuc, line 85: what about dualstack? if we don't support let's error on len(node.Status.Addresses)

ibu-imager/clusterinfo/clusterinfo.go
omertuc: I think it should be wrapped everywhere to give context, otherwise you get a vague error and you have no idea where it came from

ibu-imager/clusterinfo/clusterinfo.go
tsorya: We will support, just need to have such system up and running and i will handle it in next prs

omertuc: so for now let's error

tsorya: This will require going through all the Ips and getting the list and checking that list has more than one value. Not sure it is worth now. I will open ticket for dual stuck. Currently this code is only for our internal use so it should be fine.
I will add TODO for it though

internal/clusterconfig/recert.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	config := recertConfig{
+		DryRun:           false,
+		EtcdEndpoint:     "localhost:2379",
+		ExtendExpiration: true,
+		StaticDirs:       []string{"/kubelet", "/kubernetes", "/machine-config-daemon"},
+		StaticFiles:      []string{"/host-etc/mcs-machine-config-content.json"},
+		ClusterRename:    fmt.Sprintf("%s:%s", clusterData.ClusterName, clusterData.Domain),
+		SummaryFile:      "/var/tmp/recert-summary.yaml",
```
omertuc, line 48: Is `/var/tmp` mounted?

omertuc: We also need a clean summary file, so we can include it in must-gather

internal/clusterconfig/recert.go
```diff
@@ -0,0 +1,92 @@
...
+	clusterFullDomain := fmt.Sprintf("%s.%s", clusterData.ClusterName, clusterData.Domain)
+
+	config.CNSanReplaceRules = []string{
+		fmt.Sprintf("%s,%s", seedClusterInfo.MasterIP, clusterData.MasterIP),
+		fmt.Sprintf("api.%s,api.%s", seedFullDomain, clusterFullDomain),
+		fmt.Sprintf("api-int.%s,api-int.%s", seedFullDomain, clusterFullDomain),
+		fmt.Sprintf("*.apps.%s,*.apps.%s", seedFullDomain, clusterFullDomain),
+	}
+
+	if _, err := os.Stat(filepath.Join(ostreeDir, certsDir)); err == nil {
```
omertuc, line 61: Shouldn't we have a hard error if certs dir doesn't exist?

internal/clusterconfig/recert.go
tsorya: Yep, it is mounted

internal/clusterconfig/recert.go
tsorya: I took the logic from script, not sure totally what was expected here. Do we always should have it? In case we should always provide it sure we should fail

internal/generated/zz_generated.bindata.go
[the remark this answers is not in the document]
tsorya: Tricky one as you actually should see what asset was changed due to code changes. Maybe lets add it when we will remove bash scripts from LCA?

internal/clusterconfig/recert.go
omertuc: In IBU we should have certs, in IBI we might not need it. But let's figure it out when we're doing IBI

tsorya: cool, let me change this part

internal/clusterconfig/recert.go
omertuc: Alright, what about:

>We also need a clean summary file, so we can include it in must-gather

internal/generated/zz_generated.bindata.go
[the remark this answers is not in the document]
omertuc: >Tricky one as you actually should see what asset was changed due to code changes.

Not sure I understand, can you elaborate? You can always see the modified generated files, you just don't see them by default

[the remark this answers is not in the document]
tsorya: agree but thought it better to be more visible, though let me add it

internal/clusterconfig/recert.go
omertuc: https://docs.google.com/presentation/d/10_SAZpbc2Ri1w__89K6EuithAK0ta6qG_wV-4tjqwXo/edit#slide=id.g294d0ae4bf8_0_137

utils/utils.go
```diff
@@ -36,3 +42,31 @@ func RenderTemplateFile(srcTemplate string, params map[string]any, dest string,
 	}
 	return nil
 }
+
+// GetSNOMasterNode get master node of sno cluster
```
omertuc, line 46:
```suggestion
```

The evil inter is gone now

utils/utils.go
```diff
@@ -36,3 +42,31 @@ func RenderTemplateFile(srcTemplate string, params map[string]any, dest string,
...
+	if err != nil {
+		return nil, err
+	}
+	if len(nodesList.Items) != 1 {
+		return nil, fmt.Errorf("we should have one master node in sno cluster, current number is %d", len(nodesList.Items))
+	}
+	return &nodesList.Items[0], nil
+}
+
+// ReadYamlOrJSONFile read json/yaml file into struct
```
omertuc, line 63:
```suggestion
```
The evil linter is gone now

internal/clusterconfig/recert.go
tsorya: I see, not sure where to add it currently, let me open ticket for it

PR #88, opened 2023-11-15
Drop hack/lint from ci-job

leo8a:


donpenney (approved):

omertuc:
Great, for the record - `golint` is a stale project and they themselves say they should not be used in CI.

golangci-lint is more than enough

PR #90, opened 2023-11-15
Move code to use chroot or nmsenter executors in order to have single interface for different objects

tsorya:


main/main.go
```diff
@@ -110,15 +111,19 @@ func main() {
...
 		TimestampFormat: "2006-01-02 15:04:05",
 		FullTimestamp:   true,
 	})
-	op := ops.NewOps(logr, ops.NewExecutor(logr, true))
-	rpmOstreeClient := rpmostreeclient.NewClient("ibu-imager", op)
-
 	log := ctrl.Log.WithName("controllers").WithName("ImageBasedUpgrade")
+
+	executor := ops.NewChrootExecutor(newLogger, true, utils.Host)
+	rpmOstreeClient := rpmostreeclient.NewClient("ibu-imager", executor)
```
jc-rh, line 125:
```suggestion
	rpmOstreeClient := rpmostreeclient.NewClient("ibu-controller", executor)
```

sudomakeinstall2: This client.id is not used anywhere. It is just an extra variable.

controllers/utils/cmd.go
```diff
@@ -32,41 +25,3 @@ const (
 	PrepSetupStateroot string = "prepSetupStateroot.sh"
 	PrepCleanup        string = "prepCleanup.sh"
 )
```
sudomakeinstall2, line 27: we can remove this file and move these constants to `constants.go`

main/main.go
jc-rh: @donpenney this client is assigned to the ibu reconciler struct

PR #91, opened 2023-11-16
[Imager] Add common checks folder and fix nits in HACKING.md

leo8a:
/cc @tsorya @omertuc

### DIFF
--- a/ibu-imager/common/consts.go
+++ b/ibu-imager/common/consts.go
@@ -18,13 +18,18 @@ package common
 
 // Common constants mainly used by packages in ibu-imager
 const (
-	VarFolder = "/var"
-	BackupDir = "/var/tmp/backup"
+	VarFolder       = "/var"
+	BackupDir       = "/var/tmp/backup"
+	BackupCertsDir  = "/var/tmp/backupCertsDir"
+	BackupChecksDir = "/var/tmp/checks"
+
 	// ImageRegistryAuthFile is the pull secret. Written by the machine-config-operator
-	ImageRegistryAuthFile             = "/var/lib/kubelet/config.json"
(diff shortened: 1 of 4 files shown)


ibu-imager/common/consts.go
```diff
@@ -18,13 +18,35 @@ package common
...
+	EtcdStaticPodContainer = "etcd"
+
 	InstallationConfigurationFilesDir = "/usr/local/installation_configuration_files"
 )
+
+// FilesToRemove is the list of systemd service units, scripts, and env file
+// to be cleaned up after creating the seed image
+var FilesToRemove = []string{
+	"/etc/systemd/system/installation-configuration.env",
+	"/var/usrlocal/bin/installation-configuration.sh",
```
tsorya, line 41: Can we read files from the folder and remove accordingly ? I would like not to use list of files if it is possible

ibu-imager/common/consts.go
```diff
@@ -18,13 +18,35 @@ package common
...
+	"/var/usrlocal/bin/prepare-installation-configuration.sh",
+	"/etc/systemd/system/installation-configuration.service",
+	"/etc/systemd/system/prepare-installation-configuration.service",
+}
+
+// SystemdUnitsToDisable is the list of systemd units to be disabled
+// after creating the seed image
+var SystemdUnitsToDisable = []string{
+	"installation-configuration.service",
```
tsorya, line 50: same

ibu-imager/common/consts.go
tsorya: Sort of same as https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/cmd/create.go#L133 
just need to take file name and build path accordingly

leo8a: agree, will do it as suggested in #79 :)

donpenney (approved):

PR #93, opened 2023-11-17
Wait for etcd to be up instead of sleep-ing before recert

mresvanis:
This PR adds waiting for the temporary `etcd` container, which is used by `recert`, to be up and running, before running `recert` in the installation configuration script.


ibu-imager/installation_configuration_files/scripts/installation-configuration.sh
```diff
@@ -28,6 +28,17 @@ echo "${RELOCATION_CONFIG_PATH} has been found"
 # Replace this with a function that loads values from yaml file
 set +o allexport
 
+function wait_for_etcd {
```
tsorya, line 31: We are going to rewrite all in golang i assume so maybe worth waiting for golang?

PR #101, opened 2023-11-21
MGMT-16256: Moving installation-configuration to golang

tsorya:

PR #103, opened 2023-11-21
Move copyConfigurationFiles to seed creation workflow

leo8a:


donpenney:
Note that there is now just a single post-reboot script, with https://github.com/openshift-kni/lifecycle-agent/pull/101 converting installation-configuration.sh to golang

leo8a:
> Note that there is now just a single post-reboot script, with https://github.com/openshift-kni/lifecycle-agent/pull/101 converting installation-configuration.sh to golang

yup, thanks. This code walks throu `installation_configuration_files/*` and handle / copy services and scripts accordingly.

donpenney (approved):

PR #107, opened 2023-11-22
Restore IBU CR post pivot

jc-rh:


main/main.go
```diff
@@ -155,3 +164,28 @@ func main() {
...
+
+func restoreIBU(client client.Client) error {
+	ibu := &lcav1alpha1.ImageBasedUpgrade{}
+	filePath := filepath.Join(utils.Host, utils.IBUFilePath)
+	if err := lcautils.ReadYamlOrJSONFile(filePath, ibu); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
```
leo8a, line 176:
```suggestion
	if err := lcautils.ReadYamlOrJSONFile(filePath, ibu); err != nil && !os.IsNotExist(err) {
		return fmt.Errorf("failed to read %s file: %w", filePath, err)
	}
```

main/main.go
```diff
@@ -155,3 +164,28 @@ func main() {
...
+
+	status := ibu.Status
+	if err := client.Create(context.TODO(), ibu); err != nil {
+		if errors.IsAlreadyExists(err) {
+			return nil
+		}
+		return err
+	}
```
leo8a, line 184:
```suggestion
	if err := client.Create(context.TODO(), ibu); err != nil && !errors.IsAlreadyExists(err) {
		return fmt.Errorf("failed to create ImageBasedUpgrade: %w", err)
	}
```

main/main.go
jc-rh: When the file does not exist, we need to return and skip the rest

main/main.go
jc-rh: Same as above. Need to skip the status update if there is an existing one.

main/main.go
```diff
@@ -155,3 +164,28 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func restoreIBU(client client.Client) error {
+	ibu := &lcav1alpha1.ImageBasedUpgrade{}
+	filePath := filepath.Join(utils.Host, utils.IBUFilePath)
```
tsorya, line 170: Wondering why we are doing it here? 
Why not to do it in post-pivot phase? I don't think this is actually the right place. 
I think we should have everything in /opt/openshift folder and it should be cleaned up at the end of the process.

tsorya: I think we are creating duplication logic for mostly same operations. 
We should decide or we are pushing everything after LCA operator start or we are pushing it in in post-pivot part.
Another way we will never be able to understand the code.

jc-rh: This is because it's a two step process. I don't want the controller to try reconciling the result from the create without the status. The post-pivot command is probably running in parallel when the controller pod is brought up by kubelet. I actually prefer not to do it here, only if there is a way to import the CR with its status in one atomic operation.

tsorya: @donpenney @jc-rh we need to decide. For sure we should not have code in 2 different places that does the same.
So we have 2 options here: 
1. Post-pivot is going to do all the apply/cleanup logic as it does now
2. Operator should do it while start running. 

Another way it will be not manageable at all. We can meet and decide about design one more time but we should understand where we implement each part of the logic as currently it is mess

tsorya: I believe i need to understand why there is an issue with applying status from post-pivot code

main/main.go
```diff
@@ -155,3 +164,28 @@ func main() {
...
+	if err := lcautils.ReadYamlOrJSONFile(filePath, ibu); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+
+	status := ibu.Status
+	if err := client.Create(context.TODO(), ibu); err != nil {
```
tsorya, line 179: please provide context from caller, context can be useful afterwards

main/main.go
```diff
@@ -124,6 +129,10 @@ func main() {
 	executor := ops.NewChrootExecutor(newLogger, true, utils.Host)
 	rpmOstreeClient := rpmostreeclient.NewClient("ibu-controller", executor)
 
+	if err := restoreIBU(mgr.GetClient()); err != nil {
+		setupLog.Error(err, "unable to restore IBU CR")
```
tsorya, line 133: we should exit in case we didn't succeed to restore ibu

main/main.go
tsorya: After talking with @jc-rh i understood why we need it here

jc-rh:
@tsorya Updated as we discussed.

main/main.go
```diff
@@ -158,3 +169,36 @@ func main() {
...
+		return err
+	}
+	log.Info("IBU CR found, restoring ...")
+	if err := client.Delete(ctx, ibu); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	status := ibu.Status
```
omertuc, line 189: Why is this `status` variable necessary? Also what's the point of the `ibu.Status = status` assignment below?

omertuc: Ack, yes please add a comment

jc-rh: Done. BTW, the good news is that this does restore all status fields as is, including the timestamps.

tsorya:
before pushing it please update in channel that it requires new seed image and from now on we must add LCA operator to seed cluster

jc-rh:
> before pushing it please update in channel that it requires new seed image and from now on we must add LCA operator to seed cluster

Since we do this in the operator itself, older seed images still "work". It just won't complete the upgrade workflow. If it were in the post-pivot command, it would fail that command when the CRD is missing.

main/main.go
```diff
@@ -158,3 +169,39 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func restoreIBU(ctx context.Context, client client.Client, log *logr.Logger) error {
+	ibu := &lcav1alpha1.ImageBasedUpgrade{}
+	filePath := filepath.Join(utils.Host, utils.IBUFilePath)
```
sakhoury, line 175:
```suggestion
	filePath := common.PathOutsideChroot(utils.IBUFilePath)
```
with: "github.com/openshift-kni/lifecycle-agent/internal/common"

PR #108, opened 2023-11-22
Rename gomock vendor from uber to github

javipolo:


javipolo:
easier solution is to install latest mockgen, so closing this

```
go install go.uber.org/mock/mockgen@latest
```

mresvanis:
> easier solution is to install latest mockgen, so closing this
> 
> ```
> go install go.uber.org/mock/mockgen@latest
> ```

@javipolo the github one is deprecated, that's why it's best to keep working with the uber one.

PR #109, opened 2023-11-22
Fix problem with quoted kernel arguments

javipolo:
/cc @tsorya 
/cc @donpenney

### DIFF
--- a/internal/bindata/prepSetupStateroot.sh
+++ b/internal/bindata/prepSetupStateroot.sh
@@ -42,7 +42,10 @@ function log_it {
 
 function build_kargs {
     jq -r '.spec.kernelArguments[]' "${img_mnt}/mco-currentconfig.json" \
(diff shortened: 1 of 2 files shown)


internal/bindata/prepSetupStateroot.sh
```diff
@@ -42,7 +42,10 @@ function log_it {
 
 function build_kargs {
     jq -r '.spec.kernelArguments[]' "${img_mnt}/mco-currentconfig.json" \
-        | xargs --no-run-if-empty -I% echo -n "--karg-append % "
+        | while IFS= read -r line; do
```
romfreiman, line 45: shouldnt it move to golang?

javipolo: yes, this is a temporary fix to keep it working with 4.14 until we move to golang

PR #112, opened 2023-11-22
CNF-10428: Add functionality to allow best-effort for pre-caching

sakhoury:
This PR contains changes to allow a developer to configure pre-caching to `best-effort`, i.e. continue with Prep stage when the pre-caching job failed to pull one or more images.

/cc @donpenney @browsell

### DIFF
--- a/main/precache-workload/main.go
+++ b/main/precache-workload/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"strings"
 	"syscall"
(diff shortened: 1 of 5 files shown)


main/precache-workload/main.go
```diff
@@ -66,32 +80,37 @@ func main() {
...
+		terminateOnError(fmt.Errorf("failed to execute podman command"), bestEffort)
 	}
 	log.Info("podman is running, proceeding to pre-cache images!")
 
 	// Pre-cache images
 	status, err := workload.PullImages(precacheSpec)
 	if err != nil {
-		log.Errorf("Encountered error while pre-caching images, error: %v", err)
-		os.Exit(1)
+		terminateOnError(fmt.Errorf("encountered error while pre-caching images, error: %v", err), bestEffort)
```
donpenney, line 113: I think the only error we want to ignore for best-effort is the pull. The other failures, like chroot, the podman check, etc, we would still want to exit with a failure. It would only be pull failures that would be ignored

browsell: +1

sakhoury: I agree 100%. I initially was under the impression that the ask was for pre-caching not to fail - I will modify this to not fail for only pull failures. Thanks!

PR #114, opened 2023-11-23
Minor refactoring of plugins and common code

sakhoury:
This PR consists of a minor refactoring of the plugins and controllers. Shared code is housed under `internal/common/utils.go`.

/cc @Missxiaoguo @pixelsoccupied

### DIFF
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -59,7 +59,7 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 
 	// backup with OADP
 	r.Log.Info("Backup with Oadp operator")
-	reqBackup, backupCRs, err := r.BackupRestore.CheckIfBackupRequested(ibu.Spec.OADPContent, ctx)
+	reqBackup, backupCRs, err := r.BackupRestore.CheckIfBackupRequested(ctx, ibu.Spec.OADPContent)
 	if err != nil {
 		if k8serrors.IsInvalid(err) {
 			utils.SetStatusCondition(&ibu.Status.Conditions,
@@ -70,7 +70,6 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 				ibu.Generation)
 			return requeueWithShortInterval(), nil
 		}
-		r.Log.Error(err, "")
 		return requeueWithError(err)
 	}
 	if reqBackup {
--- a/internal/backuprestore/backup.go
+++ b/internal/backuprestore/backup.go
@@ -25,6 +25,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/openshift-kni/lifecycle-agent/internal/common"
+
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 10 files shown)


controllers/ibu_controller.go
```diff
@@ -64,7 +64,7 @@ func doNotRequeue() ctrl.Result {
 	return ctrl.Result{}
 }
 
-func requeueWithError(err error) (ctrl.Result, error) {
+func doNotRequeueWithError(err error) (ctrl.Result, error) {
 	// can not be fixed by user during reconcile
 	return ctrl.Result{}, err
 }
```
pixelsoccupied, line 70: please keep me honest here but i think the name of the func was right e.g 

https://sdk.operatorframework.io/docs/building-operators/golang/tutorial/#reconcile-loop
https://cloud.redhat.com/blog/subreconciler-patterns-in-action (original inspiration)


Right now we have 
```
func doNotRequeue() ctrl.Result {
	return ctrl.Result{}
}
``` 
which might not be too clear...basically the only way to doNotRequeue --> `ctrl.Result{}, nil`

controllers/upgrade_handlers.go
```diff
@@ -71,7 +71,7 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 			return requeueWithShortInterval(), nil
 		}
 		r.Log.Error(err, "")
-		return requeueWithError(err)
+		return doNotRequeueWithError(err)
```
pixelsoccupied, line 74: since we requeue with error...L73 error logging is redundant and can be removed

controllers/ibu_controller.go
sakhoury: Thanks for sharing the resources - that is correct. I'll restore to the original version :)

internal/backuprestore/backup.go
```diff
@@ -55,7 +57,7 @@ func (h *BRHandler) CheckIfBackupRequested(content []lcav1alpha1.ConfigMapRef, c
 	}
 
 	// extract CM
-	oadpConfigmaps, err := getConfigMaps(ctx, h, content)
+	oadpConfigmaps, err := common.GetConfigMaps(ctx, h, content)
```
Missxiaoguo, line 60: Not related to your change, but should this be?
```suggestion
	oadpConfigmaps, err := common.GetConfigMaps(ctx, h.client, content)
```

sakhoury: It should be indeed! I'll include the change :)

internal/precache/helper.go
```diff
@@ -36,16 +39,14 @@ import (
...
+// nil is returned
 func getConfigMap(ctx context.Context, c client.Client, name, namespace string) (*corev1.ConfigMap, error) {
-	cm := &corev1.ConfigMap{}
-	if err := c.Get(ctx, types.NamespacedName{
+	cm, err := common.GetConfigMap(ctx, c, v1alpha1.ConfigMapRef{
 		Name:      name,
 		Namespace: namespace,
-	}, cm); err != nil {
+	})
+	if err != nil {
```
Missxiaoguo, line 49: Actually, do you still need this helper function in precache? You could directly call the util one in the validateJobConfig and only add the IsNotFound check?

internal/backuprestore/backup.go
Missxiaoguo: Thanks!

internal/precache/helper.go
sakhoury: Thanks Angie, I have refactored as per you suggestion.

PR #115, opened 2023-11-23
Seed generation orchestration

donpenney:
This update introduces a new SeedGenerator CRD to facilitate end-to-end orchestration with LCA of the seed image generation. This will handle ACM deregistration and cleanup, if a hubKubeconfig is provided, prior to launching the ibu-imager.

This does not currently include recovery of the seed SNO after the seed image is generated.


controllers/seedgen_controller.go
```diff
@@ -0,0 +1,635 @@
...
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "assisted-installer",
+		}}
+	if err := r.Client.Delete(ctx, ns, deleteOpts...); client.IgnoreNotFound(err) != nil {
+		return fmt.Errorf("Failed to delete assisted-installer namespace: %w", err)
+	}
+
+	// TODO: We could do a query similar to the namespace query above to look for all remaining open-cluster-management.io CRDs
+	crds := []string{
```
browsell, line 268: These as well ? 
clusterclaims.cluster.open-cluster-management.io
klusterlets.operator.open-cluster-management.io

donpenney: Those get automatically cleaned up as part of deregistration, so I've never needed to manually delete them. I could change this to a query similar to the namespace one, searching for CRDs matching *.open-cluster-management.io and delete whatever's there

api/v1alpha1/seedgenerator_types.go
```diff
@@ -0,0 +1,64 @@
+/*
```
jc-rh, line 1: Apparently the new code generator assumes all types defined in a file named `types.go`.  We had to merge talm ones in this PR: https://github.com/openshift-kni/cluster-group-upgrades-operator/pull/709

sudomakeinstall2: https://github.com/kubernetes/code-generator/blob/7855250cd36120e10dc5bbf693ac4bdf42b7541f/kube_codegen.sh#L399

https://github.com/kubernetes/code-generator/blob/7855250cd36120e10dc5bbf693ac4bdf42b7541f/kube_codegen.sh#L555

donpenney: Interesting... thanks

donpenney: Created a separate dir

controllers/seedgen_controller.go
```diff
@@ -0,0 +1,644 @@
...
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
+//+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch;delete
+//+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
+//+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=delete
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=delete
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=delete
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch;delete
+
+func (r *SeedGeneratorReconciler) stringToFile(filename, str string) error {
```
jc-rh, line 91: Nit, this function doesn't use anything from SeedGeneratorReconciler?

donpenney: No, I mostly just grouped the various new bits I was adding to the reconciler, since that's where I was using them. Just a local utility function, but could easily move it to the utils module

controllers/utils/conditions.go
```diff
@@ -18,6 +18,8 @@ var ConditionTypes = struct {
 	UpgradeCompleted   ConditionType
 	RollbackInProgress ConditionType
 	RollbackCompleted  ConditionType
+	SeedGenInProgress  ConditionType
```
jc-rh, line 21: Let's create another struct for the seedgen ones?

controllers/seedgen_controller.go
```diff
@@ -0,0 +1,630 @@
...
+func (r *SeedGeneratorReconciler) createHubClient(hubKubeconfig []byte) (hubClient client.Client, err error) {
+	config, err := clientcmd.RESTConfigFromKubeConfig(hubKubeconfig)
+	if err != nil {
+		err = fmt.Errorf("Failed RESTConfigFromKubeConfig: %w", err)
+		return
+	}
+
+	hubClient, err = client.New(config, client.Options{Scheme: r.Scheme})
+	if err != nil {
+		err = fmt.Errorf("Failed to create hub client: %w", err)
```
leo8a, line 101: s/Failed/failed

there are many across this file

donpenney: Why would you want this lowercase?

leo8a: Error string should not be capitalized or end with punctuation mark 
 Inspection info: Reports format issues in error strings.

Example:
err := fmt.Errorf("Cannot read the file!")
log.Printf("Reading %s: %v", file, err)

According to [Error Strings at github.com/golang](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), error strings should not be capitalized or end with punctuation because they might appear among other context.

donpenney: Interesting. Maybe there's a golangci-lint check we can turn on for that

jc-rh: I was about to say the same thing, as it does happen for TALM. I thought we copied the linters and config from TALM ...

donpenney: The .golangci.yml is the same for both repos

donpenney: There are a couple of related golangci-lint plugins I'd like to enable, and a config I can do to add a check for the capitalized Errorf, but I'd like to do that as a separate PR. In the meantime, I can fix the capitals in this code, and create a separate PR to address the other issues and add linters

main/main.go
```diff
@@ -73,11 +77,15 @@ var (
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(lcav1alpha1.AddToScheme(scheme))
+	utilruntime.Must(seedgenv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(ocpV1.AddToScheme(scheme))
 	utilruntime.Must(mcv1.AddToScheme(scheme))
 	utilruntime.Must(velerov1.AddToScheme(scheme))
 	utilruntime.Must(operatorsv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(operatorv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(clusterv1.AddToScheme(scheme))
```
leo8a, line 86: ?
```suggestion
	utilruntime.Must(clusterv1.Install(scheme))
```

AddToScheme seems deprecated for the clusterv1 package - https://github.com/openshift-kni/lifecycle-agent/blob/main/vendor/open-cluster-management.io/api/cluster/v1/register.go#L9-L22

(also deprecated for the ocpV1 package but not touched by this PR, in case's useful)

controllers/seedgen_controller.go
donpenney: https://github.com/openshift-kni/lifecycle-agent/pull/130

main/main.go
donpenney: Given that the controller-gen we're using doesn't add an "Install", I'd prefer to keep this as-is for now until we change others, since AddToScheme and Install are pointing to the same function.

omertuc:
https://github.com/openshift-kni/lifecycle-agent/pull/133 to make diff in this PR less noisy

jc-rh (approved):

PR #117, opened 2023-11-24
[WIP] Add go staticcheck to CI

sudomakeinstall2:
The current linters are lacking some checks. Adding go staticcheck to cover missing checks.


donpenney:
This is currently an enabled plugin in our golangci-lint config:
https://github.com/openshift-kni/lifecycle-agent/blob/main/.golangci.yml#L19

We should review the available plugins that we aren't using, though, to see if there's more that would be good to add

sudomakeinstall2:
> This is currently an enabled plugin in our golangci-lint config: https://github.com/openshift-kni/lifecycle-agent/blob/main/.golangci.yml#L19
> 
> We should review the available plugins that we aren't using, though, to see if there's more that would be good to add

they might be different tools since the current one is not catching a lot things, such:
```
internal/generated/zz_generated.bindata.go:10:2: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.  (SA1019)
internal/backuprestore/restore_test.go:30:2: package "k8s.io/apimachinery/pkg/apis/meta/v1" is being imported more than once (ST1019)
internal/backuprestore/restore_test.go:30:2: package "k8s.io/apimachinery/pkg/apis/meta/v1" is being imported more than once (ST1019)

```

sudomakeinstall2:
@donpenney 

They used the same name but it is not the same tool
https://github.com/golangci/golangci-lint/blob/ce020c6be1b8f9da457436fa9b85f7f54f8107b7/pkg/golinters/staticcheck.go#L16

I think the current version lacks a lot of checks.

PR #118, opened 2023-11-24
MGMT-16283: Adding support for caBundle config and icsp

tsorya:


internal/clusterconfig/clusterconfig.go
```diff
@@ -40,7 +42,11 @@ const (
 
 	clusterIDFileName = "cluster-id-override.json"
 
-	idmsFileName = "image-digest-mirror-set.json"
+	idmsFileName  = "image-digest-mirror-set.json"
+	icspsFileName = "image-content-policy-list.json"
+
+	caBundleCMName   = "user-ca-bundle"
```
browsell, line 48: This assumes the configmap name is user-ca-bundle ? Not sure we can assume that

tsorya: Actually we can. This is hardcoded name in mco too

internal/clusterconfig/clusterconfig.go
```diff
@@ -295,3 +307,64 @@ func (r *UpgradeClusterConfigGather) getIDMSs(ctx context.Context) (v1.ImageDige
...
+	r.Log.Info("Fetching user ca bundle")
+	caBundle := &corev1.ConfigMap{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: caBundleCMName,
+		Namespace: configNamespace}, caBundle)
+	if err != nil && errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get ca bundle cm, err: %w", err)
+	}
```
sakhoury, line 356: Perhaps we can replace these lines with the function `GetConfigMap` in [func GetConfigMap(ctx context.Context, c client.Client, configMap v1alpha1.ConfigMapRef) (*corev1.ConfigMap, error) {](https://github.com/openshift-kni/lifecycle-agent/blob/main/internal/common/utils.go#L32-L43).

tsorya: I can, just not fully understand why we actually have this util function and what it gives us. Will be glad to understand.

sakhoury: Sure! This function, was defined and, is used by several plugins. The intent is to define commonly used functions in a unified location that can be easily referenced. In your case, you only need to fetch a ConfigMap once, however, there are cases where ConfigMaps need to be retrieved in several places - thus the reasoning behind its existence (avoid code duplication).

PR #121, opened 2023-11-24
Minor refactoring of pre-caching plugin

sakhoury:
Refactor pre-caching plugin to conform with Handler pattern used by the Backup&Restore and ExtraManifests plugins.

/cc @pixelsoccupied @jc-rh

### DIFF
--- a/internal/precache/precache.go
+++ b/internal/precache/precache.go
@@ -20,19 +20,25 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
+	"github.com/go-logr/logr"
+
 	"os"
 	"strings"
 
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
 
-	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;delete
 
-var log = ctrl.Log.WithName("precache")
+// PHandler handles the precaching job
+type PHandler struct {
+	client.Client
+	Log logr.Logger
+}
 
 // Config defines the configuration options for a pre-caching job.
 type Config struct {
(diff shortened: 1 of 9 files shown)


internal/precache/precache.go
```diff
@@ -167,28 +173,28 @@ func QueryJobStatus(ctx context.Context, c client.Client) (*Status, error) {
...
+	if err := deleteJob(ctx, handler.Client, LcaPrecacheJobName, LcaPrecacheNamespace); err != nil {
+		handler.Log.Error(err, "Failed to delete precaching job", "name", LcaPrecacheJobName)
 		return err
 	}
 	// Delete ConfigMap
-	if err := deleteConfigMap(ctx, c, LcaPrecacheConfigMapName, LcaPrecacheNamespace); err != nil {
-		log.Error(err, "Failed to delete precaching configmap", "name", LcaPrecacheConfigMapName)
+	if err := deleteConfigMap(ctx, handler.Client, LcaPrecacheConfigMapName, LcaPrecacheNamespace); err != nil {
+		handler.Log.Error(err, "Failed to delete precaching configmap", "name", LcaPrecacheConfigMapName)
 		return err
```
pixelsoccupied, line 198: Since we are return the errors with requeueWithError. The error will show up in the pods logs anyway. This means we are showing the same error twice? 

Maybe something like will avoid printing it twice + logging something extra for the user?
```
handler.Log.Error(fmt.Errorf("Failed to delete precaching job"), LcaPrecacheJobName))
return err 
```

Or maybe simpler would be just to remove L197 (and other place were we end with the requeueWithError). wdyt?

sakhoury: Great point! I have updated the logging to info and omitted the error message as it will be logged! Thanks Nahian :)

sudomakeinstall2 (commented):
We have this in `main.go`
https://github.com/openshift-kni/lifecycle-agent/pull/90/files#diff-327181d0a8c5e6b164561d7910f4eeffd41442d55b2a2788fda2aa2692f17ec0R114

Is it possible to use `logrus` here?

internal/precache/precache.go
```diff
@@ -100,61 +106,61 @@ type Status struct {
 }
 
 // CreateJob creates a new precache job.
-func CreateJob(ctx context.Context, c client.Client, config *Config) error {
+func (handler *PHandler) CreateJob(ctx context.Context, config *Config) error {
```
tsorya, line 109: maybe lets call handler -> h
Will make lines shorter. WDYT?

sakhoury: Good suggestion @tsorya, will do that!

sakhoury:
> We have this in `main.go` https://github.com/openshift-kni/lifecycle-agent/pull/90/files#diff-327181d0a8c5e6b164561d7910f4eeffd41442d55b2a2788fda2aa2692f17ec0R114
> 
> Is it possible to use `logrus` here?

@sudomakeinstall2 - it would be better to change logging for all plugins in a separate PR. The objective behind the current PR is to refactor the pre-caching plugin to conform with the Handler patter used by both B&R and EM plugins.

PR #129, opened 2023-11-29
Convert prep scripts to golang

sudomakeinstall2:
Convert `PrepSetupStateroot.sh` and `PrepCleanup.sh` to golang.


controllers/prep_handlers.go
```diff
@@ -173,6 +177,289 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
+	spec, ok := v["spec"].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("failed getting spec in %s", path)
+	}
+	kargs, ok := spec["kernelArguments"].([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("failed getting kernelArguments in %s", path)
+	}
+	args := make([]string, len(kargs)*2)
+	for i, karg := range kargs {
```
tsorya, line 252: Please validate kernel args logic before push , after applying full flow verify that MCO operator is not degraded

controllers/prep_handlers.go
```diff
@@ -173,6 +177,289 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
+		return "", fmt.Errorf("failed to unmarshal %s: %w", path, err)
+	}
+	version, ok := v["version"].(string)
+	if !ok {
+		return "", fmt.Errorf("failed to get version from %s: %w", path, err)
+	}
+	return version, nil
+}
+
+func getKernelArgumentsFromMCOFile(path string) ([]string, error) {
```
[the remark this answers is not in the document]
sudomakeinstall2: I will use this one, thanks.

controllers/prep_handlers.go
```diff
@@ -173,6 +177,289 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
+	for i, karg := range kargs {
+		args[2*i] = "--karg-append"
+		args[2*i+1] = karg.(string)
+	}
+	return args, nil
+}
+
+func getStaterootPath(osname string) string {
+	return fmt.Sprintf("/ostree/deploy/%s", osname)
+}
```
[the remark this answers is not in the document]
sudomakeinstall2: Sure, i will add it to utils.
I will create separate PR for refactoring other places.

internal/prep/prep.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("error while reading %s: %w", file.Name(), err)
+	}
+
+	return nil
+}
+
+func BackupCertificates(ctx context.Context, osname string, client *clusterinfo.InfoClient) error {
```
sudomakeinstall2, line 111: This is very similar to `BackupSeedCertificates`. 
After this gets merged, in a separate PR I will refactor them, so they use a common base function.

sudomakeinstall2: https://github.com/openshift-kni/lifecycle-agent/pull/142

controllers/prep_handlers.go
```diff
@@ -173,53 +169,170 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
-	result = requeueWithShortInterval()
+	mountpoint, err := r.Executor.Execute("podman", "image", "mount", ibu.Spec.SeedImageRef.Image)
+	if err != nil {
+		return fmt.Errorf("failed to mount seed image: %w", err)
+	}
 
-	return
-}
+	defer func() {
+		if _, err := r.Executor.Execute("podman", "image", "umount", ibu.Spec.SeedImageRef.Image); err != nil {
```
tsorya, line 205: same question as with rmi, if it will fail here, we will not be able to run it one more time, right?

controllers/prep_handlers.go
```diff
@@ -173,53 +169,170 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
+
+	if _, err = r.Executor.Execute(
+		"cp",
+		filepath.Join(mountpoint, fmt.Sprintf("ostree-%s.origin", seedBootedDeployment)),
+		prep.GetDeploymentOriginPath(osname, deploymentID),
+	); err != nil {
+		return fmt.Errorf("failed to restore origin file: %w", err)
+	}
+
+	if _, err = r.Executor.Execute(
```
tsorya, line 274: seems like we are running same command just with another path, lets make it function? it can be in ops.

internal/prep/prep.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	version, ok := v["version"].(string)
+	if !ok {
+		return "", fmt.Errorf("failed to get version from %s: %w", path, err)
+	}
+	return version, nil
+}
+
+func BuildKernelArgumentsFromMCOFile(path string) ([]string, error) {
+	mcJSON, err := os.Open(path)
```
tsorya, line 58: Maybe we can use ReadYamlOrJSONFile from utils?

controllers/prep_handlers.go
```diff
@@ -173,53 +169,170 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
 
-	// Write the script
-	scriptname := filepath.Join(utils.IBUWorkspacePath, utils.PrepSetupStateroot)
-	scriptcontent, _ := generated.Asset(utils.PrepSetupStateroot)
-	err = os.WriteFile(common.PathOutsideChroot(scriptname), scriptcontent, 0o700)
+	defer func() {
+		if err := os.RemoveAll(workspaceOutsideChroot); err != nil {
+			r.Log.Error(err, "failed to clean up "+workspaceOutsideChroot)
+		}
+		if _, err := r.Executor.Execute("podman", "rmi", ibu.Spec.SeedImageRef.Image); err != nil {
```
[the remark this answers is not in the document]
sudomakeinstall2: if this fails, it will only log an error. It won't fail the prep stage.

controllers/prep_handlers.go
sudomakeinstall2: no this only logs an error. It won't fail the prep stage.

sudomakeinstall2: should it fail the stage?

controllers/prep_handlers.go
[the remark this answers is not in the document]
sudomakeinstall2: Should it fail the stage? I don't think it should.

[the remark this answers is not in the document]
sudomakeinstall2: We can try to clean them up in abort or finalize.

controllers/prep_handlers.go
browsell: We should probably fail in this scenario.

ibu-imager/ops/ops.go
```diff
@@ -209,3 +211,15 @@ func (o *ops) RunRecert(recertContainerImage, authFile, recertConfigFile string,
...
 }
+
+func (o *ops) ExtractTarWithSELinux(srcPath, destPath string) error {
+	_, err := o.hostCommandsExecutor.Execute(
+		"tar", "xzf", srcPath, "-C", destPath, "--selinux",
+	)
+	return err
+}
+
+func (o *ops) RemountSysroot() error {
```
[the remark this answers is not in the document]
sudomakeinstall2: I am not sure.
but `hostCommandExecutor` is `ChrootExecutor` here. the command executed is `Executing /usr/bin/env with args [-- bash -c mount /sysroot -o remount,rw]`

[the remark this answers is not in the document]
donpenney: SSH to the host after prep is done and check the /sysroot mountpoint to see if it's rw or ro.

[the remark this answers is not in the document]
donpenney: I think it should be ok, that it will be ro on the node. You could also rsh into the LCA manager container to check there, but that's less of a concern, and would likely be dependent on whether the command is being launched in a go thread or in the main thread.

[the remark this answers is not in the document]
sudomakeinstall2:
```
sh-5.1# cat /proc/mounts | grep "/sysroot"
/dev/nvme0n1p4 /sysroot xfs ro,seclabel,relatime,attr2,inode64,logbufs=8,logbsize=32k,prjquota 0 0
/dev/nvme0n1p4 /sysroot/ostree/deploy/rhcos/var xfs rw,seclabel,relatime,attr2,inode64,logbufs=8,logbsize=32k,prjquota 0 0
```
It looks like it is read only

controllers/prep_handlers.go
```diff
@@ -173,53 +169,165 @@ func (r *ImageBasedUpgradeReconciler) queryPrecachingStatus(ctx context.Context,
...
+	}
+
+	if err = prep.RemoveETCDeletions(mountpoint, osname, deploymentID); err != nil {
+		return fmt.Errorf("failed to restore etc directory: %w", err)
+	}
+
+	prep.BackupCertificates(ctx, osname, r.ManifestClient)
+
+	if _, err = r.Executor.Execute(
+		"cp",
```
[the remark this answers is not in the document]
sudomakeinstall2: if we use this we have to add `/host` to src and dest. So I added a separate function for this.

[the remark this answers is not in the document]
donpenney: We have the utility function common.PathOutsideChroot() for prefixing with /host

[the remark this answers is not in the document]
sudomakeinstall2: I know we have to wrap src and dest in that everytime. so I added another function.
```
func CopyOutsideChroot(src, dest string) error {
	return cp.Copy(PathOutsideChroot(src), PathOutsideChroot(dest))
}
```

controllers/prep_handlers.go
tsorya: Yes, want to be sure we do as they expect it

ibu-imager/ops/ops.go
```diff
@@ -209,3 +217,85 @@ func (o *ops) RunRecert(recertContainerImage, authFile, recertConfigFile string,
...
+	_, err := o.hostCommandsExecutor.Execute("mount", "/sysroot", "-o", "remount,rw")
+	return err
+}
+
+func (o *ops) ImageExists(img string) (bool, error) {
```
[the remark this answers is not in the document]
donpenney: That shouldn't be a problem. If you run ```podman image mount $img``` multiple times, it returns the same mountpoint each time. It does seem to have an internal counter of the mount references, however, so the mountpoint wouldn't get cleaned up on the first unmount in that case. But the ```podman rmi``` would unmount it. So assuming the seed image itself hasn't been updated, a re-execution of the pull and the image mount would effectively be no-ops.

ibu-imager/ops/ops.go
```diff
@@ -209,3 +217,85 @@ func (o *ops) RunRecert(recertContainerImage, authFile, recertConfigFile string,
...
+	}
+	var images []PodmanImage
+	if err := json.Unmarshal([]byte(output), &images); err != nil {
+		return false, err
+	}
+	for _, img := range images {
+		if len(img.Repositories) == 0 {
+			continue
+		}
+		if img.Repositories[0] == imgName {
```
donpenney, line 268: I think you should loop over the rang of Repositories here, in case there's multiple tags for the seed image. ie. I will sometimes pull and retag the seed image so that the rmi doesn't fully delete the image, saving time redownloading on retries

sudomakeinstall2: updated

ibu-imager/ops/ops.go
```diff
@@ -209,3 +217,87 @@ func (o *ops) RunRecert(recertContainerImage, authFile, recertConfigFile string,
...
+	output, err := o.hostCommandsExecutor.Execute("podman", "image", "mount", "--format", "json")
+	if err != nil {
+		return false, err
+	}
+	var images []PodmanImage
+	if err := json.Unmarshal([]byte(output), &images); err != nil {
+		return false, err
+	}
+	for _, img := range images {
+		if len(img.Repositories) == 0 {
```
donpenney, line 265: nit: With the for loop, this check would now be unnecessary

sudomakeinstall2: i just saw remember that

donpenney (commented):
/lgtm

PR #131, opened 2023-11-29
CNF-10483: Upgrade graph configuration in LCA bundle

leo8a:
required for midstream builds.

/cc @jc-rh @rauhersu @donpenney

### DIFF
--- a/config/manifests/bases/lifecycle-agent.clusterserviceversion.yaml
+++ b/config/manifests/bases/lifecycle-agent.clusterserviceversion.yaml
@@ -17,11 +17,12 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
+    olm.skipRange: '>=4.14.0 <4.16.0'
     operators.openshift.io/infrastructure-features: '["disconnected"]'
     provider: Red Hat
     repository: https://github.com/openshift-kni/lifecycle-agent
     support: Red Hat
-  name: lifecycle-agent.v0.0.0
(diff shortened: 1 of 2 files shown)


config/manifests/bases/lifecycle-agent.clusterserviceversion.yaml
```diff
@@ -73,4 +74,5 @@ spec:
   maturity: alpha
   provider:
     name: Red Hat
+  replaces: lifecycle-agent.v4.13.0
```
jc-rh, line 77: Do we plan to release LCA on 4.13?

rauhersu (approved):
Lgtm

config/manifests/bases/lifecycle-agent.clusterserviceversion.yaml
rauhersu: We might use 4.15 values when 4.14 is stated in these changes and for the 'replaces' thing, I am afraid we cannot be congruent as there is no previous value to replace, but the field must exist.

Anyway, all these changes are automatically processed and updated to the right values by midstream builds so the values here shouldn't be relevant.

jc-rh: if the value doesn't matter, maybe we don't want to put 4.13 there? People may think it means it was supported on 4.13?

leo8a: `replaces: lifecycle-agent.v0.0.0` ?

rauhersu: it will work; at least folks won't get a 4.13 was available as well as @jc-rh commented.

jc-rh (approved):

PR #135, opened 2023-11-29
Update dnsmasq override generator script

donpenney:


hack/generate-dnsmasq-site-policy-section.sh
```diff
@@ -63,20 +89,28 @@ EOF
...
+        cat <<EOF
+  annotations:
+    ran.openshift.io/ztp-deploy-wave: "${ZTP_DEPLOY_WAVE}"
+EOF
+    fi
+    cat <<EOF
 spec:
   config:
+    ignition:
+      version: 3.1.0
```
alosadagrande, line 107: Why ignition version 3.1.0? I think we can use 3.4.0

donpenney: This is to align with the content generated here: https://github.com/openshift/assisted-service/blob/master/internal/network/manifests_generator.go#L134

PR #140, opened 2023-12-01
    Add ExecuteWithLiveLogger and stop logging commands output by default

sudomakeinstall2:
logging output of every `ChrootExecutor.Execute` is cluttering the log,
log the output of commands explicitly if needed by calling `ExecuteWithLiveLogger`
Also removed unused logger

### DIFF
--- a/ibu-imager/ops/execute.go
+++ b/ibu-imager/ops/execute.go
@@ -103,9 +103,17 @@ func NewChrootExecutor(logger *logrus.Logger, verbose bool, root string) Execute
 	return &chrootExecutor{executor: executor{logger, verbose}, root: root}
 }
 
-func (e *chrootExecutor) Execute(command string, args ...string) (string, error) {
+func (e *chrootExecutor) baseExecute(writer io.Writer, command string, args ...string) (string, error) {
 	commandBase := "/usr/bin/env"
 	args = append([]string{command}, args...)
 	arguments := []string{"--", "bash", "-c", strings.Join(args, " ")}
-	return e.executor.execute(e.executor.log.Writer(), e.root, commandBase, arguments...)
+	return e.executor.execute(writer, e.root, commandBase, arguments...)
+}
+
+func (e *chrootExecutor) Execute(command string, args ...string) (string, error) {
+	return e.baseExecute(nil, command, args...)
+}
+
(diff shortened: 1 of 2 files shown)


ibu-imager/ops/execute.go
```diff
@@ -103,9 +103,17 @@ func NewChrootExecutor(logger *logrus.Logger, verbose bool, root string) Execute
...
+func (e *chrootExecutor) baseExecute(writer io.Writer, command string, args ...string) (string, error) {
 	commandBase := "/usr/bin/env"
 	args = append([]string{command}, args...)
 	arguments := []string{"--", "bash", "-c", strings.Join(args, " ")}
-	return e.executor.execute(e.executor.log.Writer(), e.root, commandBase, arguments...)
+	return e.executor.execute(writer, e.root, commandBase, arguments...)
+}
+
+func (e *chrootExecutor) Execute(command string, args ...string) (string, error) {
+	return e.baseExecute(io.Discard, command, args...)
```
tsorya, line 114: You can just provide nil and it will be handled in callee

ibu-imager/ops/execute.go
```diff
@@ -103,9 +103,17 @@ func NewChrootExecutor(logger *logrus.Logger, verbose bool, root string) Execute
...
+}
+
+func (e *chrootExecutor) ExecuteWithLiveLogger(command string, args ...string) (string, error) {
```
tsorya, line 117: Should not we add this func to interface?

main/main.go
```diff
@@ -134,11 +129,6 @@ func main() {
 	// We want to remove logr.Logger first step to move to logrus
 	// in the future we will have only one of them
 	newLogger := logrus.New()
-	logger.SetFormatter(&logrus.TextFormatter{
```
jc-rh, line 137: This is meant to be done to `newLogger`?

sudomakeinstall2: no I originally added this for the `logger`. After refactoring the loggers, `logger` became unused and we forgot to remove the `logger` and this.

jc-rh: Ok. I was just wondering about the format. Looks weird with two different formats mixed together. Anyway to converge on one? Not for this PR though.

donpenney (commented):
/lgtm

PR #142, opened 2023-12-01
Remove InfoClient and move its functions to utils/client_helper.go

sudomakeinstall2:
InfoClient is a wrapper to runtimeclient.Client, we can remove
it and use runtimeclient.Client directly. Its functions are moved
to utils/client_helper.go
Doing this also enables us to use BackupCeritificates in the LCA

### DIFF
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -31,11 +31,11 @@ import (
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
-	"github.com/openshift-kni/lifecycle-agent/ibu-imager/clusterinfo"
+	commonUtils "github.com/openshift-kni/lifecycle-agent/utils"
+
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
 	"github.com/openshift-kni/lifecycle-agent/internal/precache"
 	"github.com/openshift-kni/lifecycle-agent/internal/prep"
-	commonUtils "github.com/openshift-kni/lifecycle-agent/utils"
 	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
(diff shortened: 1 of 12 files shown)


ibu-imager/seedcreator/seedcreator.go
```diff
@@ -241,39 +243,36 @@ func (s *SeedCreator) createContainerList() error {
 	return nil
 }
 
-func (s *SeedCreator) backupSeedCertificates(ctx context.Context) error {
-	s.log.Info("Backing up seed cluster certificates for recert tool")
-	if err := os.MkdirAll(common.BackupCertsDir, os.ModePerm); err != nil {
-		return fmt.Errorf("error creating %s: %w", common.BackupCertsDir, err)
+func BackupCertificates(ctx context.Context, client *clusterinfo.InfoClient, certDir string) error {
```
sudomakeinstall2, line 246: @leo8a @tsorya 
Any suggestion on where to put this function so it can be used in LCA controller?
Can't put it in utils since it will cause import cycle.
`InfoClient` is importing both utils.

tsorya: utils sounds good just need to move GetConfigMapData to utils too, not sure why those functions are part of info client.
We can actually create some client_helper interface or just file in utils and move there all the functions that wrap client operations

sudomakeinstall2:
@tsorya What do you think about this? :)

tsorya:
works for me

controllers/prep_handlers.go
```diff
@@ -287,7 +288,12 @@ func (r *ImageBasedUpgradeReconciler) SetupStateroot(ctx context.Context, ibu *l
 		return fmt.Errorf("failed to process etc.deletions: %w", err)
 	}
 
-	prep.BackupCertificates(ctx, osname, r.ManifestClient)
+	certsDir := common.PathOutsideChroot(
+		filepath.Join(common.GetStaterootPath(osname), "/var/opt/openshift/certs"),
```
donpenney, line 292: nit: We should aim to have all paths/filenames like this in consts. Maybe a separate PR for cleanup, to look for others that we could move to consts.

utils/client_helper.go
```diff
@@ -0,0 +1,187 @@
...
+		Domain:          installConfig.BaseDomain,
+		Version:         clusterVersion.Status.Desired.Version,
+		ClusterID:       string(clusterVersion.Spec.ClusterID),
+		MasterIP:        ip,
+		ReleaseRegistry: releaseRegistry,
+		Hostname:        hostname,
+	}, nil
+}
+
+// TODO: add dual stuck support
```
donpenney, line 122: nit: dual stack

jc-rh (approved):

PR #143, opened 2023-12-04
Add handleFinalize and perform backup cleanups

sudomakeinstall2:
- Add function to remove all unbooted stateroots
- Try to run all cleanup functions even if one of them returns an error
- Run backup cleanup and delete OADP operator in abort and finalize

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -205,28 +205,10 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 		switch idleCondition.Reason {
 		case string(utils.ConditionReasons.Aborting):
 			nextReconcile, err = r.handleAbort(ctx, ibu)
-			if err != nil {
-				utils.SetStatusCondition(&ibu.Status.Conditions,
-					utils.ConditionTypes.Idle,
-					utils.ConditionReasons.AbortFailed,
-					metav1.ConditionFalse,
-					"Error occurred during abort",
-					ibu.Generation,
-				)
-			}
 		case string(utils.ConditionReasons.AbortFailed):
 			nextReconcile, err = r.handleAbortFailure(ctx, ibu)
 		case string(utils.ConditionReasons.Finalizing):
 			nextReconcile, err = r.handleFinalize(ctx, ibu)
-			if err != nil {
-				utils.SetStatusCondition(&ibu.Status.Conditions,
-					utils.ConditionTypes.Idle,
-					utils.ConditionReasons.FinalizeFailed,
-					metav1.ConditionFalse,
-					"Error occurred during finalize",
-					ibu.Generation,
-				)
-			}
 		case string(utils.ConditionReasons.FinalizeFailed):
 			nextReconcile, err = r.handleFinalizeFailure(ctx, ibu)
 		}
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -205,10 +205,28 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 		switch idleCondition.Reason {
 		case string(utils.ConditionReasons.Aborting):
 			nextReconcile, err = r.handleAbort(ctx, ibu)
+			if err != nil {
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.ConditionTypes.Idle,
+					utils.ConditionReasons.AbortFailed,
+					metav1.ConditionFalse,
+					"Error occurred during abort",
+					ibu.Generation,
+				)
+			}
 		case string(utils.ConditionReasons.AbortFailed):
 			nextReconcile, err = r.handleAbortFailure(ctx, ibu)
 		case string(utils.ConditionReasons.Finalizing):
 			nextReconcile, err = r.handleFinalize(ctx, ibu)
+			if err != nil {
+				utils.SetStatusCondition(&ibu.Status.Conditions,
+					utils.ConditionTypes.Idle,
+					utils.ConditionReasons.FinalizeFailed,
+					metav1.ConditionFalse,
+					"Error occurred during finalize",
+					ibu.Generation,
+				)
+			}
 		case string(utils.ConditionReasons.FinalizeFailed):
 			nextReconcile, err = r.handleFinalizeFailure(ctx, ibu)
 		}
--- a/controllers/idle_handlers.go
+++ b/controllers/idle_handlers.go
@@ -126,10 +126,12 @@ func (r *ImageBasedUpgradeReconciler) cleanupUnbootedStateroots() error {
 		return err
 	}
 
+	bootedStateroot := ""
 	undeployIndices := make([]int, 0)
 	undeployStateroots := make([]string, 0)
 	for index, deployment := range status.Deployments {
 		if deployment.Booted {
+			bootedStateroot = deployment.OSName
 			continue
 		}
 		undeployIndices = append(undeployIndices, index)
@@ -137,7 +139,7 @@ func (r *ImageBasedUpgradeReconciler) cleanupUnbootedStateroots() error {
 	}
 	// since undeploy shifts the order, undeploy in the reverse order
 	for i := len(undeployIndices) - 1; i >= 0; i-- {
-		_, err = r.Executor.Execute("ostree", "admin", "undeploy", fmt.Sprint(i))
+		_, err = r.Executor.Execute("ostree", "admin", "undeploy", fmt.Sprint(undeployIndices[i]))
 		if err != nil {
 			return fmt.Errorf("ostree undeploy %d failed: %w", i, err)
 		}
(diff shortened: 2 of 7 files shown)


controllers/idle_handlers.go
```diff
@@ -22,29 +22,114 @@ import (
...
+		r.Log.Error(err, "failed to cleanup precaching resources")
+	}
+	if _, err := r.BackupRestore.CleanupBackups(ctx); err != nil {
+		successful = false
+		r.Log.Error(err, "failed to cleanup backups")
+	}
+	if err := r.BackupRestore.DeleteOadpOperator(ctx, backuprestore.OadpNs); err != nil {
+		successful = false
+		r.Log.Error(err, "failed to delete OADP operator")
+	}
```
sudomakeinstall2, line 90: @Missxiaoguo can we call run these cleanups even if backup has not been called?

Missxiaoguo: Yes, without oadp/backup, nothing to delete

Missxiaoguo: Probably not for the CleanupBackups. The function queries the backup CR, if the OADP is not installed and the CRD is not available, it will fail. Need to add a check in the function see https://github.com/Missxiaoguo/lifecycle-agent/blob/main/internal/backuprestore/backup.go#L276-L280

controllers/ibu_controller.go
```diff
@@ -200,10 +200,28 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 		switch idleCondition.Reason {
 		case string(utils.ConditionReasons.Aborting):
 			nextReconcile, err = r.handleAbort(ctx, ibu)
+			if err != nil {
```
jc-rh, line 203: Status update is better done inside the stage handler functions. Also, returning error will make it discard any CR update and redo the same reconcile. Return nil if we want to give up on aborting.

sudomakeinstall2: Do you think we should retry the aborting?

sudomakeinstall2: and if we want to retry, when we should give up and go to failed state?

jc-rh: Exactly, it's hard to say when to retry and for how many times or how long. For now, we can just give up on first error and improve error handling later, when we know better what kind of error is likely transient. The plan was to requeue with an interval so handleAbortFailure or handleFinalizeFailure will be triggered. Those two will check for some kind of "manual cleanup done" signal so that the IBU can still go back to idle.

controllers/idle_handlers.go
```diff
@@ -18,103 +18,187 @@ package controllers
...
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-func (r *ImageBasedUpgradeReconciler) cleanupUnbootedStateroot(stateroot string) error {
+func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	r.Log.Info("Starting handleAbort")
+
+	if r.cleanup(ctx, false, ibu) {
+		r.Log.Info("Finished handleAbort")
+		return doNotRequeue(), nil
```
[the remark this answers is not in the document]
jc-rh: That's done by https://github.com/openshift-kni/lifecycle-agent/blob/ec40a84e6273e77bed7345f7cae6ddd8eb534634/controllers/ibu_controller.go#L216 for all abort/finalize cases

controllers/idle_handlers.go
```diff
@@ -18,103 +18,205 @@ package controllers
...
+		successful = false
+		r.Log.Error(errors.New("failed to delete all the backup CRs"), "")
+	}
+	if err := r.BackupRestore.DeleteOadpOperator(ctx, backuprestore.OadpNs); err != nil {
+		successful = false
+		r.Log.Error(err, "failed to delete OADP operator")
+	}
+	if err := cleanupIBUFiles(); err != nil {
+		successful = false
+		r.Log.Error(err, "failed to cleanup ibu files")
```
tsorya, line 115: Maybe we can provide all those errors in status? User should not go to logs in order to see failures he can actually see in status? no?

controllers/idle_handlers.go
```diff
@@ -18,103 +18,208 @@ package controllers
...
 }
 
-//nolint:unparam
-func (r *ImageBasedUpgradeReconciler) handleFinalizeFailure(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
-
-	// TODO actual steps
-	// If succeeds, return doNotRequeue
-	return doNotRequeue(), nil
-}
+func removeStateroot(executor ops.Execute, stateroot string) error {
```
donpenney, line 214: Having this piece in a separate function concerns me a little, due to the risk of someone calling it without having undeployed all the deployments in that stateroot. If there are multiple places where we may be wiping the stateroot, include the undeploy commands here. If not, if only cleanupUnbootedStateroot would be called, then have this step inline in that function

controllers/idle_handlers.go
```diff
@@ -18,103 +18,208 @@ package controllers
...
+	}
+	if err := cleanupIBUFiles(); err != nil {
+		handleError(err, "failed to cleanup ibu files.")
+	}
+
+	return successful, error_message
+}
+
+// cleanupStateroot cleans all unbooted stateroots or desired stateroot
+// depending on allUnbootedStateroots argumennt
```
donpenney, line 125: nit: typo in comment

donpenney (changes requested):

controllers/idle_handlers.go
sudomakeinstall2: There are multiple places, and can't undeploy here. because all the deployments of the stateroot should be undeployed before removing stateroot files.
I will copy it back into the two functions.

donpenney: But you could have a function that wipes a specified stateroot like:
```
func removeStateroot(args, stateroot) {
    // 1. Get the list of deployments in specified stateroot
    // 2. Undeploy the deployments
    // 3. Wipe out the stateroot
}
```

and have that function called from elsewhere. Or you can have a single removeUnbootedStateroots that also determines the stateroot to be deleted, like you're doing here... because we're never going to be removing the booted stateroot, after all.

But it's unsafe to delete the stateroot without first undeploying. So either they're in the same function, or you do a check before deleting the stateroot to ensure it is clean. My preference would be to have them in the same function.

sudomakeinstall2: this is what `cleanupUnbootedStateroot(stateroot)` does.
There is also `cleanupUnbootedStateroots()` that does all this for all stateroots.

controllers/idle_handlers.go
```diff
@@ -110,11 +83,132 @@ func (r *ImageBasedUpgradeReconciler) handleFinalizeFailure(ctx context.Context,
...
+	bootedStateroot := ""
+	staterootsToRemove := make([]string, 0)
+	// since undeploy shifts the order, undeploy in the reverse order
+	for i := len(status.Deployments) - 1; i >= 0; i-- {
+		deployment := &status.Deployments[i]
+		if deployment.Booted {
+			bootedStateroot = deployment.OSName
+			continue
+		}
+		if err := r.OstreeClient.Undeploy(i); err != nil {
```
donpenney, line 159: This is actually undeploying all unbooted deployments, not just those in unbooted stateroot(s). You need to create the list of stateroots to be deleted first (or determine the booted stateroot). Then you can check that deployment to see if it's in the booted stateroot, as is done in the cleanupUnbootedStateroot function below.

Why not have this function just collect the list of unbooted stateroots, then call cleanupUnbootedStateroot for each?

sudomakeinstall2: > Why not have this function just collect the list of unbooted stateroots, then call cleanupUnbootedStateroot for each?

why though? that function query and iterates the deployments again. not efficient?

What is wrong with undeploying an unbooted deployment? (apart from name of the function)

sudomakeinstall2: this is called only in finalize. how could another deployment be in the stateroot we just created?

donpenney: Calling the other function removes code duplication. The difference between the two is one deletes all unbooted stateroots, the other just a specified stateroot. I could ask why not just have the one function that deletes all unbooted stateroots. When would you need to differentiate deleting one vs multiple? Why have two separate functions?

As for the behaviour here, I assume deleting secondary deployments from the booted stateroot was not the intent, and is therefore a bug. Being ok with the bug's behaviour doesn't mean it shouldn't be addressed.

controllers/idle_handlers.go
```diff
@@ -110,11 +83,128 @@ func (r *ImageBasedUpgradeReconciler) handleFinalizeFailure(ctx context.Context,
...
+	errorMessage := ""
+
+	var handleError = func(err error, msg string) {
+		successful = false
+		r.Log.Error(err, msg)
+		errorMessage += msg + " "
+	}
+
+	if err := r.cleanupStateroots(allUnbootedStateroots, ibu); err != nil {
+		handleError(err, "failed to cleanup stateroots.")
```
donpenney, line 103: nit: This would break the linter rules for a fmt.Errorf, which does not allow the message to end in punctuation. We should try to follow the same rules for consistency.

I guess the intent here is that we're running all steps, regardless of failure, and potentially building a single error message with multiple components, so the period is intended to separate those?

donpenney (commented):
/lgtm

controllers/idle_handlers.go
sudomakeinstall2: yes

PR #144, opened 2023-12-04
health check

pixelsoccupied:
Poll in parallel to see if ClusterOperators + MachineConfigPool + ClusterServiceVersion are all ready

### DIFF
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -23,20 +23,19 @@ import (
 	"strings"
 	"time"
 
-	v1 "k8s.io/api/core/v1"
-
-	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/openshift-kni/lifecycle-agent/internal/healthcheck"
 
 	"path/filepath"
+
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
-
 	lcautils "github.com/openshift-kni/lifecycle-agent/utils"
-
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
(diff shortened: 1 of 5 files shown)


internal/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,125 @@
...
+
+var (
+	pollInterval = 3 * time.Second
+	pollTimeout  = 30 * time.Minute
+)
+
+func ClusterServiceVersionReady(c client.Client, l logr.Logger) error {
+	l.Info("Waiting for all ClusterServiceVersion (csv) to be ready")
+
+	err := wait.PollUntilContextTimeout(context.Background(), pollInterval, pollTimeout, true, isClusterServiceVersionReady(c, l))
```
browsell, line 27: With each step having a separate poll, this could take 90 mins to timeout. Consider having a single poll

pixelsoccupied: Updated to make it parallel + max 15min of wait for each

controllers/upgrade_handlers.go
```diff
@@ -301,6 +298,66 @@ func (r *ImageBasedUpgradeReconciler) startBackup(ctx context.Context, backupCRs
...
+	err := healthChecks(r)
+	if err != nil {
+		upgradeFailedStatus(ibu, err.Error())
+		return requeueWithError(err)
+	}
+
+	return doNotRequeue(), nil
+}
+
+func healthChecks(r *ImageBasedUpgradeReconciler) error {
```
donpenney, line 313: Can this be a common utility function, rather than ImageBasedUpgradeReconciler? I may be able to use this in SeedGen, as well

controllers/upgrade_handlers.go
```diff
@@ -301,6 +298,66 @@ func (r *ImageBasedUpgradeReconciler) startBackup(ctx context.Context, backupCRs
 	return doNotRequeue(), nil
 }
 
+// nolint:unparam
+func (r *ImageBasedUpgradeReconciler) postPivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	r.Log.Info("Starting health check for different components")
+	err := healthChecks(r)
+	if err != nil {
+		upgradeFailedStatus(ibu, err.Error())
+		return requeueWithError(err)
```
Missxiaoguo, line 307: As it fails, you should stop requeue from this level. 
```suggestion
		return doNotRequeue(),nil
```

internal/common/utils.go
```diff
@@ -71,3 +75,9 @@ func CopyOutsideChroot(src, dest string) error {
 func GetStaterootPath(osname string) string {
 	return fmt.Sprintf("/ostree/deploy/%s", osname)
 }
+
+// FuncTimer check execution time
+func FuncTimer(now time.Time, name string, r logr.Logger) {
```
donpenney, line 80: nit: "start" would be a better name for the parameter than "now", since at the point where this is getting called, this time arg isn't actually "now", it's some point in the past. ie. "time elapsed since start"

donpenney (commented):
/lgtm

controllers/upgrade_handlers.go
pixelsoccupied: Updated should be easy to use now

internal/common/utils.go
pixelsoccupied: Updated

PR #149, opened 2023-12-05
OCPEDGE-664: Pregenerate kubelet certs to skip post-pivot CSR approvals

mresvanis:
This PR adds the following changes:

- adds `hostname` to `ClusterInfo`, in order to be able to replace the seed hostname with the target hostname in certs (in this PR) and static pod manifests (in a future PR)
- adds recert configuration params to replace the seed hostname with the target hostname in the `kubelet-client.pem` and `kubelet-server.pem`, with the goal of having those certs ready to go and skip the kubelet CSR generation (and hence the manual approval of the kubelet CSRs) during the post-pivot flow
- removes the post-pivot step of the kubelet CSR approvals 

### Skipping kubelet CSRs

According to how kubelet behaves, see [here](https://github.com/kubernetes/kubernetes/blob/930022cf267c2e0269fbd9be4be3fd83448e6419/pkg/kubelet/certificate/kubelet.go#L97-L103) and [here](https://github.com/kubernetes/client-go/blob/master/util/certificate/certificate_manager.go#L609-L652), if the respective certificates' IP addresses, DNSs and the subject orgs stay the same, then kubelet should be happy and it doesn't issue new CSRs. Thus we won't have to manually approve any kubelet CSRs during post-pivot.

### DIFF
--- a/ibu-imager/clusterinfo/clusterinfo.go
+++ b/ibu-imager/clusterinfo/clusterinfo.go
@@ -69,12 +69,15 @@ func (m *InfoClient) CreateClusterInfo(ctx context.Context) (*ClusterInfo, error
 		return nil, err
 	}
 
-	ip, err := m.getNodeInternalIP(ctx)
+	node, err := utils.GetSNOMasterNode(ctx, m.client)
 	if err != nil {
 		return nil, err
 	}
-
-	hostname, err := m.getNodeHostname(ctx)
+	ip, err := m.getNodeInternalIP(*node)
+	if err != nil {
+		return nil, err
+	}
+	hostname, err := m.getNodeHostname(*node)
 	if err != nil {
 		return nil, err
 	}
@@ -124,31 +127,21 @@ func (m *InfoClient) GetConfigMapData(ctx context.Context, name, namespace, key 
 }
 
 // TODO: add dual stuck support
-func (m *InfoClient) getNodeInternalIP(ctx context.Context) (string, error) {
-	node, err := utils.GetSNOMasterNode(ctx, m.client)
-	if err != nil {
-		return "", err
-	}
+func (m *InfoClient) getNodeInternalIP(node corev1.Node) (string, error) {
 	for _, addr := range node.Status.Addresses {
 		if addr.Type == corev1.NodeInternalIP {
 			return addr.Address, nil
 		}
 	}
-
 	return "", fmt.Errorf("failed to find node internal ip address")
 }
 
-func (m *InfoClient) getNodeHostname(ctx context.Context) (string, error) {
-	node, err := utils.GetSNOMasterNode(ctx, m.client)
-	if err != nil {
-		return "", err
-	}
+func (m *InfoClient) getNodeHostname(node corev1.Node) (string, error) {
 	for _, addr := range node.Status.Addresses {
 		if addr.Type == corev1.NodeHostName {
 			return addr.Address, nil
 		}
 	}
-
 	return "", fmt.Errorf("failed to find node hostname")
 }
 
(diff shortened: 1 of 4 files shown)


omertuc:
Where in LCA do we delete `/var/lib/kubelet`, as dictated by the hostname change procedure?

Or did we forget to do that?

If we didn't forget to do that, then this PR has to make sure that the certs don't get deleted

tsorya:
> Where in LCA do we delete `/var/lib/kubelet`, as dictated by the hostname change procedure?
> 
> Or did we forget to do that?
> 
> If we didn't forget to do that, then this PR has to make sure that the certs don't get deleted, when we do that

We don't need it so we don't delete. On what it influences?

omertuc:
> > Where in LCA do we delete `/var/lib/kubelet`, as dictated by the hostname change procedure?
> > Or did we forget to do that?
> > If we didn't forget to do that, then this PR has to make sure that the certs don't get deleted, when we do that
> 
> We don't need it so we don't delete. On what it influences?

What is "it"?

ibu-imager/clusterinfo/clusterinfo.go
```diff
@@ -73,6 +74,11 @@ func (m *InfoClient) CreateClusterInfo(ctx context.Context) (*ClusterInfo, error
 		return nil, err
 	}
 
+	hostname, err := m.getNodeHostname(ctx)
```
tsorya, line 77: Thinking maybe better bringing node object once and provide it to getNodeInternalIP and getNodeHostname ?

tsorya:
> > > Where in LCA do we delete `/var/lib/kubelet`, as dictated by the hostname change procedure?
> > > Or did we forget to do that?
> > > If we didn't forget to do that, then this PR has to make sure that the certs don't get deleted, when we do that
> > 
> > 
> > We don't need it so we don't delete. On what it influences?
> 
> What is "it"?

Not deleting folder currently doesn't influence anything, so my question is why we should delete it? what does it give us?

ibu-imager/clusterinfo/clusterinfo.go
mresvanis: Agreed, I also thought of that when adding this method, but I thought it adds more complexity than saves time. But let me try and refactor this piece, we should be able to do better indeed :)

omertuc:
> > > > Where in LCA do we delete `/var/lib/kubelet`, as dictated by the hostname change procedure?
> > > > Or did we forget to do that?
> > > > If we didn't forget to do that, then this PR has to make sure that the certs don't get deleted, when we do that
> > > 
> > > 
> > > We don't need it so we don't delete. On what it influences?
> > 
> > 
> > What is "it"?
> 
> Not deleting folder currently doesn't influence anything, so my question is why we should delete it? what does it give us?

It's what the node team said we should do if we wish to change the hostname. We should follow their procedures as closely as we can. I don't know why they prescribed this step, it's hard for us to confidently say it doesn't influence anything

tsorya:
> > > > > Where in LCA do we delete `/var/lib/kubelet`, as dictated by the hostname change procedure?
> > > > > Or did we forget to do that?
> > > > > If we didn't forget to do that, then this PR has to make sure that the certs don't get deleted, when we do that
> > > > 
> > > > 
> > > > We don't need it so we don't delete. On what it influences?
> > > 
> > > 
> > > What is "it"?
> > 
> > 
> > Not deleting folder currently doesn't influence anything, so my question is why we should delete it? what does it give us?
> 
> Deleting the folder what the node team said we should do if we wish to change the hostname. We should follow their procedures as closely as we can. I don't know why they prescribed this step, it's hard for us to confidently say it doesn't influence anything

I would assume we should ask one more time, as this does nothing. Doing things that do nothing bring more question than not doing them as for me. Our ci runs without it as far as i remember

ibu-imager/clusterinfo/clusterinfo.go
mresvanis: Implemented your idea and it seems cleaner/clearer PTAL.

PR #151, opened 2023-12-06
WIP: Create seed generation doc

donpenney:


docs/seed-image-generation.md
```diff
@@ -0,0 +1,141 @@
...
+  seedAuth: <encoded authfile>
+  hubKubeconfig: <encoded kubeconfig>
+```
+
+### Creating the seedimage SeedGenerator CR
+
+The `seedimage` `SeedGenerator` CR allows the user to provide the following information:
+
+- `seedImage`: The pullspec (ie. registry/repo:tag) for the generated image
+- `recertImage`: (Optional) Allows user to specify the recert tool image to pass to ibu-imager
```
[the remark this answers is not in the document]
donpenney: Since this field is really dev-only and this doc is geared towards end-user, I'll drop it altogether from the doc. It is still referenced in HACKING.md

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+
+The seed SNO configuration has some pre-requisites:
+
+- The CPU topology must align with the target SNO(s).
+  - Same number of cores.
+  - Same tuned performance configuration (ie. reserved CPUs).
+  - FIPS enablement (? *TODO*: confirm)
+- Deployed in the same manner as target SNO(s).
+  - Same ACM/MCE version.
+    - Exception: Hub for seed SNO must not have extra ACM addons enabled (ie. observability). The LCA orchestration confirms that no such addons are present on the seed SNO as part of its system config validation.
```
pixelsoccupied, line 38: Should we mention some info on when seed SNO is deployed with ZTP? i.e when deployed with ZTP we should just  comment our siteconfig from kustomize to get the best result

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+- OADP operator must be deployed.
+- Container storage must be setup as shared between stateroots, such as with a separate partition.
+- Required dnsmasq configuration to support updating cluster name, domain, and IP from the seed image as part of IBU.
+
+### Shared Container Storage
+
+Image Based Upgrade (IBU) requires that container storage (`/var/lib/containers`) is setup to be shared between stateroots. This can be done by creating a separate partition for container storage when the node is installed. Please see this [blog article](https://cloud.redhat.com/blog/a-guide-to-creating-a-separate-disk-partition-at-installation-time) for more information on how this is done using a `MachineConfig`.
```
pixelsoccupied, line 45: We have some ZTP code from back in days to create disk part during install time with Siteconfig. Though it was originally used for a different purpose...maybe we can do parts of what's required? Improving/integrating this with ZTP would be pretty nice...

https://github.com/openshift-kni/cnf-features-deploy/blob/master/ztp/gitops-subscriptions/argocd/ImageRegistry.md

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+            Type=xfs
+            Options=defaults,prjquota
+
+            [Install]
+            RequiredBy=local-fs.target
+          enabled: true
+          name: var-lib-containers.mount
+```
+
+### Required dnsmasq Configuration
```
pixelsoccupied, line 96: After generating these files we have to now apply it to seed image (with or without PGT) and then finally launch the seedgen yeah?

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+The seed SNO configuration has some pre-requisites:
+
+- The CPU topology must align with the target SNO(s).
+  - Same number of cores.
+  - Same tuned performance configuration (ie. reserved CPUs).
+  - FIPS enablement (? *TODO*: confirm)
+- Deployed in the same manner as target SNO(s).
+  - Same ACM/MCE version.
+    - Exception: Hub for seed SNO must not have extra ACM addons enabled (ie. observability). The LCA orchestration confirms that no such addons are present on the seed SNO as part of its system config validation.
+- OADP operator must be deployed.
```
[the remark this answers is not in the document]
browsell: OADP must be installed.

docs/seed-image-generation.md
donpenney: The ZTP siteconfig feature is limited, only supporting a subset of the ignition storage fields, and doesn't support all the ones being set here. And in addition to the RH blog article, this machineconfig aligns with the method for partitioning /var/lib/containers described in the formal documentation here:
https://docs.openshift.com/container-platform/4.14/installing/installing_vsphere/installing-vsphere.html#installation-disk-partitioning_installing-vsphere

docs/seed-image-generation.md
donpenney: You need to either include this as site-policy in your gitops or manually apply the machineconfig. Otherwise, the IBU will not succeed, which is why this is now included as a validation test if you use the seed gen orchestration, and why I include this reference here. Once it is in a GA rhacm release, we can remove this and specify that rhacm release as a minimum requirement.

docs/seed-image-generation.md
romfreiman, line 39: and in case of fresh install?

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+- Container storage must be setup as shared between stateroots, such as with a separate partition.
+- Required dnsmasq configuration to support updating cluster name, domain, and IP from the seed image as part of IBU.
```
romfreiman, line 41: i would say it requires minimal version of ACM... since there is not standalone dnsmasq configuration

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+
+            [Install]
+            RequiredBy=local-fs.target
+          enabled: true
+          name: var-lib-containers.mount
+```
+
+### Required dnsmasq Configuration
+
+Image Based Upgrade (IBU) requires that dnsmasq is configured such that it is able to use updated cluster name, domain, and IP address, as these will all be different for the target SNO compared to the seed SNO. The dnsmasq configuration is managed by a `MachineConfig` created during installation by the assisted-installer named `50-master-dnsmasq-configuration`. The changes required for IBU are not yet available in a GA ACM release, however, so a workaround is needed to override this `MachineConfig` until such time as we can define a minimum ACM version requirement.
```
romfreiman, line 98: not sure we have to document it, no?

docs/seed-image-generation.md
```diff
@@ -0,0 +1,329 @@
...
+
+The Lifecycle Agent provides orchestration of IBU Seed Image generation via the `SeedGenerator` CRD. As part of orchestration, the LCA will:
+
+- Perform system configuration checks to ensure any required configuration is present.
+- Perform any necessary system cleanup prior to generating the seed image.
+- Launch the ibu-imager tool, which will:
+  - Shutdown the cluster operators
+  - Prepare seed image config
+  - Generate and publish the seed image
+  - Restore the cluster operators
```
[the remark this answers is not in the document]
donpenney: This is done today when the ibu-imager finishes generating the seed image. I don't mean "restore" in the sense of "backup and restore". Maybe "recover" would be clearer. It restarts kubelet, which relaunches the operators.

docs/seed-image-generation.md
donpenney: Not sure what you're asking here. We require that the seed SNO (and the target SNO) have the OADP operator deployed.

docs/seed-image-generation.md
donpenney: This can be removed once we can define a minimum ACM version.

docs/seed-image-generation.md
jc-rh: Technically it's not required for IBI use first time. It would still be required for the next IBU on the installed node though.

PR #152, opened 2023-12-06
Convert PrepGetSeedImage.sh script to golang

sudomakeinstall2:
Pull seed image and copy the image list file for precaching


Makefile
```diff
@@ -130,7 +130,7 @@ bashate: ## Run bashate.
 	hack/bashate.sh
 
 .PHONY: ci-job
-ci-job: common-deps-update generate fmt vet golangci-lint unittest shellcheck bashate update-bindata bundle-check imager-unittest
```
donpenney, line 133: You should also be able to drop the go-bindata import from the dependencymagnet file and clean up the vendor folder:
https://github.com/openshift-kni/lifecycle-agent/blob/main/pkg/dependencymagnet/doc.go#L10

As well, remove the other go-bindata components added by the original commit:
https://github.com/openshift-kni/lifecycle-agent/commit/dc7c5671a5f4ff4274e0b8d78b79a976c75bf8f4#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

From that, it looks like you may also be able to drop the build-machinery-go import from dependencymagnet.

cc: @jc-rh

sudomakeinstall2: thanks, updated

internal/common/consts.go
```diff
@@ -47,6 +47,8 @@ const (
 	LvmConfigDir                      = "lvm-configuration"
 	LvmDevicesPath                    = "/etc/lvm/devices/system.devices"
 	CABundleFilePath                  = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+
+	LcaNamespace = "openshift-lifecycle-agent"
```
pixelsoccupied, line 51: There's also `LcaPrecacheNamespace` ...do you mind removing that one just use this one?

sakhoury: +1

sakhoury (commented):
/lgtm

internal/common/consts.go
tsorya, line 51: can't we get it from ENV variable of k8s? 
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace

PR #153, opened 2023-12-06
Upgrade post pivot with extra manifest apply and OADP restore

Missxiaoguo:
The directories that store the manifests will be cleaned up once each recovery has completed.
i.e., /opt/du-extra-manifests
       /opt/OADP/secret
       /opt/OADP/dpa
       /opt/OADP/veleroRestore

Upgrade will fail when there's issue with creation of extramanifest, recovery of OADP configuration and failed restore CRs.

### DIFF
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -18,23 +18,20 @@ package controllers
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
-
-	"github.com/openshift-kni/lifecycle-agent/internal/healthcheck"
-
-	"path/filepath"
-
-	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
-	v1 "k8s.io/api/core/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
+	"github.com/openshift-kni/lifecycle-agent/internal/common"
+	"github.com/openshift-kni/lifecycle-agent/internal/extramanifest"
+	"github.com/openshift-kni/lifecycle-agent/internal/healthcheck"
+	v1 "k8s.io/api/core/v1"
+
 	lcautils "github.com/openshift-kni/lifecycle-agent/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
@@ -43,20 +40,6 @@ var (
 	// todo: this value might need adjusting later
 	defaultRebootTimeout = 60 * time.Minute
 )
-
-// simple custom error types to handle backup cases
-type (
-	BackupFailedError     string
-	BackupValidationError string
-)
-
-func (b BackupFailedError) Error() string {
-	return string(b)
-}
-
-func (b BackupValidationError) Error() string {
-	return string(b)
-}
 
 var isPrePivot = func(r *ImageBasedUpgradeReconciler, ibu *lcav1alpha1.ImageBasedUpgrade) (bool, error) {
 	currentStaterootName, err := r.RPMOstreeClient.GetCurrentStaterootName()
@@ -74,7 +57,7 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 	prePivot, err := isPrePivot(r, ibu)
 	if err != nil {
 		//todo: abort handler? e.g delete desired stateroot
-		upgradeFailedStatus(ibu, err.Error())
+		utils.SetUpgradeStatusFailed(ibu, err.Error())
 		return doNotRequeue(), nil
 	}
 
@@ -84,7 +67,7 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 		ctrlResult, err := r.prePivot(ctx, ibu)
 		if err != nil {
 			//todo: abort handler? e.g delete desired stateroot
-			upgradeFailedStatus(ibu, err.Error())
+			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			return doNotRequeue(), nil
 		}
 
@@ -99,16 +82,14 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 		if err != nil {
 			//todo: abort handler? e.g delete desired stateroot
 			r.Log.Error(err, "")
-			upgradeFailedStatus(ibu, err.Error())
+			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			return doNotRequeue(), nil
 		}
 	} else {
 		r.Log.Info("Pivot successful, starting post pivot steps")
 		ctrlResult, err := r.postPivot(ctx, ibu)
 		if err != nil {
-			//todo: abort handler? e.g delete desired stateroot
-			upgradeFailedStatus(ibu, err.Error())
-			return doNotRequeue(), nil
+			return requeueWithError(err)
 		}
 
 		// postPivot requested a requeue
@@ -117,19 +98,7 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 		}
 
 		r.Log.Info("Done handleUpgrade")
-
-		utils.SetStatusCondition(&ibu.Status.Conditions,
-			utils.GetCompletedConditionType(lcav1alpha1.Stages.Upgrade),
-			utils.ConditionReasons.Completed,
-			metav1.ConditionTrue,
-			"Upgrade completed",
-			ibu.Generation)
-		utils.SetStatusCondition(&ibu.Status.Conditions,
-			utils.GetInProgressConditionType(lcav1alpha1.Stages.Upgrade),
-			utils.ConditionReasons.Completed,
-			metav1.ConditionFalse,
-			"Upgrade completed",
-			ibu.Generation)
+		utils.SetUpgradeStatusCompleted(ibu)
 		return doNotRequeue(), nil
 	}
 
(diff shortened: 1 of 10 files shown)


Missxiaoguo:
/hold
Waiting for https://github.com/openshift-kni/lifecycle-agent/pull/144 to be done first

controllers/upgrade_handlers.go
```diff
@@ -259,45 +215,135 @@ func (r *ImageBasedUpgradeReconciler) prePivot(ctx context.Context, ibu *lcav1al
...
+	err = r.BackupRestore.RestoreOadpConfigurations(ctx)
+	if err != nil {
+		if backuprestore.IsBRStorageBackendUnavailableError(err) {
+			upgradeFailedStatus(ibu, err.Error())
+			return doNotRequeue(), nil
+		}
+		return requeueWithError(err)
+	}
+
+	// Handle restore by OADP
```
sudomakeinstall2, line 244: nit: I think comment and the log below it say one thing. Maybe the comment is not necessary.

internal/extramanifest/extramanifest.go
```diff
@@ -18,28 +18,54 @@ package extramanifest
...
+	"github.com/openshift-kni/lifecycle-agent/ibu-imager/ops"
+	"github.com/openshift-kni/lifecycle-agent/internal/common"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
 
-const extraManifestPath = "extra-manifests/"
+const extraManifestPath = "/opt/du-extra-manifests"
```
browsell, line 36: nit, drop the du prefix. This could be used for other use cases

internal/extramanifest/extramanifest.go
```diff
@@ -79,78 +109,31 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
...
-			manifest.SetResourceVersion(existingManifest.GetResourceVersion())
-			manifest.SetUID(existingManifest.GetUID())
-			if err := h.Client.Update(ctx, manifest); err != nil {
-				return err
-			}
-			h.Log.Info("Updated manifest", "name", manifest.GetName(), "namespace", manifest.GetNamespace())
-		} else {
-			h.Log.Info("Created manifest", "name", manifest.GetName(), "namespace", manifest.GetNamespace())
-		}
+	h.Log.Info("Applying DU extra manifests from path", "path", extraManifestPath)
```
browsell, line 123: Drop the DU

internal/extramanifest/extramanifest.go
```diff
@@ -79,78 +109,31 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
...
-			return nil, err
-		}
-		cms = append(cms, existingCm)
+	// Remove the extra manifests directory
+	if err := os.RemoveAll(common.PathOutsideChroot(extraManifestPath)); err != nil {
+		return err
 	}
-
-	return cms, nil
+	h.Log.Info("DU extra manifests path removed", "path", extraManifestPath)
```
browsell, line 135: Drop the DU

internal/backuprestore/backup_test.go
```diff
@@ -209,101 +209,52 @@ func TestTriggerBackup(t *testing.T) {
...
+			expectedBackupTracker: BackupTracker{
 				PendingBackups: []string{"backup1", "backup2", "backup3", "backup4"},
-			}, {
-				ProgressingBackups: []string{"backup5"},
-			}},
+			},
 		},
 		{
-			name: "Backups applied in the first group but have failed status",
+			name: "Backups applied but have failed/failed validation status",
```
sudomakeinstall2, line 234: is there a typo here?

controllers/upgrade_handlers.go
```diff
@@ -259,45 +215,135 @@ func (r *ImageBasedUpgradeReconciler) prePivot(ctx context.Context, ibu *lcav1al
...
-// startBackup manages backup flow and returns with possible requeue
-func (r *ImageBasedUpgradeReconciler) startBackup(ctx context.Context, backupCRs []*velerov1.Backup) (ctrl.Result, error) {
-	// sort backupCRs by wave-apply
-	sortedBackupGroups, err := backuprestore.SortByApplyWaveBackupCrs(backupCRs)
+func (r *ImageBasedUpgradeReconciler) postPivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	r.Log.Info("Checking cluster health")
+	// TODO: cluster health check
+
+	// Restore extra manifests
+	r.Log.Info("Applying DU extra manifests")
```
browsell, line 223: Remove DU

internal/extramanifest/extramanifest.go
```diff
@@ -79,78 +109,31 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
...
-			if err := h.Client.Update(ctx, manifest); err != nil {
-				return err
-			}
-			h.Log.Info("Updated manifest", "name", manifest.GetName(), "namespace", manifest.GetNamespace())
-		} else {
-			h.Log.Info("Created manifest", "name", manifest.GetName(), "namespace", manifest.GetNamespace())
-		}
+	h.Log.Info("Applying DU extra manifests from path", "path", extraManifestPath)
+	args := []string{"--kubeconfig", common.KubeconfigFile, "apply", "-f"}
+	_, err = executor.Execute("oc", append(args, extraManifestPath)...)
```
sudomakeinstall2, line 125: can we use the client defined in `ImageBasedReconciler`?

Missxiaoguo: Yes, we are using the client in ImageBasedReconciler. In the upgrade handler, `r.Executor` is passed into this function.

sudomakeinstall2: oh sorry I meant the runtime client. I see this handler has also a client. Can we use that instead of `oc`?

Missxiaoguo: I see what you mean! To use runtime client, you will need proper RBAC in order to create the resource. The problem is that LCA doesn't have the knowledge of the KIND of these resources. We cannot assume we will ony have sriovnetwork, sriovnetworknodepolicies etc in the extra manfiests configmap.

browsell: yes we cannot assume the KIND and LCA should not care, it needs to be agnostic.

pixelsoccupied: @Missxiaoguo this looks like job for client-go's out-of-cluster feature see https://github.com/kubernetes/client-go/blob/master/examples/out-of-cluster-client-configuration/main.go 

Should be a lot faster than using pure exec

donpenney: @pixelsoccupied wouldn't that still have the RBAC requirements?

pixelsoccupied: i dont think so! Generally when a kubeconfig is not specified it falls back to InClusterConfig i.e SA...which in-turn needs RBAC. But if we pass KUBECONFIG (like for kubectl or out-of-cluster)...that requirement should be skipped. But please double check!

Missxiaoguo: @pixelsoccupied Sure! Let me give a shot!

controllers/upgrade_handlers.go
```diff
@@ -158,47 +135,26 @@ func (r *ImageBasedUpgradeReconciler) rebootToNewStateRoot() error {
...
+	r.Log.Info("Handling backups with OADP operator")
+	ctrlResult, err := r.handleBackup(ctx, ibu)
 	if err != nil {
-		if k8serrors.IsInvalid(err) {
+		if backuprestore.IsBRFailedError(err) {
+			upgradeFailedStatus(ibu, err.Error())
+			return doNotRequeue(), nil
+		}
+		if backuprestore.IsBRFailedValidationError(err) || backuprestore.IsBRNotFoundError(err) {
 			utils.SetStatusCondition(&ibu.Status.Conditions,
```
tsorya, line 146: I wonder if we can create ready templates for each status and not to provide full description each time?
It just very long code each time when we set status, making code very difficult to read 
Not relevant to this PR i assume

internal/extramanifest/extramanifest.go
```diff
@@ -79,78 +109,31 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
 	return nil
 }
 
-// ApplyExtraManifestsFromDir applies the extra manifests from the given directory
-func (h *EMHandler) ApplyExtraManifestsFromDir(ctx context.Context, fromDir string) error {
-	manifestDir := filepath.Join(fromDir, extraManifestPath)
-	manifestYamls, err := os.ReadDir(manifestDir)
+// ApplyExtraManifests applies the extra manifests from the preserved extra manifests directory
+func (h *EMHandler) ApplyExtraManifests(ctx context.Context, executor ops.Execute) error {
```
tsorya, line 113: I wonder, we are doing it in post pivot phase too. 
applying all manifests from /opt/openshift/extra_manifests folder. Why we need same logic twice?

internal/extramanifest/extramanifest.go
tsorya: There is https://github.com/openshift/library-go/blob/master/pkg/assets/create/creater.go for such things as far as i remember, at least ocp clusterbootstrap uses this way with admin kubeconfig

tsorya:
@browsell I think i am missing something, here we apply additional manifest that we actually do the same 
https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/postpivot/postpivot.go#L203.
I am not sure i understand why we are trying to do same thing in both places?

internal/extramanifest/extramanifest.go
donpenney: I believe this would be the day2 bits that we'd want applied after the system is stable, similar to the day2 policies for ztp gitops, vs those that need to be applied during the earlier stage as part of bringing the system back up

internal/extramanifest/extramanifest.go
Missxiaoguo: Got this working with client-go. Will push an update

internal/extramanifest/extramanifest.go
tsorya: Hmm, 
Maybe we need to delete https://github.com/openshift-kni/lifecycle-agent/blob/main/ibu-imager/postpivot/postpivot.go#L203 and use code here

internal/extramanifest/extramanifest.go
```diff
@@ -18,28 +18,55 @@ package extramanifest
...
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
 
-const extraManifestPath = "extra-manifests/"
+const ExtraManifestPath = "/opt/extra-manifests"
```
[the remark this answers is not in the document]
Missxiaoguo: It may not be suitable to have this in the /opt/openshift folder since it is deleted once the systemd service completes its task. However, these manifests are applied at a later stage, after the cluster has achieved stability, by the LCA.
These manifests are user provided through the `spec.extraManifests` in IBU CR. Unlike the manifests applied in post-pivot.go, they serve distinct purposes and are applied at different times. From my perspective, I don't anticipate any issues as long as proper cleanup is done once the task is completed.

Missxiaoguo:
Wait for https://github.com/openshift-kni/lifecycle-agent/pull/157 be merged first

controllers/utils/conditions.go
```diff
@@ -240,3 +240,45 @@ func GetPreviousStage(stage lcav1alpha1.ImageBasedUpgradeStage) lcav1alpha1.Imag
 	}
 	return ""
 }
+
+// SetUpgradeStatusFailed updates the upgrade status to failed with message
```
donpenney, line 244: nit: Should we include the call to updateStatus in these functions as well? These just set the condition in-memory, I believe, with a call to r.updateStatus at the end of the reconciler. Maybe the updateStatus should be moved here as well (as UpdateUpgradeStatus, or UpdateIBUStatus), and called from these functions so that they're updating in-memory and in-db?

Doesn't have to be in this PR. It can be a discussion point and a later PR

internal/backuprestore/backup.go
```diff
@@ -310,29 +295,37 @@ func (h *BRHandler) ExportOadpConfigurationToDir(ctx context.Context, toDir, oad
...
 
-			filePath := filepath.Join(toDir, oadpSecretPath, secretName+".yaml")
-			if err := writeSecretToFile(storageSecret, filePath); err != nil {
-				return err
-			}
+		storageSecret.SetUID("")
+		storageSecret.SetResourceVersion("")
+
+		filePath := filepath.Join(toDir, oadpSecretPath, secretName+".yaml")
+		if err := writeSecretToFile(storageSecret, filePath); err != nil {
```
donpenney, line 324: There's a MarshalToFile function in utils (https://github.com/openshift-kni/lifecycle-agent/blob/main/utils/utils.go#L27) that I think can replace writeSecretToFile. We should probably change the mod to 0o600, though.
This can be a separate PR

internal/backuprestore/common.go
```diff
@@ -52,28 +54,17 @@ const (
...
-// Constants for restore phase
-const (
-	RestorePending    RestorePhase = "RestorePending"
-	RestoreFailed     RestorePhase = "RestoreFailed"
-	RestoreCompleted  RestorePhase = "RestoreCompleted"
-	RestoreInProgress RestorePhase = "RestoreInProgress"
-)
-
 var (
+	hostDir = "/host"
```
donpenney, line 66: nit: controllers/utils/constants.go has a Host constant already that can be used, although I see you're using a var so you can override it in tests. You could change this to initialize using the constant, then. But maybe we need to make that const a var in utils?

donpenney (commented):
/lgtm

internal/backuprestore/common.go
Missxiaoguo: Exactly, to use the one in controllers/utils/constants.go, I need to change it to var.  I didn't do that because I was afraid that the common variable be accidentally overridden.

PR #158, opened 2023-12-08
Seed Generator Orchestration - post-imager recovery

donpenney:


controllers/utils/constants.go
```diff
@@ -1,15 +1,29 @@
 package utils
 
-// IBUName defines the valid name of the CR for the controller to reconcile
+import (
+	"path/filepath"
+)
+
 const (
+	LcaNamespace string = "openshift-lifecycle-agent"
```
browsell, line 8: Cleanup for another PR. Can we use the one already defined in https://github.com/openshift-kni/lifecycle-agent/blob/main/internal/common/consts.go#L51 or we could pass this in as an environment variable.

donpenney: Didn't realize this one had gotten created. I was originally getting it from an env var, since it wasn't in an accessible constant, so thought I'd create one

donpenney: Done

main/main.go
```diff
@@ -143,6 +146,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := initSeedGen(context.TODO(), mgr.GetClient(), &setupLog); err != nil {
```
sakhoury, line 149: Perhaps we can create a context variable and pass the same context to both the `initIBU` and `initSeedGen` functions?

sakhoury (commented):
/lgtm

PR #159, opened 2023-12-11
Fix reconciliation for upgrade

Missxiaoguo:
Status decision is made inside of pre-pivot and post-pivot functions. The caller should simply return what it receives from these functions.

### DIFF
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -64,45 +64,11 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 	// WARNING: the pod may not know if we are boot loop (for now)
 	if prePivot {
 		r.Log.Info("Starting pre pivot steps and will pivot to new stateroot with a reboot")
-		ctrlResult, err := r.prePivot(ctx, ibu)
-		if err != nil {
-			//todo: abort handler? e.g delete desired stateroot
-			utils.SetUpgradeStatusFailed(ibu, err.Error())
-			return doNotRequeue(), nil
-		}
-
-		// prePivot requested a requeue
-		if ctrlResult != doNotRequeue() {
-			return ctrlResult, nil
-		}
-
-		// Write an event to indicate reboot attempt
-		r.Recorder.Event(ibu, v1.EventTypeNormal, "Reboot", "System wil now reboot")
-		err = r.rebootToNewStateRoot()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


controllers/upgrade_handlers.go
```diff
@@ -230,8 +204,13 @@ func (r *ImageBasedUpgradeReconciler) postPivot(ctx context.Context, ibu *lcav1a
 		}
 		return requeueWithError(err)
 	}
+	if !result.IsZero() {
```
donpenney, line 207: What case is this handling? We're returning nil here, so I guess this means the restore didn't complete and we're expecting the reconciler to be called again? A comment here would be nice to have, as this isn't obvious (to me, at least)

controllers/upgrade_handlers.go
```diff
@@ -64,45 +64,11 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *lc
 	// WARNING: the pod may not know if we are boot loop (for now)
 	if prePivot {
 		r.Log.Info("Starting pre pivot steps and will pivot to new stateroot with a reboot")
-		ctrlResult, err := r.prePivot(ctx, ibu)
-		if err != nil {
-			//todo: abort handler? e.g delete desired stateroot
-			utils.SetUpgradeStatusFailed(ibu, err.Error())
```
donpenney, line 70: The prePivot/postPivot functions are now the only places where Upgrade gets set to Failed, I guess? So we need to ensure all failure point in these functions do so. Not related to this PR, maybe, but I think it would be good to have that well-commented in these functions to make that requirement clear.

donpenney:
As a side note, my comments are both minor and can be handled (or not) in a separate PR. Given that this is addressing an important fix, I'd like to get this merged.

/lgtm

controllers/upgrade_handlers.go
Missxiaoguo: Exactly. Sure, will add comment in a further PR.

controllers/upgrade_handlers.go
Missxiaoguo: Yes, all decisions(requeue+failures) should be made in one place which is inside of those functions. Will add comments in a further cleanup PR. 
As both completed and failed cases should stop the reconciliation (return doNotRequeue(), nil), with previous impelementation where the caller and called functions have mixed decisions, you cannot distingiush whether the result from called functions are failed or completed cases.

PR #162, opened 2023-12-11
Refactor: Move network config to clusterconfig

tsorya:


internal/clusterconfig/clusterconfig.go
```diff
@@ -47,12 +47,18 @@ const (
...
+	networkDir = "/opt/openshift/network-configuration"
 )
 
+var hostPath = common.Host
+var listOfNetworkFilesPaths = []string{
+	"/etc/hostname",
+	"/etc/NetworkManager/system-connections",
+}
+
 var (
```
pixelsoccupied, line 62: Consider moving the new vars in the var() block

pixelsoccupied (commented):
Just a tiny comment but otherwise the changes are straightforward and some nice refactoring...LGTM from me

internal/clusterconfig/clusterconfig.go
tsorya: right, missed it

donpenney (commented):
/lgtm

PR #163, opened 2023-12-11
Fix kernel arguments quoting

sudomakeinstall2:
Marshal kernel arguments to preserve quotations after reboot. Passing `--karg-append cgroup_no_v1=\"all\"` to ostree admin deploy will result in the `cgroup_no_v1=all` kernel argument. To ensure quotation marks are retained, use `--karg-append \"cgroup_no_v1=\\\"all\\\"\"`, resulting in the final kernel argument: `cgroup_no_v1="all"`.

\cc @tsorya


sudomakeinstall2:
\hold

PR #165, opened 2023-12-12
Create doc for using backup and restore

Missxiaoguo:
/cc @browsell @donpenney @pixelsoccupied

### DIFF
--- a/docs/backuprestore-with-oadp.md
+++ b/docs/backuprestore-with-oadp.md
@@ -1,0 +1,459 @@
+# Backup and restore with the Openshift API for Data Protection(OADP)
+
+## Overview
+
+The lifecycle Agent operator (LCA) provides functionality for backing up and restoring platform and application resources using OADP during image-based upgrades. During the upgrade stage, the LCA performs the following actions:
+
+Before the cluster is rebooted to the new stateroot:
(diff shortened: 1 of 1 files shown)


docs/backuprestore-with-oadp.md
```diff
@@ -0,0 +1,459 @@
...
+```
+anntotations:
+  lca.openshift.io/apply-wave: "1"
+```
+If the annotation is provided in the backup or restore CRs, they will be applied in increasing order based on the annotation value. The LCA will move on the next group of CRs only after completing the previous group with the same wave number.
+If no `lca.openshift.io/apply-wave` annotation is defined in the backup or restore CRs, which means no particular order is required, they will be applied all together. 
+
+## Install OADP and configure OADP on target cluster via ZTP GitOps
+
+ Install OADP via [GitOps ZTP pipeline](https://docs.openshift.com/container-platform/4.14/scalability_and_performance/ztp_far_edge/ztp-configuring-managed-clusters-policies.html).
```
donpenney, line 33: Should we have a section for non-gitops install? Or maybe a note here that the non-PGT bits can just be applied (minus the wave annotation), plus some site-specific customizations?

PR #169, opened 2023-12-13
Prevent patching seedImageRef while IBU is in progress

sudomakeinstall2:
- Add validation to prevent changes to spec.seedImageRef while IBU is in progress

Changing spec.seedImageRef while IBU is in progress can cause unexpected behaviour.
Validation has been added to IBU CR to prevent this.

Without `!has(oldSelf.spec.seedImageRef)` as the first part of rule we cannot
create the initial ibu that does not have `spec.seedImageRef`.

- Check seed image version as soon as possible

Check the seed image version against spec.seedImageRef.version as soon as possible in
`setupStateroot` to avoid unnecessary steps in case of an error

- Use MY_POD_NAMESPACE env variable to get LCA namespace

Remove LcaNamespace const variable and get LCA namespace using
MY_POD_NAMESPACE env variable to decouple LCA from a fixed namespace


omertuc:
Is there any reason where we would allow *any* patching at all, not just to this field?

sudomakeinstall2:
```
type ImageBasedUpgradeSpec struct {
	//+kubebuilder:validation:Enum=Idle;Prep;Upgrade;Rollback
	Stage            ImageBasedUpgradeStage `json:"stage,omitempty"`
	SeedImageRef     SeedImageRef           `json:"seedImageRef,omitempty"`
	AdditionalImages ConfigMapRef           `json:"additionalImages,omitempty"`
	OADPContent      []ConfigMapRef         `json:"oadpContent,omitempty"`
	ExtraManifests   []ConfigMapRef         `json:"extraManifests,omitempty"`
	RollbackTarget   string                 `json:"rollbackTarget,omitempty"`
}
```
Patching `spec.stage` is necessary, apart from that I think we can prevent patches to all other fields.
@Missxiaoguo

omertuc:
Great, that would be better IMO

browsell:
We should block all changes to the spec if the stage is not idle

browsell:
Please squash before merge

jc-rh:
> We should block all changes to the spec if the stage is not idle

OADP and extra manifest could come with the patch that moves it to the upgrade stage.

internal/common/consts.go
```diff
@@ -48,8 +48,8 @@ const (
 	LvmDevicesPath                    = "/etc/lvm/devices/system.devices"
 	CABundleFilePath                  = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
 
-	LcaNamespace        = "openshift-lifecycle-agent"
-	Host         string = "/host"
+	EnvLCANamespace = "MY_POD_NAMESPACE"
```
donpenney, line 51: Can we convert the const to a var that's initialized via os.GetEnv, rather than calling os.GetEnv everywhere?

sudomakeinstall2: I thought about it, but it is not that simple.
Can't use this `const LcaNamespace = os.getEnv("")` and I think having a global (not const) variable is not good.

donpenney: We've got a couple of cases of global vars, but ok. I'm not sure what the concern would be here. Having the global seems preferable to me, personally. Are you concerned about buggy code overwriting the var? or a store trampler somehow doing so?

sakhoury: +1 also not in favour of using global variables. An alternative approach may be that we can define a common utility function to retrieve the LCA Namespace:
```golang

const DefaultLCANamespace = "openshift-lifecycle-agent"
...
func GetLCANamespace() string {
	ns := os.Getenv(common.EnvLCANamespace)
	if ns == nil {
		return DefaultLCANamespace
	}
	return ns
}
```

sudomakeinstall2: dropped the commit from this PR. it was out of scope.

donpenney (commented):
/lgtm

PR #172, opened 2023-12-13
Some changes to backup&restore plugin

Missxiaoguo:
- Fail the upgrade if more than one DPA found
-  Wait until DPA is reconciled before checking storage backend and do not exit the storage check immediately if no storages found
- Use utils function to read/write resource from/to file
- Use hostPath from common
- Re-organize some logs

### DIFF
--- a/internal/backuprestore/backup.go
+++ b/internal/backuprestore/backup.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
+	"github.com/openshift-kni/lifecycle-agent/utils"
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
@@ -240,7 +241,7 @@ func (h *BRHandler) extractBackupFromConfigmaps(ctx context.Context, configmaps 
 	return backups, nil
 }
 
-// ExportOadpConfigurationToDir exports the OADP DataProtectionApplication CRs and required storage creds to a given location
+// ExportOadpConfigurationToDir exports the OADP DataProtectionApplication CR and required storage creds to a given location
 func (h *BRHandler) ExportOadpConfigurationToDir(ctx context.Context, toDir, oadpNamespace string) error {
 	dpaList := &unstructured.UnstructuredList{}
 	dpaList.SetGroupVersionKind(dpaGvkList)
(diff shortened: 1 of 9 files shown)


internal/backuprestore/backup.go
```diff
@@ -258,24 +259,37 @@ func (h *BRHandler) ExportOadpConfigurationToDir(ctx context.Context, toDir, oad
...
+	// even if there are more than one DPA created. We only want to write
+	// the reconciled DPA to a given directory if there are multiple found.
 	for _, dpa := range dpaList.Items {
 		// Unset uid and resource version
 		dpa.SetUID("")
 		dpa.SetResourceVersion("")
 
+		ok := isDPAReconciled(&dpa)
+		if !ok {
+			h.Log.Info("The DataProtectionApplication CR is not reconciled, skipping", "name", dpa.GetName())
```
donpenney, line 284: Should this be part of the health check?

Missxiaoguo: This is pre-pivot phase. Trying to dump the live DPA to new stateroot

Missxiaoguo: But as we talked, there will be some OADP checks in advance to make sure they are set.

donpenney: Yes, that's what I mean. Shouldn't we block the upgrade if there's a DPA that hasn't reconciled?

Missxiaoguo: This check here is meant to dump the Reconciled one only if there are multiple(although it's unlikely). In theory, only one DPA should be created in one oadp namespace. But OADP doesn't reject the creation of the second one, instead it just fails it (Reconciled:False).

donpenney: A scenario in which it seems reasonable to me to block the upgrade... An unexpected and unsupported configuration. As opposed to skipping past what presumably could be the desired and unreconciled DPA? ie. what if the DPA we need is truly not reconciled at this moment?

Missxiaoguo: > Shouldn't we block the upgrade if there's a DPA that hasn't reconciled?
Yes, we should block the upgrade if there's no DPA gets reconciled.  But if there is a reconciled DPA and another non-reconciled DPA, we should continue

Missxiaoguo: > As opposed to skipping past what presumably could be the desired and unreconciled DPA? 

This may be possible. Yeah, we can discuss this further. I thought as OADP doesn't block the creation of the second one, we should not block as long as there's a working one.

browsell: We only expect one, so more we should treat as an error.

Missxiaoguo: More validation enhancement before upgrade will be done in https://issues.redhat.com/browse/CNF-10656

donpenney (commented):
/lgtm

donpenney (commented):
/lgtm

PR #173, opened 2023-12-13
Changes required for IBI 

tsorya:
1. Moving seed manifest.json to /var/seed_data in order to allow reeding seed info manifest in IBI process
2. Delete ovn certs folders as part of seed creation and not just filter
   them from tar
3. Set hostname by writing value from hostname var in cluster info and
   not copying etc/hostname file
4. Move cluster id change to post pivot

### DIFF
--- a/ibu-imager/postpivot/postpivot.go
+++ b/ibu-imager/postpivot/postpivot.go
@@ -92,6 +92,10 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
 		return err
 	}
 
+	if err := utils.RunOnce("set_cluster_id", p.workingDir, p.log, p.setNewClusterID, ctx, client, clusterInfo); err != nil {
+		return err
+	}
+
 	// Restore lvm devices
 	if err := utils.RunOnce("recover_lvm_devices", p.workingDir, p.log, p.recoverLvmDevices); err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


ibu-imager/postpivot/postpivot.go
```diff
@@ -92,6 +92,10 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
 		return err
 	}
 
+	if err := utils.RunOnce("set_cluster_id", p.workingDir, p.log, p.setNewClusterID, ctx, client, clusterInfo); err != nil {
```
donpenney, line 95: Can this be done by recert at the same time it's handling the cluster naming? Right now, there's a window where the API is available, but the clusterID is that of the seed. @omertuc ?
In the meantime, maybe this can be the first change applied once the API is available

ibu-imager/postpivot/postpivot.go
```diff
@@ -327,3 +339,25 @@ func (p *PostPivot) changeRegistryInCSVDeployment(ctx context.Context, client ru
...
+func (p *PostPivot) setNewClusterID(ctx context.Context, client runtimeclient.Client, clusterInfo *clusterinfo.ClusterInfo) error {
+	p.log.Info("Set new cluster id")
+	clusterVersion := &v1.ClusterVersion{}
+	if err := client.Get(ctx, types.NamespacedName{Name: "version"}, clusterVersion); err != nil {
+		return err
+	}
+
+	patch := []byte(fmt.Sprintf(`{"spec":{"clusterID":"%s"}}`, clusterInfo.ClusterID))
+
+	p.log.Infof("Applying csv deploymet patch %s", string(patch))
```
donpenney, line 357: nit: typo in deployment

donpenney (commented):
/lgtm

PR #174, opened 2023-12-14
Also update conditions' `ObservedGeneration`

omertuc:
~There's an `ObservedGeneration` field on every condition within the
status, not just on the status itself.~

~LCA should make sure this field is kept up to date on its conditions.~

~If we don't do that, commands like `oc wait --for=condition=somecondition=somevalue`
will wait forever, as [apparently](https://github.com/kubernetes/kubernetes/pull/122319) they ignore conditions which have a generation
older than the resource's own `.metadata.generation`~

~Also added TODO to reconsider this change, I'm not sure how much such
blanket update of all conditions actually makes sense.~

EDIT: Reviews are understandably against this patch, please consider this PR to be more of an issue/suggestion and feel free to come up with better solutions to this problem

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -374,6 +374,13 @@ func validateStageTransition(ibu *lcav1alpha1.ImageBasedUpgrade, isAfterPivot bo
 
 func (r *ImageBasedUpgradeReconciler) updateStatus(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) error {
 	ibu.Status.ObservedGeneration = ibu.ObjectMeta.Generation
+
+	// TODO: Does a blanket update of all conditions even make sense, or should
+	// we be more selective?
+	for i := range ibu.Status.Conditions {
+		ibu.Status.Conditions[i].ObservedGeneration = ibu.ObjectMeta.Generation
+	}
+
 	err := common.RetryOnConflictOrRetriable(retry.DefaultRetry, func() error {
 		return r.Status().Update(ctx, ibu)
 	})


jc-rh:
Not sure if this makes sense. Our conditions should be updated with the generation that moves the stage field in the spec. Are you waiting for a condition from a previous stage for some reason?

controllers/ibu_controller.go
```diff
@@ -374,6 +374,13 @@ func validateStageTransition(ibu *lcav1alpha1.ImageBasedUpgrade, isAfterPivot bo
 
 func (r *ImageBasedUpgradeReconciler) updateStatus(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) error {
 	ibu.Status.ObservedGeneration = ibu.ObjectMeta.Generation
+
+	// TODO: Does a blanket update of all conditions even make sense, or should
+	// we be more selective?
+	for i := range ibu.Status.Conditions {
+		ibu.Status.Conditions[i].ObservedGeneration = ibu.ObjectMeta.Generation
```
mresvanis, line 381: We seem to be setting the `ObservedGeneration` properly when setting the status conditions via [this](https://github.com/openshift-kni/lifecycle-agent/blob/f3d164e4d31fe63e07c2214cc90efd6d1852c6c3/controllers/utils/conditions.go#L104). 

Perhaps you observed that `kubeclt/oc wait` for a specific status condition didn't work as expected because of an incorrect `ObservedGeneration` (my best guess is with the `Idle` condition)?

sudomakeinstall2:
I think we already update the condition generation every time we update the condition. Can you share the scenario where the condition generation did not get updated?

controllers/ibu_controller.go
jc-rh: Hmm, I guess it can happen whenever there is a spec change that doesn't trigger any condition updates. In theory, we probably should always update the observedGeneration field for the conditions related to the current desired stage.

omertuc: >my best guess is with the Idle condition)?

Yes my issue was that I was Idle aand the generation didn't match so wait for idle condition never ends

omertuc:
Feeling out of my depth on this one so consider this PR to be more an Issue than a PR

controllers/ibu_controller.go
sudomakeinstall2: @jc-rh I don't think there is any such changes, currently.

jc-rh: > @jc-rh I don't think there is any such changes, currently.

an OADP or extra manifest update while in prep?

jc-rh:
I took a stab at it: https://github.com/openshift-kni/lifecycle-agent/pull/175

PR #175, opened 2023-12-14
Update observedGeneration for conditions related to the current stage

jc-rh:


donpenney (commented):
/lgtm

sudomakeinstall2:
I am not sure how this helps. especially now that patches are disabled.

Missxiaoguo:
> I am not sure how this helps. especially now that patches are disabled. If ibu is in idle and user patches spec (not stage) does the generation get synced using this?

Do we have all fields disabled now for patch? Probably not for i.e.,additionalImages?

sudomakeinstall2:
@Missxiaoguo changes to oadpcontent, seedImageRef and extraManifests are blocked after transition to Prep.

PR #176, opened 2023-12-14
seedgen: Fix recert podman pull for disconn registries

leo8a:
In a totally disconn env, the recert image may be mirrored to the local (private) registry as well.

### DIFF
--- a/controllers/seedgen_controller.go
+++ b/controllers/seedgen_controller.go
@@ -413,7 +413,7 @@ func (r *SeedGeneratorReconciler) getRecertImagePullSpec(seedgen *seedgenv1alpha
 func (r *SeedGeneratorReconciler) pullRecertImagePullSpec(seedgen *seedgenv1alpha1.SeedGenerator) error {
 	recertImage := r.getRecertImagePullSpec(seedgen)
 
-	_, err := r.Executor.Execute("podman", "pull", recertImage)
+	_, err := r.Executor.Execute("podman", "pull", "--authfile", common.ImageRegistryAuthFile, recertImage)
 	if err != nil {
 		return fmt.Errorf("failed to pull recertImage (%s): %w", recertImage, err)
 	}


donpenney:
> assuming recert and seed image share creds for the same local registry

I don't think you can make that assumption. The auth file in the secret, as it stands today, only needs the write-auth for the repository to which the seed image is to be written. As well, if the seed SNO is setup to use a disconnected registry, it should already have the auth needed to pull the recert image from that registry, no? The only scenario in which you'd need to be able to specify a different auth file for this pull would be if the recert image was mirrored to a different registry than that used for installing the SNO, isn't it?

donpenney:
If you wanted to reuse the write-auth for pulling the recert image, I think you would need to merge it with the pull secrets for the cluster before writing it to file

leo8a:
> As well, if the seed SNO is setup to use a disconnected registry, it should already have the auth needed to pull the recert image from that registry, no?

this is precisely what fails in my disconn env (see below error). The disconnected cluster has the creds configured but the pulling of the recert image fails (cause code doesn't specify where to find those creds).

```
2023-12-14T18:02:33Z	ERROR	Reconciler error	{"controller": "seedgenerator", "controllerGroup": "lca.openshift.io", "controllerKind": "SeedGenerator", "SeedGenerator": {"name":"seedimage"}, "namespace": "", "name": "seedimage", "reconcileID": "1a241eaa-394f-4fd5-82cb-3b7652dd717b", "error": "failed to pull recert image: failed to pull recertImage (jumphost.inbound.vz.bos2.lab:8443/edge-infrastructure/recert:latest): Trying to pull jumphost.inbound.vz.bos2.lab:8443/edge-infrastructure/recert:latest...\nError: initializing source docker://jumphost.inbound.vz.bos2.lab:8443/edge-infrastructure/recert:latest: reading manifest latest in jumphost.inbound.vz.bos2.lab:8443/edge-infrastructure/recert: unauthorized: access to the requested resource is not authorized\n: exit status 125"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/opt/app-root/src/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/src/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/src/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227
``

we can use a different auth than the one used for pushing the seed, I don't have problems with that... but at least an authfile should be added to the podman pull command (if provided) otherwise it seems to fail for disconnected environments with above error

File: controllers/seedgen_controller.go [hunk 1]
```diff
@@ -413,7 +413,7 @@ func (r *SeedGeneratorReconciler) getRecertImagePullSpec(seedgen *seedgenv1alpha
 func (r *SeedGeneratorReconciler) pullRecertImagePullSpec(seedgen *seedgenv1alpha1.SeedGenerator) error {
 	recertImage := r.getRecertImagePullSpec(seedgen)
 
-	_, err := r.Executor.Execute("podman", "pull", recertImage)
+	_, err := r.Executor.Execute("podman", "pull", "--authfile", seedgenAuthFile, recertImage)
```
donpenney, line 416: I think you would need to get the .dockerconfigjson from the openshift-config/pull-secret secret and merge it with the seedgenAuth before writing the file. 

I guess I should have defined this with "--authfile /var/lib/kubelet/config.json", which should be the pull secrets for the cluster.

leo8a: Ok, for the moment, I think we can keep the fix simple by just adding the `"--authfile /var/lib/kubelet/config.json"` suggested... in any case, seems like a minor issue.

if a new / different creds is required for the recert image in the future, we can write the code for it

donpenney (approved):

PR #177, opened 2023-12-14
MGMT-16334: [Imager] Add selective cleanup based on successful stages

leo8a:
Adds a conditional cleanup to avoid the execution of certain cleanup steps that could cause issues.

/cc @tsorya @donpenney

### DIFF
--- a/ibu-imager/seedrestoration/seedcleanup.go
+++ b/ibu-imager/seedrestoration/seedcleanup.go
@@ -88,9 +88,12 @@ func (s *SeedRestoration) CleanupSeedCluster() error {
 	if s.recertSkipValidation {
 		s.log.Info("Skipping restoring crypto via recert tool")
 	} else {
-		if err := s.ops.RestoreOriginalSeedCrypto(s.recertContainerImage, s.authFile); err != nil {
-			s.log.Errorf("Error restoring certificates: %v", err)
-			errors = append(errors, err)
+		recertFilePath := filepath.Join(common.BackupChecksDir, "recert.done")
+		if _, err := os.Stat(recertFilePath); err == nil && !os.IsNotExist(err) {
+			if err := s.ops.RestoreOriginalSeedCrypto(s.recertContainerImage, s.authFile); err != nil {
+				s.log.Errorf("Error restoring certificates: %v", err)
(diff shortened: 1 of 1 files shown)


ibu-imager/seedcreator/seedcreator.go
```diff
@@ -77,11 +79,13 @@ func (s *SeedCreator) CreateSeedImage() error {
 	if err := s.copyConfigurationFiles(); err != nil {
 		return fmt.Errorf("failed to add configuration files: %w", err)
 	}
+	s.successfulStages["ConfigFiles"] = true
```
donpenney, line 82: Please define a set of constants for these, rather than repeating the literals

ibu-imager/cmd/restore.go
```diff
@@ -47,8 +47,15 @@ func restore() {
 	hostCommandsExecutor := ops.NewNsenterExecutor(log, true)
 	op := ops.NewOps(log, hostCommandsExecutor)
 
+	successfulStages = map[string]bool{
```
tsorya, line 50: I don't think we should use this option, we should reuse same logic as we do in creator or remove run.once logic in creator and move to this one. Lets not use 2 different ways for setting if we run function or not

leo8a: the problem I see with using run.once is that right now there are a couple of steps (i.e. recert validations and stopping cluster services) that don't use it... these two are important to track for selective cleanup later on

tsorya: Can we validate that we should run recert or we should start containers? For example currently if there was an issue with stopping containers and etcd is still up cleanup will try to spawn etcd and stuck till timeout, so we could check for example if cluster is up and running before trying to cleanup it? 
I mean having to different options to set Done will be very confusing and will break in a week

leo8a:
depends on #181

ibu-imager/cmd/restore.go
leo8a: Agree on maintaining just one method. Waiting for #181 to land, so we can adjust this to use the `*.done` file approach.

donpenney (commented):
/lgtm

PR #178, opened 2023-12-15
Fix condition observedGeneration update

jc-rh:


omertuc:
lol, good catch

/lgtm

jc-rh:
> lol, good catch
> 
> /lgtm

You had it right. Should have copied yours :)

PR #179, opened 2023-12-15
interfaces and tests for ugHandler

pixelsoccupied:
Exposed interfaces and refactor for UTs


controllers/upgrade_handlers.go
```diff
@@ -169,6 +174,14 @@ func (r *ImageBasedUpgradeReconciler) prePivot(ctx context.Context, ibu *lcav1al
 	return doNotRequeue(), nil
 }
 
+func (r *ImageBasedUpgradeReconciler) getDesiredStaterootRepo(ibu *lcav1alpha1.ImageBasedUpgrade) string {
+	return fmt.Sprintf("/host/ostree/deploy/rhcos_%s/var", ibu.Spec.SeedImageRef.Version)
+}
+
+func (r *ImageBasedUpgradeReconciler) remountSysroot() (string, error) {
```
sudomakeinstall2, line 181: we already have a function for this in ops.

controllers/upgrade_handlers.go
```diff
@@ -169,6 +174,14 @@ func (r *ImageBasedUpgradeReconciler) prePivot(ctx context.Context, ibu *lcav1al
 	return doNotRequeue(), nil
 }
 
+func (r *ImageBasedUpgradeReconciler) getDesiredStaterootRepo(ibu *lcav1alpha1.ImageBasedUpgrade) string {
+	return fmt.Sprintf("/host/ostree/deploy/rhcos_%s/var", ibu.Spec.SeedImageRef.Version)
```
sudomakeinstall2, line 178: I don't think it is important, but we have defined functions like `PathOutsideChroot`, `GetStaterootPath` maybe they can be used here.

Also I don't think this function needs to be in reonciler.

controllers/upgrade_handlers.go
```diff
@@ -169,13 +176,34 @@ func (r *ImageBasedUpgradeReconciler) prePivot(ctx context.Context, ibu *lcav1al
 	return doNotRequeue(), nil
 }
 
+func (r *ImageBasedUpgradeReconciler) ExportIbuCr(ibu *lcav1alpha1.ImageBasedUpgrade, stateRootRepo string) error {
```
tsorya, line 179: ExportIbuCr -> ExportIBUCR ?

controllers/upgrade_handlers.go
pixelsoccupied: Updated

controllers/upgrade_handlers.go
pixelsoccupied: Added a todo to investigate it later and removed func from reonciler

controllers/upgrade_handlers.go
pixelsoccupied: Updated

controllers/upgrade_handlers.go
```diff
@@ -41,20 +42,28 @@ var (
...
+	HandleBackup(r reconcile.Reconciler, ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error)
+	HandleRestore(r reconcile.Reconciler, ctx context.Context) (ctrl.Result, error)
+	RebootToNewStateRoot(r reconcile.Reconciler, ibu *lcav1alpha1.ImageBasedUpgrade) error
+	ExportIBUCR(ibu *lcav1alpha1.ImageBasedUpgrade, filePath string) error
+	PostPivot(r reconcile.Reconciler, ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error)
+	PrePivot(r reconcile.Reconciler, ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error)
+	IsPrePivot(r reconcile.Reconciler, ibu *lcav1alpha1.ImageBasedUpgrade) (bool, error)
+}
+
+type UpgHandler struct{}
```
sudomakeinstall2, line 56: Can you add what UpgHandler needs from reconciler (BackupRestore, log, etc) to this instead?

pixelsoccupied: Yeah i thought of that! But there are things that we need directly from reconciler but are shared by other handlers...e.g executor. So just taking out (e.g) BackRestore and leaving out few others didn't really make sense as this will also open us to ever increasing number of params imo...

sudomakeinstall2: I meant everything it needs. Can't we just initialize it with everything it needs in `main.go` like everything else, you can pass in executor there, too.
```
	r := rr.(*ImageBasedUpgradeReconciler)
	ctrlResult, err := r.UpgradeHandler.HandleBackup(r, ctx, ibu)
```
because this does not look good to me.

controllers/ibu_controller.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/go-logr/logr"
```
tsorya, line 24: not sure why it was moved here but i think it is mistake

controllers/upgrade_handlers.go
```diff
@@ -24,15 +24,16 @@ import (
 	"strings"
 	"time"
 
+	lcautils "github.com/openshift-kni/lifecycle-agent/utils"
```
tsorya, line 27: should be part if next imports block

controllers/upgrade_handlers.go
```diff
@@ -312,3 +309,48 @@ func (r *ImageBasedUpgradeReconciler) handleRestore(ctx context.Context) (ctrl.R
...
+	ibu.ResourceVersion = ""
+	if err := lcautils.MarshalToFile(ibu, filePath); err != nil {
+		return err
+	}
+	ibu.ResourceVersion = rv
+	return nil
+}
+
+// todo: can we use PathOutsideChroot/GetStaterootPath
+func getDesiredStaterootRepo(ibu *lcav1alpha1.ImageBasedUpgrade) string {
```
tsorya, line 354: why not GetStaterootPath ?

controllers/ibu_controller.go
pixelsoccupied: this is auto sorted (L34 moved L24)

controllers/upgrade_handlers.go
pixelsoccupied: What do you mean? btw it was auto sorted and moved at the top. (L35 now L27)

controllers/upgrade_handlers.go
```diff
@@ -312,3 +309,48 @@ func (r *ImageBasedUpgradeReconciler) handleRestore(ctx context.Context) (ctrl.R
...
+	r.Log.Info(fmt.Sprintf("Wait for %s to be killed via SIGTERM", defaultRebootTimeout.String()))
+	time.Sleep(defaultRebootTimeout)
+
+	return fmt.Errorf("failed to reboot. This should never happen! Please check the system")
+}
+
+func (UpgHandler) ExportIBUCR(ibu *lcav1alpha1.ImageBasedUpgrade, stateRootRepo string) error {
+	filePath := filepath.Join(stateRootRepo, utils.IBUFilePath)
+	// Temporarily empty resource version so the file can be used to restore status
+	rv := ibu.ResourceVersion
```
donpenney, line 344: https://github.com/openshift-kni/lifecycle-agent/pull/189 revises some of these, including:
* moving the ResourceVersion to the restore side to avoid clearing it everywhere it's saved
* renaming isPrePivot to isOrigStaterootBooted, for clarity due to rollback usage

controllers/upgrade_handlers.go
donpenney: This is also revised in https://github.com/openshift-kni/lifecycle-agent/pull/189

controllers/ibu_controller.go
donpenney: The auto-sort would be done within groups of imports. Looks like this was moved into the group of LCA modules from a group of public modules, then resorted

donpenney: The modules I've written, I've separated into groups of:
* core golang modules
* LCA modules
* external utils/modules
* APIs

controllers/upgrade_handlers.go
pixelsoccupied: Like we spoke about....will get rid mockgen in upgradehandler and refactor from the top level . That should get rid of the ugly part but will lose some of the mockgen functionality. Updating the code now

PR #188, opened 2023-12-19
Seed image format version OCI image label

omertuc:
IBU reconciler now has a `checkSeedImageCompatibility` method that
checks if the seed image is compatible with the current version of the
lifecycle-agent. It does so by inspecting the OCI image's labels and
checking if the specified format version equals the hard-coded one that
this version of the lifecycle agent expects. That format version is set
by the imager during the image build process, and is only manually
bumped by developers when the image format changes in a way that is
incompatible with previous versions of the lifecycle-agent.

### DIFF
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -68,8 +69,62 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
 		return fmt.Errorf("failed to pull image: %w", err)
 	}
 
-	return nil
+	r.Log.Info("Checking seed image compatibility")
+	if err := r.checkSeedImageCompatibility(ctx, ibu.Spec.SeedImageRef.Image); err != nil {
+		return fmt.Errorf("checking seed image compatibility: %w", err)
+	}
 
+	return nil
+}
+
+// checkSeedImageCompatibility checks if the seed image is compatible with the
+// current version of the lifecycle-agent by inspecting the OCI image's labels
+// and checking if the specified format version equals the hard-coded one that
+// this version of the lifecycle agent expects. That format version is set by
+// the imager during the image build process, and is only manually bumped by
+// developers when the image format changes in a way that is incompatible with
+// previous versions of the lifecycle-agent.
+func (r *ImageBasedUpgradeReconciler) checkSeedImageCompatibility(_ context.Context, seedImageRef string) error {
+	inspectArgs := []string{
+		"inspect",
+		"--format", "json",
+		seedImageRef,
+	}
+
+	var inspect []struct {
+		Labels map[string]string `json:"Labels"`
+	}
+
+	// TODO: use the context when execute supports it
+	if inspectRaw, err := r.Executor.Execute("podman", inspectArgs...); err != nil || inspectRaw == "" {
+		return fmt.Errorf("failed to inspect image: %w", err)
+	} else {
+		if err := json.Unmarshal([]byte(inspectRaw), &inspect); err != nil {
+			return fmt.Errorf("failed to unmarshal image inspect output: %w", err)
+		}
+	}
+
+	if len(inspect) != 1 {
+		return fmt.Errorf("expected 1 image inspect result, got %d", len(inspect))
+	}
+
+	seedFormatLabelValue, ok := inspect[0].Labels[common.SeedFormatOCILabel]
+	if !ok {
+		return fmt.Errorf(
+			"seed image %s is missing the %s label, please build a new image using the latest version of the imager",
+			seedImageRef, common.SeedFormatOCILabel)
+	}
+
+	// Hard equal since we don't have backwards compatibility guarantees yet.
+	// In the future we might want to have backwards compatibility code to
+	// handle older seed formats and in that case we'll look at the version
+	// number and do the right thing.
+	if seedFormatLabelValue != fmt.Sprintf("%d", common.SeedFormatVersion) {
+		return fmt.Errorf("seed image format version mismatch: expected %d, got %s",
+			common.SeedFormatVersion, seedFormatLabelValue)
+	}
+
+	return nil
 }
 
 func readPrecachingList(imageListFile, clusterRegistry, seedRegistry string, overrideSeedRegistry bool) (imageList []string, err error) {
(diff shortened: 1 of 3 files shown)


controllers/prep_handlers.go
```diff
@@ -68,8 +69,56 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
...
+		"inspect",
+		"--format", "json",
+		"--authfile", pullSecretFilename,
+		seedImageRef,
+	}
+
+	var inspect struct {
+		Labels map[string]string `json:"Labels"`
+	}
+	if inspectRaw, err := r.Executor.Execute("podman", inspectArgs...); err != nil || inspectRaw == "" {
```
donpenney, line 98: One advantage to your method of adding the compatibility version in a label means we could potentially inspect the image manifest before pulling it, such as with ```skopeo inspect```. Given that the images are more than 1G, that could be very beneficial.

internal/common/consts.go
```diff
@@ -57,6 +57,10 @@ const (
 	InstallConfigCM = "cluster-config-v1"
 	// InstallConfigCMNamespace cm namespace
 	InstallConfigCMNamespace = "kube-system"
+
+	// Bump this every time the seed format changes in a backwards incompatible way
+	SeedFormatVersion = 1
```
donpenney, line 62: Since "1" is being used as the default if the version isn't present, maybe this should start at 2?

donpenney (commented):
This looks like a great first step for a compatibility check, something we can build upon as we figure out what we'll need. And addresses a need we've seen multiple times.

controllers/prep_handlers.go
omertuc: Yep. Will leave that for a future PR

controllers/prep_handlers.go
```diff
@@ -68,8 +69,56 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
...
+	seedFormatLabelValue, ok := inspect.Labels[common.SeedFormatOCILabel]
+	if !ok {
+		// Pretend it's the first version
+		seedFormatLabelValue = "1"
+	}
+
+	// Hard equal since we don't have backwards compatibility guarantees yet.
+	// In the future we might want to have backwards compatibility code to
+	// handle older seed formats and in that case we'll look at the version
+	// number and do the right thing.
```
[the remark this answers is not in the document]
omertuc: yes, but for now we're moving fast and breaking things, so we shouldn't worry about compatibility - just need to be more explicit that we're breaking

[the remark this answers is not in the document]
donpenney: Agreed. I think this is good for our immediate needs

internal/common/consts.go
omertuc: Since seeds created before this PR are perfectly valid, I chose the default to be the same as this so I won't unnecessarily block all latest seeds

donpenney: Given that we had such a case yesterday, though, which was one of the drivers for adding this, it's really just "seeds created with the last couple of PRs" that would be valid.

omertuc: But I can just add the label manually

controllers/prep_handlers.go
```diff
@@ -68,8 +69,57 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
...
+		"--authfile", pullSecretFilename,
+		seedImageRef,
+	}
+
+	var inspect struct {
+		Labels map[string]string `json:"Labels"`
+	}
+	if inspectRaw, err := r.Executor.Execute("podman", inspectArgs...); err != nil || inspectRaw == "" {
+		return fmt.Errorf("failed to inspect image: %w", err)
+	} else {
```
sudomakeinstall2, line 100: else seems unnecessary

omertuc: inspectRaw scope is lost otherwise

controllers/prep_handlers.go
```diff
@@ -68,8 +69,57 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
...
+}
+
+// checkSeedImageCompatibility checks if the seed image is compatible with the
+// current version of the lifecycle-agent by inspecting the OCI image's labels
+// and checking if the specified format version equals the hard-coded one that
+// this version of the lifecycle agent expects. That format version is set by
+// the imager during the image build process, and is only manually bumped by
+// developers when the image format changes in a way that is incompatible with
+// previous versions of the lifecycle-agent.
+func (r *ImageBasedUpgradeReconciler) checkSeedImageCompatibility(ctx context.Context, pullSecretFilename string, seedImageRef string) error {
```
donpenney, line 87: Linter is failing:
```suggestion
//
//nolint:unparam
func (r *ImageBasedUpgradeReconciler) checkSeedImageCompatibility(ctx context.Context, pullSecretFilename, seedImageRef string) error {
```

internal/common/consts.go
donpenney, line 62: Spacing change for go fmt:
```suggestion
	SeedFormatVersion  = 1
```

controllers/prep_handlers.go
omertuc: Created #200

controllers/prep_handlers.go
omertuc: Created #201

controllers/prep_handlers.go
```diff
@@ -68,8 +69,59 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
...
+// and checking if the specified format version equals the hard-coded one that
+// this version of the lifecycle agent expects. That format version is set by
+// the imager during the image build process, and is only manually bumped by
+// developers when the image format changes in a way that is incompatible with
+// previous versions of the lifecycle-agent.
+func (r *ImageBasedUpgradeReconciler) checkSeedImageCompatibility(_ context.Context, pullSecretFilename, seedImageRef string) error {
+	inspectArgs := []string{
+		"inspect",
+		"--format", "json",
+		"--authfile", pullSecretFilename,
```
donpenney, line 91: The podman inspect command does not support `--authfile` option.

controllers/prep_handlers.go
```diff
@@ -68,8 +69,62 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
 		return fmt.Errorf("failed to pull image: %w", err)
 	}
 
+	r.Log.Info("Checking seed image compatibility")
+	if err := r.checkSeedImageCompatibility(ctx, pullSecretFilename); err != nil {
```
donpenney, line 73:
```suggestion
	if err := r.checkSeedImageCompatibility(ctx, ibu.Spec.SeedImageRef.Image); err != nil {
```

omertuc:
Tested

/unhold

PR #189, opened 2023-12-19
Add support for rollback and ostree admin set-default

donpenney:
Added use of "ostree admin set-default" feature, if enabled, to manage setting the desired next-boot deployment.

This feature also allows for the introduction of rollback support.

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
@@ -417,5 +418,5 @@ func (r *ImageBasedUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func getDesiredStaterootName(ibu *lcav1alpha1.ImageBasedUpgrade) string {
-	return fmt.Sprintf("rhcos_%s", ibu.Spec.SeedImageRef.Version)
+	return fmt.Sprintf("rhcos_%s", strings.ReplaceAll(ibu.Spec.SeedImageRef.Version, "-", "_"))
 }
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -241,6 +241,10 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 
 func isRollbackAllowed(ibu *lcav1alpha1.ImageBasedUpgrade, isAfterPivot bool) bool {
 	if !isAfterPivot {
+		rollbackInProgressCondition := meta.FindStatusCondition(ibu.Status.Conditions, string(utils.ConditionTypes.RollbackInProgress))
+		if rollbackInProgressCondition != nil && rollbackInProgressCondition.Status == metav1.ConditionTrue {
+			return true
+		}
 		return false
 	}
 	upgradeInProgressCondition := meta.FindStatusCondition(ibu.Status.Conditions, string(utils.ConditionTypes.UpgradeInProgress))
--- a/controllers/rollback_handlers.go
+++ b/controllers/rollback_handlers.go
@@ -18,18 +18,90 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
+	"github.com/openshift-kni/lifecycle-agent/internal/common"
+	"github.com/openshift-kni/lifecycle-agent/internal/ostreeclient"
+	lcautils "github.com/openshift-kni/lifecycle-agent/utils"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	lcav1alpha1 "github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
-	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 //nolint:unparam
-func (r *ImageBasedUpgradeReconciler) handleRollback(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+func (r *ImageBasedUpgradeReconciler) startRollback(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	utils.SetRollbackStatusInProgress(ibu, "Initiating rollback")
 
-	// TODO actual steps
-	// If completed, update conditions and return doNotRequeue
+	stateroot, err := r.RPMOstreeClient.GetUnbootedStaterootName()
+	if err != nil {
+		utils.SetRollbackStatusFailed(ibu, err.Error())
+		return requeueWithError(err)
+	}
+
+	if _, err := r.Executor.Execute("mount", "/sysroot", "-o", "remount,rw"); err != nil {
+		return requeueWithError(err)
+	}
+
+	stateRootRepo := common.PathOutsideChroot(ostreeclient.StaterootPath(stateroot))
+
+	// Save the CR for post-reboot restore, stripping Upgrade conditions and ResourceVersion from IBU CR before saving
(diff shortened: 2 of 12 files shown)


controllers/rollback_handlers.go
```diff
@@ -42,5 +114,24 @@ func (r *ImageBasedUpgradeReconciler) handleRollback(ctx context.Context, ibu *l
...
+	if err != nil {
+		//todo: abort handler? e.g delete desired stateroot
+		utils.SetRollbackStatusFailed(ibu, err.Error())
+		return doNotRequeue(), nil
+	}
+
+	if origStaterootBooted {
+		r.Log.Info("Pivot for rollback successful, starting post pivot steps")
+		return r.finishRollback(ctx, ibu)
+	} else {
```
hexfusion, line 133: nit: since the above if statement returns the else is not necessary

controllers/upgrade_handlers.go
```diff
@@ -157,6 +160,25 @@ func (r *ImageBasedUpgradeReconciler) prePivot(ctx context.Context, ibu *lcav1al
 	}
 	ibu.ResourceVersion = rv
 
+	// Save a copy of the IBU in the current stateroot in case of uncontrolled rollback
+	ibuCopy := ibu.DeepCopy()
+	ibuCopy.ResourceVersion = ""
```
hexfusion, line 165: Should we add a comment on why we reset the ResourceVersion here?

ibu-imager/ostreeclient/rpmostreeclient.go
```diff
@@ -98,6 +101,41 @@ type rpmOstreeVersionData struct {
...
+	for _, deployment := range deployments {
+		if deployment.OSName != bootedStateroot {
+			return deployment.OSName
+		}
+	}
+
+	return ""
+}
+
+func getDeploymentIndex(deployments []Deployment, osname string) int {
```
hexfusion, line 130:
```suggestion
func getDeploymentIndex(deployments []Deployment, osname string) (int, error) {
	for index, deployment := range deployments {
		if deployment.OSName == osname {
			return index, nil
		}
	}
	return 0, fmt.Errorf("failed to find deployment with osname %s", osname)
}
```

controllers/upgrade_handlers.go
tsorya: We are doing twice same operation, lets have function for it? Like writeIBUToFile ? in which we will copy it and write without resource version ?

controllers/rollback_handlers.go
```diff
@@ -18,18 +18,90 @@ package controllers
...
+	deploymentIndex, err := r.RPMOstreeClient.GetUnbootedDeploymentIndex()
+	if err != nil {
+		utils.SetRollbackStatusFailed(ibu, err.Error())
+		return requeueWithError(fmt.Errorf("failed to get unbooted deployment index: %w", err))
+	}
+
+	// Set the new default deployment
+	r.Log.Info("Checking for set-default feature")
+
+	if r.OstreeClient.IsOstreeAdminSetDefaultFeatureEnabled() {
```
tsorya, line 72: Logic seems very similar between both functions, can we somehow unite it?

ibu-imager/ostreeclient/rpmostreeclient.go
```diff
@@ -98,6 +101,41 @@ type rpmOstreeVersionData struct {
 	Root VersionData `yaml:"rpm-ostree"`
 }
 
+func getBootedStaterootName(deployments []Deployment) string {
+	for _, deployment := range deployments {
+		if deployment.Booted {
+			return deployment.OSName
+		}
+	}
+	return ""
```
tsorya, line 110: This is not error?

ibu-imager/ostreeclient/rpmostreeclient.go
```diff
@@ -98,6 +101,41 @@ type rpmOstreeVersionData struct {
...
+}
+
+func getUnbootedStaterootName(deployments []Deployment) string {
+	// Determine the booted stateroot
+	bootedStateroot := getBootedStaterootName(deployments)
+	if bootedStateroot == "" {
+		return ""
```
tsorya, line 117: Why not to return error here?

internal/ostreeclient/ostreeclient.go
```diff
@@ -46,3 +52,25 @@ func (c *Client) Undeploy(ostreeIndex int) error {
...
+		// Already set as default deployment
+		return nil
+	}
+
+	args := []string{"admin", "set-default", strconv.Itoa(index)}
+	_, err := c.executor.Execute("ostree", args...)
+	return err
+}
+
+func StaterootPath(stateroot string) string {
```
tsorya, line 74: use GetStaterootPath from utils please

controllers/rollback_handlers.go
donpenney: Do you mean the if/else clauses? These are very different paths. If set-default is available, the rollback handler will use it to set the new default deployment automatically. If it is not, it checks to see if the user has manually set it, and fails the rollback otherwise, with a message reminding the user to manually set it. And at some point, the "set-default not available" code path will be removed, once we define a minimum-required release.

controllers/rollback_handlers.go
```diff
@@ -18,18 +18,90 @@ package controllers
...
 
 //nolint:unparam
-func (r *ImageBasedUpgradeReconciler) handleRollback(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+func (r *ImageBasedUpgradeReconciler) startRollback(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
+	utils.SetRollbackStatusInProgress(ibu, "Initiating rollback")
+
+	stateroot, err := r.RPMOstreeClient.GetUnbootedStaterootName()
+	if err != nil {
+		utils.SetRollbackStatusFailed(ibu, err.Error())
+		return requeueWithError(err)
```
Missxiaoguo, line 42: Looks like we want to fail the rollback in this case as you have set the rollback condition to fail, then we should return "doNotRqueue(), nil", otherwise it will keep retrying.
Same comment for other places have called "utils.SetRollbackStatusFailed(ibu, err.Error())"

controllers/rollback_handlers.go
```diff
@@ -18,18 +18,90 @@ package controllers
...
+		utils.SetUpgradeStatusFailed(ibu, err.Error())
+		return doNotRequeue(), nil
+	}
+
+	return doNotRequeue(), nil
+}
+
+//nolint:unparam
+func (r *ImageBasedUpgradeReconciler) finishRollback(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
 	utils.SetStatusCondition(&ibu.Status.Conditions,
```
Missxiaoguo, line 105: nit: You have SetRollbackStatusCompleted() defined, but looks like you are not using it.

controllers/upgrade_handlers.go
donpenney: Moving the ResourceVersion reset for IBU and SeedGen to the restore side, so it's in one spot, which simplifies the saving

main/main.go
```diff
@@ -223,6 +223,9 @@ func initIBU(ctx context.Context, c client.Client, log *logr.Logger) error {
 		return err
 	}
 
+	// Strip the ResourceVersion, otherwise the restore fails
+	ibu.SetResourceVersion("")
```
jc-rh, line 227: This is actually not needed. I forgot to clean it up from the saving function. It was needed when I had the restore in a script where the curl commands for post and patch use the same source file. With api calls, we get the right revision in the structure after the create succeeds.

donpenney: I'm pretty sure I hit the same issues with the seedgen CR. I'd rather leave it here, and we can remove it in a separate PR if it's truly not needed.

controllers/ibu_controller.go
```diff
@@ -241,6 +242,10 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 
 func isRollbackAllowed(ibu *lcav1alpha1.ImageBasedUpgrade, isAfterPivot bool) bool {
 	if !isAfterPivot {
+		rollbackInProgressCondition := meta.FindStatusCondition(ibu.Status.Conditions, string(utils.ConditionTypes.RollbackInProgress))
+		if rollbackInProgressCondition != nil && rollbackInProgressCondition.Status == metav1.ConditionTrue {
+			return true
```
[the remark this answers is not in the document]
jc-rh: Feel free to leave the cleanup for later.

PR #190, opened 2023-12-20
health check unit tests

pixelsoccupied:
/cc @donpenney 
/cc @Missxiaoguo 
/cc @jc-rh

### DIFF
--- a/internal/healthcheck/healthcheck_test.go
+++ b/internal/healthcheck/healthcheck_test.go
@@ -1,0 +1,655 @@
+package healthcheck
+
+import (
+	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
+	mcv1 "github.com/openshift/api/machineconfiguration/v1"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


internal/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,687 @@
...
+			s.AddKnownTypes(configv1.GroupVersion, &configv1.ClusterVersion{})
+			s.AddKnownTypes(configv1.GroupVersion, &configv1.ClusterVersionList{})
+			s.AddKnownTypes(mcv1.GroupVersion, &mcv1.MachineConfigPoolList{})
+			s.AddKnownTypes(mcv1.GroupVersion, &mcv1.MachineConfigPool{})
+			s.AddKnownTypes(configv1.GroupVersion, &configv1.ClusterOperatorList{})
+			s.AddKnownTypes(configv1.GroupVersion, &configv1.ClusterOperator{})
+			s.AddKnownTypes(operatorsv1alpha1.SchemeGroupVersion, &operatorsv1alpha1.ClusterServiceVersionList{})
+			s.AddKnownTypes(operatorsv1alpha1.SchemeGroupVersion, &operatorsv1alpha1.ClusterServiceVersion{})
+			s.AddKnownTypes(configv1.GroupVersion, &configv1.Infrastructure{})
+			s.AddKnownTypes(v1.SchemeGroupVersion, &v1.Node{})
```
Missxiaoguo, line 677: nit: I think these do not have to be called every time a testcase is executed. Perhaps move them out of the for loop or define them only once in an init function?

pixelsoccupied: Yep! Updated to use init()

jc-rh (approved):

PR #191, opened 2023-12-20
Add printer columns for ibu

sudomakeinstall2:
- Add State column that shows the type of last condition whose status has been changed
- Add Details column that shows the message of last condition whose status has been changed

SetStatusCondition tries to keep the conditions sorted by the time their status was last changed


donpenney (commented):
/lgtm

PR #194, opened 2023-12-21
Recreate IBU when deleted

jc-rh:


controllers/ibu_controller.go
```diff
@@ -411,7 +412,18 @@ func (r *ImageBasedUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			},
 			CreateFunc:  func(ce event.CreateEvent) bool { return true },
 			GenericFunc: func(ge event.GenericEvent) bool { return false },
-			DeleteFunc:  func(de event.DeleteEvent) bool { return false },
+			DeleteFunc: func(de event.DeleteEvent) bool {
+				if de.Object.GetName() == utils.IBUName {
+					filePath := common.PathOutsideChroot(utils.IBUFilePath)
+					if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
+						return lcautils.MarshalToFile(de.Object, filePath)
```
pixelsoccupied, line 419: curious if you tried using finalizer or maybe that wouldnt work? I'm thinking if deletionTimestamp is set...just keep a copy...remove the finalizer and the apply the copy?

jc-rh: Not sure if you can recreate it right away in the finalizer. Delete event happens after it's actually removed from etcd.

PR #195, opened 2023-12-21
Mirror registry : Adding verification if we should override mirror registry or not

tsorya:
1. If seed has mirror registry and cluster not, try to override
2. If cluster has mirror registry and seed not verify if seeds registry is part of sources in icsps and idms and if it is don't override
3. If cluster has mirror registry and seed regsitry is not part of sources try to override seeds registry

### DIFF
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -72,7 +72,7 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
 
 }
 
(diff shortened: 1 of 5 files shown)


donpenney:
This resolves an issue in a scenario I was trying, using "connected" registries for seed and target SNO: Running an IBU from a stable release to a nightly build, which have different registries for the release image. Tested this PR in my setup, and it works as expected. I did not test with mirrors.

donpenney (approved):

PR #199, opened 2023-12-31
MGMT-16479: Adding new command to lca that will run IBI installation flow

tsorya:
IBI FLOW: Adding new command to lca that will run IBI installation flow
Run state root preparation flow on iso
Precache images

### DIFF
--- a/ibu-imager/ops/execute.go
+++ b/ibu-imager/ops/execute.go
@@ -16,6 +16,7 @@ import (
 //go:generate mockgen -source=execute.go -package=ops -destination=mock_execute.go
 type Execute interface {
 	Execute(command string, args ...string) (string, error)
+	ExecuteWithLiveLogger(command string, args ...string) (string, error)
 }
 
 type executor struct {
@@ -59,6 +60,10 @@ func (e *regularExecutor) Execute(command string, args ...string) (string, error
 	return e.executor.execute(nil, "", command, args...)
 }
 
+func (e *regularExecutor) ExecuteWithLiveLogger(command string, args ...string) (string, error) {
+	return e.executor.execute(e.executor.log.Writer(), "", command, args...)
+}
+
 type nsenterExecutor struct {
 	executor
 }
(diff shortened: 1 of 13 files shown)


ibu-imager/ops/execute.go
```diff
@@ -67,7 +72,11 @@ func NewNsenterExecutor(logger *logrus.Logger, verbose bool) Execute {
 	return &nsenterExecutor{executor: executor{logger, verbose}}
 }
 
-func (e *nsenterExecutor) Execute(command string, args ...string) (string, error) {
+func (e *nsenterExecutor) ExecuteWithLiveLogger(command string, args ...string) (string, error) {
```
eranco74, line 75: Something is off here.
you have `ExecuteWithLiveLogger` twice.
Also, why wrap `Execute` with `baseExecute`?

ibu-imager/ops/ops.go
```diff
@@ -208,7 +208,7 @@ func (o *ops) ExtractTarWithSELinux(srcPath, destPath string) error {
 }
 
 func (o *ops) RemountSysroot() error {
-	_, err := o.hostCommandsExecutor.Execute("mount", "/sysroot", "-o", "remount,rw")
+	_, err := o.hostCommandsExecutor.Execute("mount", "/mnt/sysroot", "-o", "remount,rw")
```
eranco74, line 211: Won't this break IBU?

internal/ostreeclient/ostreeclient.go
```diff
@@ -15,41 +16,60 @@ type IClient interface {
...
 
 func (c *Client) Deploy(osname, refsepc string, kargs []string) error {
 	args := []string{"admin", "deploy", "--os", osname}
+	if c.ibi {
+		args = append(args, "--sysroot", "/mnt")
+	}
 	args = append(args, kargs...)
 	args = append(args, refsepc)
-	if c.IsOstreeAdminSetDefaultFeatureEnabled() {
+	if c.IsOstreeAdminSetDefaultFeatureEnabled() && !c.ibi {
```
eranco74, line 60: Why not just set it as default in case of IBI?

ibu-imager/ops/ops.go
tsorya: mistake

internal/ostreeclient/ostreeclient.go
tsorya: I will change places between them but i don't fully understand the comment

ibu-imager/ops/execute.go
tsorya: cause this is interface with 2 functions? and here we have 3 different options of this interface. chroot, nsenter and os.cmd

main/precache-workload/main.go
```diff
@@ -91,20 +89,18 @@ func main() {
 
 	log.Info("Loaded precache spec file.")
 
-	// Change root directory to /host
-	if err := syscall.Chroot(common.Host); err != nil {
-		terminateOnError(fmt.Errorf("failed to chroot to %s, err: %w", common.Host, err), Failure)
-	}
-	log.Infof("chroot %s successful", common.Host)
-
```
sakhoury, line 99: The pre-caching k8s job needs to chroot to the mounted host filesystem. May I ask why you removed this?

tsorya: I changed it to be the same as all other code. Now it uses chroot executor

PR #202, opened 2024-01-03
Fixing mirror registry in case seed registry is part of mirror configuration sources

tsorya:


donpenney:
Now that we have a compatibility version identifier with https://github.com/openshift-kni/lifecycle-agent/pull/188, does this PR need to increment the SeedFormatVersion value in internal/common/consts.go?

donpenney (approved):

PR #206, opened 2024-01-04
Dummy change to test #204

omertuc:
Ignore


omertuc:
Could it be that [non-openshift org labels](https://github.com/openshift/release/blob/1659b748737627b61939fc9650d8c33da40134d7/core-services/prow/02_config/_labels.yaml#L1016-L1022) are ignored?

donpenney:
> Could it be that [non-openshift org labels](https://github.com/openshift/release/blob/1659b748737627b61939fc9650d8c33da40134d7/core-services/prow/02_config/_labels.yaml#L1016-L1022) are ignored?
> 
> EDIT: nah there's also some `operator-framework` org labels in there, nevermind

I think it needs to be included in the additional_labels list here, maybe:
https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L652

PR #207, opened 2024-01-04
precache: Include env vars from manager pod when creating precache job

donpenney:


controllers/prep_handlers.go
```diff
@@ -153,6 +154,29 @@ func readPrecachingList(imageListFile, clusterRegistry, seedRegistry string, ove
 	return imageList, nil
 }
 
+func (r *ImageBasedUpgradeReconciler) getPodEnvVars(ctx context.Context) (envVars []corev1.EnvVar, err error) {
```
sakhoury, line 157: This function deserves to be with other util-based functions.

internal/precache/precache.go
```diff
@@ -44,12 +46,17 @@ type PHandler struct {
...
 	// To run pre-caching job with an adjusted niceness, which affects process scheduling.
 	// Niceness values range from -20 (most favorable to the process) to 19 (least favorable to the process).
 	NicePriority int
+
 	// To configure the I/O-scheduling class and priority of a process.
 	IoNiceClass    int // 0: none, 1: realtime, 2: best-effort, 3: idle
 	IoNicePriority int // priority (0..7) in the specified scheduling class, only for the realtime and best-effort classes
+
+	// Allow for environment variables to be passed in
+	EnvVars []corev1.EnvVar
```
sakhoury, line 59: Can you also update the precache doc with the env variable?

controllers/prep_handlers.go
donpenney: Does it? It's a very limited use function. Why would it need to be in a common utils module, vs just an internal-use function?

internal/precache/precache.go
donpenney: Updated

controllers/prep_handlers.go
sakhoury: Retrieving the environment variables from the pod sounds like a useful function that may benefit other use-cases in the future, we may consider housing it in a shared internal location such as `internal/common/utils.go`. Alternatively, if you strongly feel that it has very limited usability, specifically for precaching only, then why not move this function to precache/helper.go directly?

donpenney: Originally, I was coding this to just copy vars directly from the os.Environ, and had the code in precache/helper.go, along with handling for the unit tests. However, this led to a need to filter a number of kubernetes-added and build-added vars, so ultimately decided grabbing from the manager spec was best. Grabbing this from the reconciler seemed like the best place to do it, where the job is getting created from the manager pod, and also keeps that "get my pod spec" out of the precache unit test path. If anything, this could have just been inline in the launcher, between the NewConfig and CreateJob calls, rather than a separate function

sakhoury: > If anything, this could have just been inline in the launcher, between the NewConfig and CreateJob calls, rather than a separate function

This would be better, and in the future, if other use-cases require extracting pod env-vars, we can create the function separately in utils.

donpenney: That's why it's just an internal function, though, as I wouldn't expect it to be needed for anything else

jc-rh (approved):

PR #208, opened 2024-01-05
`SEEDGEN_IMAGER_SKIP_RECERT` to skip recert validation

omertuc:
When the imager is launched through the `SeedGenerator` CR, we want the
to have the option to skip the recert stage of the seed image generation.

To that end this commit adds a new environment variable
`SEEDGEN_IMAGER_SKIP_RECERT` which when set to the value `TRUE`, will
make the `SeedGeneratorReconciler` ask the imager to skip recert
validation

This is useful when recert fails but we still want an image to be
created so that it can be shared for debugging and troubleshooting
purposes

### DIFF
--- a/controllers/seedgen_controller.go
+++ b/controllers/seedgen_controller.go
@@ -73,6 +73,10 @@ var (
 	imagerContainerName    = "ibu_imager"
 )
 
+const (
+	EnvSkipRecert = "SEEDGEN_IMAGER_SKIP_RECERT"
+)
+
 //+kubebuilder:rbac:groups=lca.openshift.io,resources=seedgenerators,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=lca.openshift.io,resources=seedgenerators/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=lca.openshift.io,resources=seedgenerators/finalizers,verbs=update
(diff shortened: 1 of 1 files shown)


donpenney:
I don't think this should be a field in the CRD. Like with the precaching "best-effort" option, I'd go with an env var

donpenney:
> I don't think this should be a field in the CRD. Like with the precaching "best-effort" option, I'd go with an env var

Maybe an env var like SEEDGEN_IMAGER_ADDITIONAL_ARGS, that the user could set whatever args they want? ie.
```
SEEDGEN_IMAGER_ADDITIONAL_ARGS="--skip-recert-validation --skip-cleanup"
```

omertuc:
> > I don't think this should be a field in the CRD. Like with the precaching "best-effort" option, I'd go with an env var
> 
> Maybe an env var like SEEDGEN_IMAGER_ADDITIONAL_ARGS, that the user could set whatever args they want? ie.
> 
> ```
> SEEDGEN_IMAGER_ADDITIONAL_ARGS="--skip-recert-validation --skip-cleanup"
> ```

I wouldn't like to get into the business of separating args, it's non-trivial (you can't just split by space). I prefer dedicated environment variables.

Ideally I think the SeedGeneratorReconciler <-> ibu imager communication should also be done with a config file and or environment variables, instead of CLI which is made mostly for humans, so I wouldn't want to leak that CLI to the CR users in case we'll want to do remove that CLI in the future

donpenney (approved):

PR #210, opened 2024-01-09
Improve Rollback and Seedgen status reporting and error handling

donpenney:
Updates include:
- Include Stage in IBU display
- Include State and Details in SeedGen display
- Make seedgen spec immutable and only handle Create events
- Add mutex to prevent IBU and SeedGen reconcilers from running at the same time
- Delete IBU CR prior to generating seed image, so it is not included in the seed
- Update Rollback states and progress messages to provide more useful feedback for user
- Improve seedgen error handling and reporting

/cc @browsell @jc-rh


donpenney:
/cc @pixelsoccupied 
Updated to rebase against https://github.com/openshift-kni/lifecycle-agent/pull/179

PR #211, opened 2024-01-10
workload paritioning annotation

pixelsoccupied:


config/manager/kustomization.yaml
```diff
@@ -16,3 +16,15 @@ images:
   newTag: 4.14.0
 patches:
 - path: related-images/patch.yaml
+- patch: |
```
browsell, line 19: why not just update the deployment spec directly, https://github.com/openshift-kni/lifecycle-agent/blob/main/config/manager/manager.yaml#L13

pixelsoccupied: Updated to include the annotation directly

browsell:
These pods will fail to be admitted without the associated namespace annotation

pixelsoccupied:
Added `workload.openshift.io/allowed: management` in ns

PR #213, opened 2024-01-10
restrict permissions on files generated by LCA for IBU

pixelsoccupied:
/cc @donpenney

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -31,22 +31,23 @@ import (
 	k8syaml "sigs.k8s.io/yaml"
 )
 
-// MarshalToFile marshals anything and writes it to the given file path
+// MarshalToFile marshals anything and writes it to the given file path. file only readable by root
 func MarshalToFile(data any, filePath string) error {
 	marshaled, err := json.Marshal(data)
 	if err != nil {
 		return err
 	}
-	return os.WriteFile(filePath, marshaled, 0o644)
+	return os.WriteFile(filePath, marshaled, 0o600)
 }
 
 // MarshalToYamlFile marshals any object to YAML and writes it to the given file path
+// file only readable by root
 func MarshalToYamlFile(data any, filePath string) error {
 	marshaled, err := k8syaml.Marshal(data)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


utils/utils.go
```diff
@@ -37,16 +37,27 @@ func MarshalToFile(data any, filePath string) error {
...
+func MarshalToFileReadOnlyByRoot(data any, filePath string) error {
+	marshaled, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(filePath, marshaled, common.FilePerm600)
 }
 
 // MarshalToYamlFile marshals any object to YAML and writes it to the given file path
+// file only readable by root
```
browsell, line 54: same as above function ?

donpenney:
I think all we need to do here is make sure the directories we create are 0700 perms and the files are 0600. ie. Just change the permissions for the existing calls

utils/utils.go
pixelsoccupied: Yeah i thought we would needed to split the write func depending if it's "non-ibu" component or not. But change is applicable globally. Removed it now

internal/common/consts.go
```diff
@@ -61,6 +63,10 @@ const (
 	// Bump this every time the seed format changes in a backwards incompatible way
 	SeedFormatVersion  = 2
 	SeedFormatOCILabel = "com.openshift.lifecycle-agent.seed_format_version"
+
+	// common file permissions
+	FilePerm600 os.FileMode = 0o600 // For secret files.
```
donpenney, line 68: If you're going to create constants for the perms, I would suggest something like RootOnlyRW and RootOnlyRWX. There doesn't seem to be much value in having a constant with the value in the name

donpenney: Additionally, if you're creating constants, use them for all instances. There are multiple places where we already use `0o700` for directories and `0o600` for files. Don't just update the permissions you're changing.

pixelsoccupied: Yep makes sense. The intention was have minimal changes so i removed the const to keep it consistent

donpenney (approved):

PR #217, opened 2024-01-12
Simplify Dockerfile

donpenney:


rauhersu:
Hey @donpenney thanks for including me as a reviewer: these changes affecting Dockerfiles do impact midstream build pipelines. As I have seen, changes should be pretty straightforward to be mapped to midstream, but can we put them on hold until the next week? The reason is that I am in the middle of some extra changes and I would like keep this upstream-midstream build stage in sync until I get back to this project.

Note: these changes won't break midstream in case you merge them, but they will be ignored; the builds will differ as we will be based on  the old Dockerfile.

donpenney:
> Hey @donpenney thanks for including me as a reviewer: these changes affecting Dockerfiles do impact midstream build pipelines. As I have seen, changes should be pretty straightforward to be mapped to midstream, but can we put them on hold until the next week? The reason is that I am in the middle of some extra changes and I would like keep this upstream-midstream build stage in sync until I get back to this project.
> 
> Note: these changes won't break midstream in case you merge them, but they will be ignored; the builds will differ as we will be based on the old Dockerfile.

Of course! I thought this was mostly in line with what you've got, but this change isn't urgent. And also needs coordination with the release repo, in order to change the base load

donpenney:
Also, this could be split into two PRs:

- Simplify the dockerfile to a single compile layer
- Move from ubi to ubi-minimal

donpenney:
I've split out the move to ubi-minimal to a separate PR: https://github.com/openshift-kni/lifecycle-agent/pull/218

jc-rh (approved):

PR #220, opened 2024-01-15
Filtering known catalog source images from container list

tsorya:


omertuc:
Why?

tsorya:
There is no point to save them in list and precache as catalogs have pull always policy. 
In this specific case it breaks mirror registry sometimes, so in case default catalog sources were deleted but their images are still present on filesystem i still need to filter them

omertuc:
Please record this information in a code comment or your commit message

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,39 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
 	defer deadlineCancel()
 }
 
-func (s *SeedCreator) getCatalogImages(ctx context.Context) ([]string, error) {
+// Catalog sources have pull always policy and there is no point to precache them.
```
omertuc, line 498: Please follow go function doc conventions

e.g. `filterCatalogImages filters a list of images... so that...`

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,40 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
...
+
 	catalogSources := &operatorsv1alpha1.CatalogSourceList{}
 	allNamespaces := runtime.ListOptions{Namespace: metav1.NamespaceAll}
 	if err := s.client.List(ctx, catalogSources, &allNamespaces); err != nil {
 		return nil, fmt.Errorf("failed to list all catalogueSources %w", err)
 	}
 
 	for _, catalogSource := range catalogSources.Items {
-		images = append(images, catalogSource.Spec.Image)
+		if !lo.Contains(defaultCatalogsImages, strings.Split(catalogSource.Spec.Image, ":")[0]) {
```
omertuc, line 520: Split on `:` can break if registry has port numbers. Needs proper parsing

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,40 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
...
+			catalogImages = append(catalogImages, catalogSource.Spec.Image)
+		}
 	}
 
+	s.log.Infof("Removing list of catalog images from full image list, catalog images to remove %s",
+		append(catalogImages, defaultCatalogsImages...))
+	images = lo.Filter(images, func(image string, index int) bool {
```
omertuc, line 527: nit:

```suggestion
	images = lo.Filter(images, func(image string, _ int) bool {
```

ibu-imager/seedcreator/seedcreator.go
tsorya: Not relevant for this specific case, those are known registers so if it breaks we will match and this was expected for this case

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -234,15 +234,10 @@ func (s *SeedCreator) createContainerList(ctx context.Context) error {
 		return err
 	}
 	images := strings.Split(output, "\n")
-	catalogImages, err := s.getCatalogImages(ctx)
+	images, err = s.filterCatalogImages(ctx, images)
 	if err != nil {
 		return err
 	}
-	s.log.Infof("Removing list of catalog images from full image list, catalog images to remove %s", catalogImages)
```
omertuc, line 241: So wait, the old code was also filtering catalog images? So what difference does your new code make?

tsorya: https://github.com/openshift-kni/lifecycle-agent/pull/220/files#diff-af44b48262b65ed56f5f9d016b4adb3fa8f72e0313e666d5a325b97426b718a8R501

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,42 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
...
 }
 
-func (s *SeedCreator) getCatalogImages(ctx context.Context) ([]string, error) {
+// filterCatalogImages filters catalog source images as catalog sources have pull always policy
+// and there is no point to precache them.
+// List of catalog images are taken from catalog sources that are currently configured on cluster and known ocp ones.
+// Filtering known images will allow us to fix the case where those images were pulled in seed
+// and after it their catalog sources were removed but images are still part of podman images list as we create seed
+func (s *SeedCreator) filterCatalogImages(ctx context.Context, images []string) ([]string, error) {
+	defaultCatalogsImages := []string{"registry.redhat.io/redhat/community-operator-index",
```
eranco74, line 504: Why do we need the defaultCatalogImages + get the catalogSources list from the API?
Why not filter out any image that starts with one of these `registry...-index` ?

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,42 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
...
 	allNamespaces := runtime.ListOptions{Namespace: metav1.NamespaceAll}
 	if err := s.client.List(ctx, catalogSources, &allNamespaces); err != nil {
 		return nil, fmt.Errorf("failed to list all catalogueSources %w", err)
 	}
 
 	for _, catalogSource := range catalogSources.Items {
-		images = append(images, catalogSource.Spec.Image)
+		// Splitting on : is technically wrong here because of port numbers
+		// but in this circumstance the code is still correct despite that
+		if !lo.Contains(defaultCatalogsImages, strings.Split(catalogSource.Spec.Image, ":")[0]) {
```
eranco74, line 522: Do we really care about this condition?
Just join the two slices, you just use it as a filter later.

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,42 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
 	defer deadlineCancel()
 }
 
-func (s *SeedCreator) getCatalogImages(ctx context.Context) ([]string, error) {
+// filterCatalogImages filters catalog source images as catalog sources have pull always policy
+// and there is no point to precache them.
```
eranco74, line 499: Wait what?
The result of this PR is that the catalog source images will not be pre-cached?
Are you sure we want to do that?

ibu-imager/seedcreator/seedcreator.go
tsorya: https://github.com/openshift-kni/lifecycle-agent/pull/220/files#diff-af44b48262b65ed56f5f9d016b4adb3fa8f72e0313e666d5a325b97426b718a8R501

ibu-imager/seedcreator/seedcreator.go
tsorya: Yes

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,42 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
...
 	for _, catalogSource := range catalogSources.Items {
-		images = append(images, catalogSource.Spec.Image)
+		// Splitting on : is technically wrong here because of port numbers
+		// but in this circumstance the code is still correct despite that
+		if !lo.Contains(defaultCatalogsImages, strings.Split(catalogSource.Spec.Image, ":")[0]) {
+			catalogImages = append(catalogImages, catalogSource.Spec.Image)
+		}
 	}
 
+	s.log.Infof("Removing list of catalog images from full image list, catalog images to remove %s",
```
eranco74, line 527:
```suggestion
	s.log.Infof("Removing default catalog images from full image list, going to remove images that start with one of: %s", defaultCatalogsImages)
	s.log.Infof("Removing catalog images from full image list, catalog images to remove %s", catalogImages)
```

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +495,42 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
 	defer deadlineCancel()
 }
 
-func (s *SeedCreator) getCatalogImages(ctx context.Context) ([]string, error) {
+// filterCatalogImages filters catalog source images as catalog sources have pull always policy
+// and there is no point to precache them.
+// List of catalog images are taken from catalog sources that are currently configured on cluster and known ocp ones.
+// Filtering known images will allow us to fix the case where those images were pulled in seed
+// and after it their catalog sources were removed but images are still part of podman images list as we create seed
```
eranco74, line 502:
```suggestion
// and after it their catalog sources were removed but images are still part of crictl images output as we create seed
```

ibu-imager/seedcreator/seedcreator.go
eranco74: They aren't going to be there twice, the catalogImages will have the full image name.
While defaultCatalogsImages just have the prefix.

ibu-imager/seedcreator/seedcreator.go
eranco74, line 527: TBH, I'm not sure we want to log the images we might delete.
If we delete an image let's log that.

ibu-imager/seedcreator/seedcreator.go
```diff
@@ -500,19 +496,37 @@ func (s *SeedCreator) waitTillOvnKubeNodeIsDown(ctx context.Context) {
...
+// List of catalog images are taken from catalog sources that are currently configured on cluster and known ocp ones.
+// Filtering known images will allow us to fix the case where those images were pulled in seed
+// and after it their catalog sources were removed but images are still part of podman images output as we create seed
+func (s *SeedCreator) filterCatalogImages(ctx context.Context, images []string) ([]string, error) {
+	// Regex matches list of known catalog images, there are 4 of them at least in 4.15
+	//registry.redhat.io/redhat/community-operator-index:v4.15
+	//registry.redhat.io/redhat/redhat-operator-index:v4.15
+	//registry.redhat.io/redhat/certified-operator-index:v4.15
+	//registry.redhat.io/redhat/redhat-marketplace-index:v4.15
+	defaultCatalogsRegex := regexp.MustCompile(`^registry\.redhat\.io\/redhat\/.+?-index:.+?`)
```
omertuc, line 510: Why the question marks?

Also you don't have to escape `\/`, `/` works too

omertuc: The last question mark seems to be wrong:

image

omertuc:
/hold

eran comments

eranco74 (commented):
Nice!
/lgtm
/approve

PR #224, opened 2024-01-16
Update OWNERS to add Raul for Dockerfile and CSV updates

donpenney:


OWNERS
```diff
@@ -31,3 +31,9 @@ filters:
   ^ibu-imager/clusterinfo/clusterinfo.go:
     labels:
       - cluster-config-api-changed
+  ^Dockerfile:
+    reviewers:
+    - rauhersu
+  ^bundle/manifests/lifecycle-agent.clusterserviceversion.yaml:
```
donpenney, line 37: Updated to bundle file instead of config file, per discussion with @rauhersu

rauhersu: thanks @donpenney ! We will likely have CSVs under 'config' and 'bundle' dirs in sync on a MR basis but if something is relevant to midstream (builds we deliver to customers) is the 'bundle' one. Again, thanks for this initiative, this will definitely help to keep upstream-midstream dev in sync.

rauhersu: /lgtm

PR #227, opened 2024-01-16
Remove unused fields from IBU CRD

donpenney:
The initial IBU CRD definition included a spec field for RollbackTarget and a status field for recording StateRoots. These fields were ultimately unneeded. They are removed with this commit.

/cc @jc-rh @browsell


donpenney:
> Don't we want to store the initial stateroot name so we know which stateroot to rollback on? or more accurately the deployment id of the A side.
> 
> https://github.com/openshift-kni/lifecycle-agent/blob/9ff71276db124e563af5adad60f35f53eb15241a/controllers/rollback_handlers.go#L50
> 
> is the first deployment always the one to rollback to?

There will only be two stateroots. Rollback will be to the first deployment in the unbooted stateroot.

PR #230, opened 2024-01-17
MGMT-16572: Rename 'ibu-imager' to 'lca-cli'

leo8a:
Decision to rename `ibu-imager` to `lca-cli` is due to imager just does not define the full scope (i.e. steps within IBU and IBI workflows) of the tool anymore.

The `lca-cli` is now used for the following actions (not limited to):
    seed generation (IBU/IBI)
    post-pivot (IBI/IBU)
    seed installation (IBI)
    live iso generation (IBI)

/cc @eranco74 @donpenney 


omertuc:
Do we need to bump seedformatversion? Or would this new LCA version be compatible with older seeds?

omertuc:
We probably want to rename the env `SEEDGEN_IMAGER_SKIP_RECERT` too, as the name `IMAGER` doesn't really refer to anything anymore

omertuc:
`HACKING.md` still has a reference to "IBU Imager"

omertuc:
`seedcleanup.go` still has a comment saying `... comprises the Imager workflow ...`

omertuc:
`seedcreator.go` still has a comment saying `... comprises the Imager workflow for creating ...`

leo8a:
Thanks for spotting all those!

donpenney:
I think this is a great idea. The new name is also good if we decide to merge the currently 3 separate binaries into one, moving manager and precaching into subcommands, which I'd also like to do at some point. I'll have some time to review this afternoon

controllers/seedgen_controller.go
```diff
@@ -72,11 +72,11 @@ var (
 	lcaImage               string
 	seedgenAuthFile        = filepath.Join(utils.SeedgenWorkspacePath, "auth.json")
 	storedManagedClusterCR = filepath.Join(utils.SeedgenWorkspacePath, "managedcluster.json")
-	imagerContainerName    = "ibu_imager"
+	lcaCliContainerName    = "lca_cli"
```
donpenney, line 75: This should likely be something like lca_image_builder for the container name

Dockerfile
```diff
@@ -12,14 +12,14 @@ COPY go.* .
...
-COPY ibu-imager ibu-imager
+COPY lca-cli lca-cli
 COPY internal internal
 COPY main main
 COPY utils utils
 
 # Build the binaries
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a -o build/manager main/main.go
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a -o build/ibu-imager main/ibu-imager/main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a -o build/lca-cli main/lca-cli/main.go
```
donpenney, line 22: This will need to be coordinated with the downstream build. cc @rauhersu

leo8a:
Force pushed to rebase and added latest suggestions.

Dockerfile
rauhersu: yeah, ping me when you decide to merge this and I will align midstream to these  changes.

donpenney (approved):
Thanks Leo!

PR #232, opened 2024-01-17
MGMT-16601: allow installation without providing host ip in advance

tsorya:
[MGMT-16601](https://issues.redhat.com//browse/MGMT-16601): allow installation without providing host ip in advance
In case host ip was not provided in manifest.json post-pivot configuration will start manually nodeip service, wait till it creates primary-ip file and read ip from it.
After getting ip it will create dnsmasq and dispatcher script override file and restart NM with dnsmasq

### DIFF
--- a/ibu-imager/postpivot/postpivot.go
+++ b/ibu-imager/postpivot/postpivot.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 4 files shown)


ibu-imager/postpivot/postpivot.go
```diff
@@ -51,6 +52,11 @@ func NewPostPivot(scheme *runtime.Scheme, log *logrus.Logger, ops ops.Ops,
 	}
 }
 
+const (
+	nodeIpFile       = "/run/nodeip-configuration/primary-ip"
+	dnsmasqOverrides = "/etc/default/sno_dnsmasq_configuration_overrides"
+)
+
 func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
```
eranco74, line 60: This gets executed before or after prepare-installation-configuration.sh?

eranco74 (commented):
Overall this looks good!
Moving logic out of prepare-installation-configuration.sh script is great, I think that in the end it can just wait for the configuration, and configure the networking if network config was provided.
Please add UT...

ibu-imager/postpivot/postpivot.go
tsorya: Yep, it is installation-configuration.service and yes my idea to remove prepare-installation-configuration.sh at all at the end

PR #233, opened 2024-01-18
Use `condition.reason` instead of `condition.type` as state in printercolumns

sudomakeinstall2:
Condition.reason represent the state better than condition.type

Also:
- Remove precache summary from status messages
- Fix typo in main


sudomakeinstall2:
/hold

need to make a new seed to test

donpenney (commented):
/lgtm

PR #237, opened 2024-01-19
Extract, export and apply CRs from policies

jc-rh:
This is for CNF-9623. The purpose of this change is to capture the policies specific to the newer version of the OCP and apply the CRs post pivot. Usually the newer policies should have been applied on the seed node. However, cluster specific policy or policies with cluster specific templating will need to be applied on the target node.

### DIFF
--- a/internal/extramanifest/policy.go
+++ b/internal/extramanifest/policy.go
@@ -18,6 +18,7 @@ import (
 )
 
 // +kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=policies,verbs=get;list
+
 // This annotation helps with the ordering as well as whether the policy should be applied. This is the same expectation as
 // the normal ZTP process post non-image-based cluster installation.
 const ztpDeployWaveAnnotation = "ran.openshift.io/ztp-deploy-wave"
--- a/bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
+++ b/bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
@@ -354,6 +354,13 @@ spec:
           - list
           - watch
         - apiGroups:
+          - policy.open-cluster-management.io
+          resources:
+          - policies
+          verbs:
+          - get
+          - list
+        - apiGroups:
           - rbac.authorization.k8s.io
           resources:
           - clusterrolebindings
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -264,6 +264,13 @@ rules:
   - list
   - watch
 - apiGroups:
+  - policy.open-cluster-management.io
+  resources:
+  - policies
+  verbs:
+  - get
+  - list
+- apiGroups:
   - rbac.authorization.k8s.io
   resources:
   - clusterrolebindings


internal/extramanifest/extramanifest.go
```diff
@@ -124,6 +128,34 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
...
+		return err
+	}
+
+	for i, policy := range policies {
+		objects, err := getConfigurationObjects(policy)
+		if err != nil {
+			return err
+		}
+		for _, object := range objects {
+			fileName := strconv.Itoa(i) + "_" + object.GetName() + "_" + object.GetNamespace() + ".yaml"
```
omertuc, line 148: nit

```suggestion
			fileName := fmt.Sprintf("%d_%s_%s.yaml", strconv.Itoa(i), object.GetNamespace(), object.GetName())
```

internal/extramanifest/policy.go
```diff
@@ -0,0 +1,125 @@
...
+
+	policies := &policiesv1.PolicyList{}
+	if err := h.Client.List(ctx, policies, listOpts...); err != nil {
+		return nil, err
+	}
+
+	var policyWaveMap = make(map[*policiesv1.Policy]int)
+
+	for _, policy := range policies.Items {
+		// Ignore policies with remediationAction enforce
```
omertuc, line 37: It's better if comments explain **why** we do things, as opposed to **what** we're doing (unless it's not obvious). 

In this particular case it's hard to know why you chose to ignore those policies without such comment. The commit message is also completely empty - if someone were to come back to this code with the hopes of changing it, they will be in the dark

internal/extramanifest/policy.go
```diff
@@ -0,0 +1,125 @@
...
+		if strings.EqualFold(string(policy.Spec.RemediationAction), "enforce") {
+			h.Log.Info("Ignoring policy " + policy.Name + " with remediationAction enforce")
```
omertuc, line 39: nit

```suggestion
			h.Log.Info(fmt.Sprintf("Ignoring policy %s with remediationAction enforce", policy.Name))
```

internal/extramanifest/policy.go
```diff
@@ -0,0 +1,125 @@
...
+		deployWave, found := policy.GetAnnotations()[ztpDeployWaveAnnotation]
+		if found {
+			deployWaveInt, err := strconv.Atoi(deployWave)
+			if err != nil {
+				// err convert from string to int
+				h.Log.Error(err, "%s in policy %s is not an interger: %s", ztpDeployWaveAnnotation, policy.GetName())
+			}
+			_, err = getParentPolicyNameAndNamespace(policy.GetName())
+			if err != nil {
+				h.Log.Info("Ignoring policy " + policy.Name + " with invalid name")
```
omertuc, line 52: nit

```suggestion
				h.Log.Info(fmt.Sprintf("Ignoring policy %s with invalid name", policy.Name))
```

internal/extramanifest/policy.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=policies,verbs=get;list
+
+const ztpDeployWaveAnnotation = "ran.openshift.io/ztp-deploy-wave"
```
omertuc, line 21: A word about what `ztp-deploy-wave` is and why we sort the policies according to it? Google was not really helpful here

internal/extramanifest/policy.go
```diff
@@ -0,0 +1,125 @@
...
+	var policyWaveMap = make(map[*policiesv1.Policy]int)
+
+	for _, policy := range policies.Items {
+		// Ignore policies with remediationAction enforce
+		if strings.EqualFold(string(policy.Spec.RemediationAction), "enforce") {
+			h.Log.Info("Ignoring policy " + policy.Name + " with remediationAction enforce")
+			continue
+		}
+
+		deployWave, found := policy.GetAnnotations()[ztpDeployWaveAnnotation]
```
omertuc, line 43: Do we simply ignore all policies without this wave annotation? Is this the intention? 

If all policies should have wave, should this maybe be an error instead?

internal/extramanifest/policy.go
```diff
@@ -0,0 +1,125 @@
...
+		if found {
+			deployWaveInt, err := strconv.Atoi(deployWave)
+			if err != nil {
+				// err convert from string to int
+				h.Log.Error(err, "%s in policy %s is not an interger: %s", ztpDeployWaveAnnotation, policy.GetName())
```
omertuc, line 48:
```suggestion
				h.Log.Error(err, "%s in policy %s is not an integer: %s", ztpDeployWaveAnnotation, policy.GetName())
```

omertuc: Also this is `Error`, not some form of `Errorf` - as in, it accepts a list of "keys" and "values" (keys in even positions, values in odd positions), not merely format string parameters

omertuc (changes requested):
Some nits and requests for clarification. 

Is there a ticket relevant for this work?

go.mod
```diff
@@ -81,20 +82,23 @@ require (
...
+	k8s.io/klog/v2 v2.110.1 // indirect
+	k8s.io/kube-openapi v0.0.0-20231010175941-2dd684a91f00 // indirect
 	k8s.io/utils v0.0.0-20230726121419-3b25d923346b // indirect
+	open-cluster-management.io/multicloud-operators-subscription v0.11.0 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+replace k8s.io/client-go => k8s.io/client-go v0.28.2
```
browsell, line 102: why this version ?

internal/extramanifest/policy.go
browsell: A mechanism to order manifest application

controllers/upgrade_handlers.go
```diff
@@ -148,6 +148,11 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedUp
 	}
 
 	u.Log.Info("Writing extra-manifests into new stateroot")
+	labels := map[string]string{"lca.openshift.io/for-ocp-version": ibu.Spec.SeedImageRef.Version}
+	if err := u.ExtraManifest.ExtractAndExportManifestFromPoliciesToDir(ctx, labels, staterootVarPath); err != nil {
```
browsell, line 152: How do we handle the no policy case (i.e. SNO that is not using acm/gitops) ?

jc-rh:
> Some nits and requests for clarification.
> 
> Is there a ticket relevant for this work?

https://issues.redhat.com/browse/CNF-9623

internal/extramanifest/policy.go
jc-rh: Copied code from TALM :stuck_out_tongue:

internal/extramanifest/policy.go
jc-rh: Currently, non image based install process uses this to order the policy enforcement as part of ZTP.

internal/extramanifest/policy.go
jc-rh: The annotation helps with the ordering as well as whether the policy should be included in the auto created CGU post cluster installation. Our reference du profile policies all have it and we ask customers do the same to their policies if they are meant to be included.

internal/extramanifest/policy.go
jc-rh: There is a plan to replace logr with logrus. But I will fix it anyways.

internal/extramanifest/extramanifest.go
```diff
@@ -37,11 +37,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const ExtraManifestPath = "/opt/extra-manifests"
+const (
+	ExtraManifestPath  = "/opt/extra-manifests"
```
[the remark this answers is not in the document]
jc-rh: We have other things saved under /opt. I will create a jira to track that.

go.mod
jc-rh: We are not locked to this version. For some reason, importing the policy APIs forces us to do v12 incompatible in the require block and replaces it with the actual version here.

controllers/upgrade_handlers.go
jc-rh: Maybe another spec field where the actual manifests are stored in config maps, instead of just name/kind for exporting.

internal/extramanifest/extramanifest.go
jc-rh: Updated

internal/extramanifest/policy.go
jc-rh: Comment added

internal/extramanifest/policy.go
jc-rh: Updated

internal/extramanifest/policy.go
jc-rh: Updated

internal/extramanifest/policy.go
sudomakeinstall2: I don't think we can replace logr completely. controller-sdk uses logr.

browsell: Migration to logrus tracked under  CNF-10370

controllers/upgrade_handlers.go
jc-rh: Will add a CRD check

internal/extramanifest/policy.go
omertuc: So all policies have it, then my question remains:

> should this maybe be an error instead?

controllers/upgrade_handlers.go
```diff
@@ -148,6 +150,11 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedUp
 	}
 
 	u.Log.Info("Writing extra-manifests into new stateroot")
```
browsell, line 152: Can we add some comments here on  this mechanism.

jc-rh: Updated

internal/extramanifest/policy.go
jc-rh: > So all policies have it, then my question remains:
> 
> > should this maybe be an error instead?

All policies that should be auto applied as part of new install or upgrade should have it. It's possible that other policies are placed to some cluster temporarily.

jc-rh: > I don't think we can guarantee that customer provided policies would have the wave annotation.

That's our current requirement if customer wants to include their policy as part of the ZTP CGU.

controllers/upgrade_handlers.go
```diff
@@ -148,6 +150,14 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedUp
...
+	// Extract from policies can be done by matching labels on the policy or the CR itself
+	// Currently we expect user to properly label CRs with site specific content
+	// as those policies should/can not be applied on the seed
```
browsell, line 155: nit, must not

internal/extramanifest/policy.go
browsell: if we require all inform policies to have a wave then a policy without a wave is an error

controllers/upgrade_handlers.go
jc-rh: updated

sakhoury (commented):
/lgtm

PR #245, opened 2024-01-24
Generate cluster id in case it's not in the seedReconfig

eranco74:


Makefile
```diff
@@ -61,7 +61,7 @@ CRD_OPTIONS ?= "crd"
 default: help
 
 test:
-	@echo "Stub test target"
+	go test ./... -coverprofile cover.out
```
omertuc, line 64: We already have `make unittest`, delete this one

lca-cli/postpivot/postpivot.go
```diff
@@ -377,8 +380,8 @@ func (p *PostPivot) setNewClusterID(ctx context.Context, client runtimeclient.Cl
 	}
 
 	if seedReconfiguration.ClusterID == "" {
-		// TODO: Generate a fresh cluster ID in case it is empty
-		return fmt.Errorf("cluster id is empty, automatic generation is not yet implemented")
+		p.log.Info("cluster id is empty, generating...")
+		seedReconfiguration.ClusterID = uuid.New().String()
```
omertuc, line 384: Log only after and include the ID in the log?

lca-cli/postpivot/postpivot.go
```diff
@@ -377,8 +380,8 @@ func (p *PostPivot) setNewClusterID(ctx context.Context, client runtimeclient.Cl
 	}
 
 	if seedReconfiguration.ClusterID == "" {
```
browsell, line 382: This would be a fatal error on IBU.

eranco74: In IBU you will always have the clusterID as part of the config, otherwise it should fail during prep.
TBH, this change is not a must for IBI, we plan to generate the cluster ID in IBIO (we just didn't figure out where to store it yet).

eranco74:
Following https://github.com/openshift/image-based-install-operator/pull/34, this is no longer required

PR #246, opened 2024-01-24
Use recert stable tag

omertuc:
The `:latest` tag of the recert image is not stable, it contains changes
from the `main` branch. We should default to the `:v0` tag instead which
we merge into less often.


eranco74:
```
fatal: detected dubious ownership in repository at '/go/src/github.com/openshift-kni/lifecycle-agent'
```
The failure seems unrelated to the change
/retest

leo8a:
should we also update the `HACKING.md` and `lca-cli/README.md` files to reflect the use of the `:v0` tag... in those guides for disconn envs we have `latest` as the tag to mirror

PR #247, opened 2024-01-24
Fix late update of in-progress status

sudomakeinstall2:
update CR status and requeue immediately after stage transition so that in-progress status is visible properly.

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -204,11 +204,7 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 					return
 				}
 			}
-			nextReconcile, err = r.handleStage(ctx, ibu, ibu.Spec.Stage)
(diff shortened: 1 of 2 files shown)


controllers/ibu_controller.go
```diff
@@ -204,11 +204,8 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 					return
 				}
 			}
-			nextReconcile, err = r.handleStage(ctx, ibu, ibu.Spec.Stage)
-			if err != nil {
-				_ = utils.UpdateIBUStatus(ctx, r.Client, ibu)
-				return
-			}
+			err = utils.UpdateIBUStatus(ctx, r.Client, ibu)
+			return requeueImmediately(), err
```
jc-rh, line 208:
```suggestion
			nextReconcile =  requeueImmediately()
```

jc-rh (approved):

PR #250, opened 2024-01-25
MGMT-15783: moving pull secret to seed reconfiguration struct.

tsorya:
IBU/I will set pull secret in struct and post pivot will create pull-secret secret that will be applied with all the manifests whenever
cluster will be up
Flow:
1. We will move seed pull secret aside, write new pull secret to
/var/lib/kubelet/config.json.
2. After api is up and images were pulled, we will write back seed pull
   secret in order for MCO to be happy
3. After it we will apply new pull secret to cluster

### DIFF
--- a/api/seedreconfig/seedreconfig.go
+++ b/api/seedreconfig/seedreconfig.go
@@ -74,6 +74,9 @@ type SeedReconfiguration struct {
 	// Example of nmstate configurations can be found in this link https://nmstate.io/examples.html
 	// This field will be used for IBI process as in IBU we will copy nmconnection files
 	RawNMStateConfig string `json:"raw_nm_state_config,omitempty"`
+
+	// PullSecret is the secret to use when pulling images. Equivalent to install-config.yaml's pullSecret.
+	PullSecret string `json:"pull_secret,omitempty"`
 }
 
 type KubeConfigCryptoRetention struct {
--- a/internal/clusterconfig/clusterconfig.go
+++ b/internal/clusterconfig/clusterconfig.go
@@ -36,10 +36,7 @@ const (
 	proxyName     = "cluster"
 	proxyFileName = "proxy.json"
 
-	pullSecretName     = "pull-secret"
-	pullSecretFileName = "pullsecret.json"
-
-	configNamespace = "openshift-config"
+	pullSecretName = "pull-secret"
 
 	idmsFileName  = "image-digest-mirror-set.json"
 	icspsFileName = "image-content-source-policy-list.json"
@@ -83,9 +80,6 @@ func (r *UpgradeClusterConfigGather) FetchClusterConfig(ctx context.Context, ost
 	}
 	manifestsDir := filepath.Join(clusterConfigPath, manifestDir)
 
-	if err := r.fetchPullSecret(ctx, manifestsDir); err != nil {
-		return err
-	}
 	if err := r.fetchProxy(ctx, manifestsDir); err != nil {
 		return err
 	}
@@ -110,31 +104,10 @@ func (r *UpgradeClusterConfigGather) FetchClusterConfig(ctx context.Context, ost
 	return nil
 }
 
-func (r *UpgradeClusterConfigGather) fetchPullSecret(ctx context.Context, manifestsDir string) error {
-	r.Log.Info("Fetching pull-secret", "name", pullSecretName, "namespace", configNamespace)
-
-	secret := corev1.Secret{}
-	if err := r.Client.Get(ctx, types.NamespacedName{Name: pullSecretName, Namespace: configNamespace}, &secret); err != nil {
-		return err
-	}
-
-	s := corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      secret.Name,
-			Namespace: secret.Namespace,
-		},
-		Data: secret.Data,
-		Type: secret.Type,
-	}
-	typeMeta, err := r.typeMetaForObject(&s)
-	if err != nil {
-		return err
-	}
-	s.TypeMeta = *typeMeta
-
-	filePath := filepath.Join(manifestsDir, pullSecretFileName)
-	r.Log.Info("Writing pull-secret to file", "path", filePath)
-	return utils.MarshalToFile(s, filePath)
+func (r *UpgradeClusterConfigGather) fetchPullSecret(ctx context.Context) (string, error) {
+	r.Log.Info("Fetching pull-secret")
+	return utils.GetSecretData(
+		ctx, common.PullSecretName, common.OpenshiftConfigNamespace, corev1.DockerConfigJsonKey, r.Client)
 }
 
 func (r *UpgradeClusterConfigGather) fetchProxy(ctx context.Context, manifestsDir string) error {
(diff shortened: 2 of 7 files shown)


api/seedreconfig/seedreconfig.go
```diff
@@ -67,6 +67,9 @@ type SeedReconfiguration struct {
 	// Example of nmstate configurations can be found in this link https://nmstate.io/examples.html
 	// This field will be used for IBI process as in IBU we will copy nmconnection files
 	RawNMStateConfig string `json:"raw_nm_state_config,omitempty"`
+
+	// The pull secret that obtained from the Pull Secret page on the Red Hat OpenShift Cluster Manager site.
```
omertuc, line 71:
```suggestion
	// PullSecret is the secret to use when pulling images.
```

(taken from the [installer comment](https://github.com/openshift/installer/blob/31607ba7ad2111ebcaf0d07c5387c211ab3f16b8/pkg/types/installconfig.go#L137))

internal/common/consts.go
```diff
@@ -66,6 +66,9 @@ const (
 	// Bump this every time the seed format changes in a backwards incompatible way
 	SeedFormatVersion  = 3
 	SeedFormatOCILabel = "com.openshift.lifecycle-agent.seed_format_version"
+
+	PullSecretName  = "pull-secret"
+	ConfigNamespace = "openshift-config"
```
omertuc, line 71:
```suggestion
	OpenshiftConfigNamespace = "openshift-config"
```

lca-cli/postpivot/postpivot.go
```diff
@@ -541,3 +548,31 @@ func (p *PostPivot) applyNMStateConfiguration(seedReconfiguration *clusterconfig
 
 	return nil
 }
+
+func (p *PostPivot) createPullSecretFile(seedReconfiguration *clusterconfig_api.SeedReconfiguration, pullSecretFile string) error {
+	if seedReconfiguration.PullSecret == "" {
+		p.log.Infof("Pull secret was not provided")
```
omertuc, line 554: Add TODO to make this a hard error, as seed images in the future are not going to have a pull secret and the cluster can't operate without it

lca-cli/postpivot/postpivot.go
```diff
@@ -541,3 +548,31 @@ func (p *PostPivot) applyNMStateConfiguration(seedReconfiguration *clusterconfig
 
 	return nil
 }
+
+func (p *PostPivot) createPullSecretFile(seedReconfiguration *clusterconfig_api.SeedReconfiguration, pullSecretFile string) error {
```
omertuc, line 552:
```suggestion
func (p *PostPivot) createPullSecretFile(pullSecret string, pullSecretFile string) error {
```

lca-cli/postpivot/postpivot.go
```diff
@@ -58,6 +59,7 @@ const (
 	sshKeyEarlyAccessFile = "/home/core/.ssh/authorized_keys.d/ib-early-access"
 	userCore              = "core"
 	sshMachineConfig      = "99-%s-ssh"
+	pullSecretFileName    = "pullsecret.json"
```
omertuc, line 62: What is the significance of this file? Why are we creating it?

tsorya: same as with machine config, adding it to manifests folder and will apply everything from there. 
Will add comment to the function

eranco74: Why do we need to create the file? why not create the secret from code?
Also note that @leo8a is removing the secret from the seed [here](https://github.com/openshift-kni/lifecycle-agent/pull/256)

tsorya: We can do it but i think we will need first of all to create docker config on filesystem, after Leo will remove old ps (assumption) and in this case this logic will be the same as with ssh key, we handle everything the same way

eranco74: I don't see how creating a secret file to be applied to k8s relates to the docker config file (just the secret value decoded in json format) that we should place on the filesystem

eranco74: @tsorya it's worth adding this :point_up: as a comment in the code (above line 84)

lca-cli/postpivot/postpivot.go
```diff
@@ -113,6 +121,11 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
 	}
 	p.waitForApi(ctx, client)
 
+	if err := utils.MoveFileIfExists(common.ImageRegistryAuthFile+seedPullSecretPrefix,
```
eranco74, line 124: What's this about?
what exactly happens here?

lca-cli/postpivot/postpivot.go
```diff
@@ -541,3 +554,47 @@ func (p *PostPivot) applyNMStateConfiguration(seedReconfiguration *clusterconfig
...
+// createPullSecretFile creates auth file on filesystem in order to be able to pull images
+// and runs createPullSecretManifest to write secret in manifests folder
+func (p *PostPivot) createPullSecretFile(pullSecret, pullSecretFile, pullSecretManifest string) error {
+	// TODO: Should return error in the future as cluster will not be operational without it
+	if pullSecret == "" {
+		p.log.Infof("Pull secret was not provided")
+		return nil
+	}
+	p.log.Infof("Move seed PS file aside")
+	if err := utils.MoveFileIfExists(pullSecretFile, pullSecretFile+"_seed"); err != nil {
```
eranco74, line 567: I don't think the seed PS should be there, see https://github.com/openshift-kni/lifecycle-agent/pull/256

lca-cli/postpivot/postpivot.go
```diff
@@ -541,3 +554,47 @@ func (p *PostPivot) applyNMStateConfiguration(seedReconfiguration *clusterconfig
...
+		return err
+	}
+
+	p.log.Infof("Writing provided pull secret to %s", pullSecretFile)
+	if err := os.WriteFile(pullSecretFile, []byte(pullSecret), 0o600); err != nil {
+		return fmt.Errorf("failed to write pull secret to %s, err %w", pullSecret, err)
+	}
+
+	return p.createPullSecretManifest(pullSecret, pullSecretManifest)
```
eranco74, line 576: Consider moving this function call to wherever you call `createPullSecretFile`.
Or rename `createPullSecretFile` to `createPullSecretFileAndManifest`

lca-cli/postpivot/postpivot.go
tsorya: If it will not be there, nothing will happened, though i assume it should be just empty

tsorya:
We should consider moving ps logic to recert, as it already does similar things, but to unblock the ticket and allow ps cleanup, lets push this code and then we can remove it and move to recert

internal/clusterconfig/clusterconfig.go
```diff
@@ -39,7 +39,7 @@ const (
 	pullSecretName     = "pull-secret"
 	pullSecretFileName = "pullsecret.json"
 
-	configNamespace = "openshift-config"
+	OpenshiftConfigNamespace = "openshift-config"
```
eranco74, line 42: You can use the one from consts.go

lca-cli/postpivot/postpivot.go
```diff
@@ -58,6 +58,11 @@ const (
 	sshKeyEarlyAccessFile = "/home/core/.ssh/authorized_keys.d/ib-early-access"
 	userCore              = "core"
 	sshMachineConfig      = "99-%s-ssh"
+
+	// secret yaml with ps in it, will be added to manifests folder,
+	// it will be applied with all other manifests
+	pullSecretFileName   = "pull-secret.json"
+	seedPullSecretPrefix = "_seed"
```
eranco74, line 65:
```suggestion
	seedPullSecretSuffix = "_seed"
```

tsorya:
backward compatible
/remove-label cluster-config-api-changed

PR #252, opened 2024-01-25
Add `InfraID` to `SeedReconfiguration`

omertuc:
During installation, OpenShift installer takes the cluster name and
appends a few random characters to it (the exact algorithm can be seen
here [1]) and calls it the infrastructure ID.

This ID is then scattered all over etcd and the filesystem, and just
like the cluster name, cluster base domain and hostname, recert supports
changing the infra-ID (as it is derived from the cluster-name)

If not given, recert generates a random infra-ID from the cluster name.
But recert also supports passing a custom infra ID to it.

In case of IBU, we want to use the infra ID of the original SNO. This is
taken from the OpenShift config.openshift.io/v1 Infrastructure CR, under
`.status.infrastructureName`.

In case of IBI, we want IBIO to pregenerate the infra ID as IBIO needs
to pass it to ClusterInstall

To achieve that, this commit adds the InfraID parameter to the
`SeedReconfiguration` struct to allow both LCA and IBIO to set the infra
ID. This parameter is then passed to recert via the recert configuration
file.

This commit is not a breaking change, as if this field is left empty,
recert will simply generate a random one.

[1] https://github.com/openshift/installer/blob/e91df626c10e569e7613249053b7b9b264db42df/pkg/asset/installconfig/clusterid.go#L57-L79


omertuc:
>This commit is not a breaking change, as if this field is left empty,
recert will simply generate a random one.

/remove-label cluster-config-api-changed

omertuc:
> This commit is not a breaking change, as if this field is left empty,
> recert will simply generate a random one.

/remove-label cluster-config-api-changed

eranco74 (commented):
/lgtm

omertuc:
> This commit is not a breaking change, as if this field is left empty,
> recert will simply generate a random one.

/remove-label cluster-config-api-changed

omertuc:
Tested, ready for review

omertuc:
> This commit is not a breaking change, as if this field is left empty,
> recert will simply generate a random one.

/remove-label cluster-config-api-changed

PR #256, opened 2024-01-26
CNF-11033: Delete pull secret from seed image

leo8a:
We shouldn't be distributing the pull-secret of seed clusters. This cleans up the seed cluster of its pull-secret before seed creation to avoid unwanted distribution of this secret.

Also, added a commit to restore the original pull-secret in the seed cluster as soon as it gets back from seed creation for the cluster operators needing it.

### DIFF
--- a/controllers/seedgen_controller.go
+++ b/controllers/seedgen_controller.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
@@ -256,6 +257,30 @@ func (r *SeedGeneratorReconciler) currentAcmCrds(ctx context.Context) (acmCrdLis
 	return
 }
 
+func (r *SeedGeneratorReconciler) sanitizePullSecret(ctx context.Context) error {
+	// Sanitize cluster's pull-secret from sensitive data
+	sanitizedPullSecret, _ := lcautils.UpdatePullSecretFromDockerConfig(ctx, r.Client, []byte(common.PullSecretEmptyData))
+
+	deadlineCtx, deadlineCancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer deadlineCancel()
+	err := wait.PollUntilContextCancel(deadlineCtx, 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
+		r.Log.Info("Waiting for MCO to override pull-secret file")
+		dockerConfigJSON, err := os.ReadFile(filepath.Join(common.Host, common.ImageRegistryAuthFile))
+		if err != nil {
+			r.Log.Info(fmt.Sprintf("Failed to read %s file with error %s, will retry",
+				common.ImageRegistryAuthFile, err))
+			return false, nil
+		}
+		r.Log.Info(fmt.Sprintf("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON))))
+		return strings.TrimSpace(string(dockerConfigJSON)) == string(sanitizedPullSecret.Data[".dockerconfigjson"]), nil
+	})
+	if err != nil {
+		return fmt.Errorf("timed out waiting for MCO to override pull-secret file: %w", err)
+	}
+
+	return nil
+}
+
 // Clean up ACM and other resources on the cluster
 func (r *SeedGeneratorReconciler) cleanupClusterResources(ctx context.Context) error {
 	// Ensure that the dependent resources are deleted
@@ -368,6 +393,11 @@ func (r *SeedGeneratorReconciler) cleanupClusterResources(ctx context.Context) e
 		}}
 	if err := r.Client.Delete(ctx, observabilitySecret, deleteOpts...); client.IgnoreNotFound(err) != nil {
 		return fmt.Errorf("failed to delete observability secret: %w", err)
+	}
+
+	r.Log.Info("Sanitize cluster's pull-secret before seed creation")
+	if err := r.sanitizePullSecret(ctx); err != nil {
+		return fmt.Errorf("failed sanitizing cluster's pull-secret: %w", err)
 	}
 
 	return nil
@@ -644,6 +674,17 @@ func (r *SeedGeneratorReconciler) generateSeedImage(ctx context.Context, seedgen
 		}
 	} else {
 		r.Log.Info(fmt.Sprintf("No hubKubeconfig found in secret %s. Skipping hub interaction", utils.SeedGenSecretName))
+	}
+
+	// Get the cluster's pull-secret
+	clusterPullSecretData, err := lcautils.GetSecretData(ctx, common.PullSecretName, common.OpenshiftConfigNamespace, corev1.DockerConfigJsonKey, r.Client)
+	if err != nil {
+		return fmt.Errorf("could not access pull-secret %s in %s: %w", common.PullSecretName, common.OpenshiftConfigNamespace, err)
+	}
+
+	// Save the cluster's pull-secret in order to restore it after the lca-cli is complete
+	if err := os.WriteFile(common.PathOutsideChroot(utils.StoredPullSecret), []byte(clusterPullSecretData), 0o600); err != nil {
+		return fmt.Errorf("failed to write pull-secret to %s: %w", utils.StoredPullSecret, err)
 	}
 
 	// Clean up cluster resources
--- a/lca-cli/seedcreator/seedcreator.go
+++ b/lca-cli/seedcreator/seedcreator.go
@@ -303,14 +303,15 @@ func (s *SeedCreator) backupVar() error {
 
 	// Define the 'exclude' patterns
 	excludePatterns := []string{
+		"*/.bash_history",
 		"/var/tmp/*",
-		"/var/lib/log/*",
 		"/var/log/*",
+		"/var/lib/lca",
+		"/var/lib/log/*",
+		"/var/lib/cni/bin/*",
 		"/var/lib/containers/*",
 		"/var/lib/kubelet/pods/*",
-		"/var/lib/cni/bin/*",
-		"/var/lib/ovn-ic/etc/ovnkube-node-certs/*",
-		"/var/lib/lca",
+		common.OvnNodeCerts + "/*",
 	}
 
 	// Build the tar command
--- a/main/main.go
+++ b/main/main.go
@@ -374,6 +374,19 @@ func initSeedGen(ctx context.Context, c client.Client, log *logr.Logger) error {
 		return err
 	}
 
+	// Restore original pull-secret after seed creation.
+	// During seed creation, the pull-secret was removed; it needs to be restored back
+	// to allow the cluster operators to fully recover in the seed cluster.
+	log.Info("Restore original pull-secret after seed creation")
+	dockerConfigJSON, err := os.ReadFile(common.PathOutsideChroot(utils.StoredPullSecret))
+	if err != nil {
+		return fmt.Errorf("failed to read original pull-secret from %s file: %w", utils.StoredPullSecret, err)
+	}
+
+	if _, err := lcautils.UpdatePullSecretFromDockerConfig(ctx, c, dockerConfigJSON); err != nil {
+		return fmt.Errorf("failed to restore original pull-secret in seed cluster: %w", err)
+	}
+
 	log.Info("Restore successful and saved SeedGenerator CR removed")
 	return nil
 }
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -302,3 +302,24 @@ func MoveFileIfExists(source, dest string) error {
 	}
 	return nil
 }
+
+func UpdatePullSecretFromDockerConfig(ctx context.Context, c client.Client, dockerConfigJSON []byte) (*corev1.Secret, error) {
+	newPullSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.PullSecretName,
+			Namespace: common.OpenshiftConfigNamespace,
+		},
+		Data: map[string][]byte{
+			".dockerconfigjson": dockerConfigJSON,
+		},
+		Type: corev1.SecretTypeDockerConfigJson,
+	}
+
+	if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
+		return c.Update(ctx, newPullSecret)
+	}); err != nil {
+		return nil, fmt.Errorf("failed to update pull-secret resource: %w", err)
+	}
+
+	return newPullSecret, nil
+}
--- a/controllers/utils/constants.go
+++ b/controllers/utils/constants.go
@@ -21,4 +21,6 @@ const (
 var (
 	SeedGenStoredCR       = filepath.Join(SeedgenWorkspacePath, "seedgen-cr.json")
 	SeedGenStoredSecretCR = filepath.Join(SeedgenWorkspacePath, "seedgen-secret.json")
+
+	StoredPullSecret = filepath.Join(SeedgenWorkspacePath, "pull-secret.json")
 )
--- a/internal/common/consts.go
+++ b/internal/common/consts.go
@@ -78,6 +78,7 @@ const (
 	SeedFormatOCILabel = "com.openshift.lifecycle-agent.seed_format_version"
 
 	PullSecretName           = "pull-secret"
+	PullSecretEmptyData      = "{\"auths\":{\"registry.connect.redhat.com\":{\"username\":\"empty\",\"password\":\"empty\",\"auth\":\"ZW1wdHk6ZW1wdHk=\",\"email\":\"\"}}}"
 	OpenshiftConfigNamespace = "openshift-config"
 )
 


main/main.go
```diff
@@ -356,6 +358,32 @@ func initSeedGen(ctx context.Context, c client.Client, log *logr.Logger) error {
...
+		return fmt.Errorf("failed to read .dockerconfigjson file: %w", err)
+	}
+
+	pullSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.PullSecretName,
+			Namespace: "openshift-config",
+		},
+		Data: map[string][]byte{
+			".dockerconfigjson": dockerConfigJSON,
```
eranco74, line 376: I think the value in the secret should be base64 encoded while the file content isn't

eranco74 (commented):
Looks good, just need to fix the encoding in the restore operation

lca-cli/seedcreator/seedcreator.go
```diff
@@ -304,12 +304,13 @@ func (s *SeedCreator) backupVar() error {
...
-		"/var/lib/log/*",
 		"/var/log/*",
+		"/var/lib/log/*",
+		"/var/lib/cni/bin/*",
 		"/var/lib/containers/*",
 		"/var/lib/kubelet/pods/*",
-		"/var/lib/cni/bin/*",
-		"/var/lib/ovn-ic/etc/ovnkube-node-certs/*",
+		common.OvnNodeCerts + "/*",
+		common.ImageRegistryAuthFile,
```
tsorya, line 313: @eranco74 @leo8a i actually think it will not be very easy just to avoid bringing file. 
File is managed by mco and we can't just override it.

tsorya: @leo8a did you try to run full flow and see what is going on with cluster after reboot?

omertuc: We also need to consider how to change the pull secret reboot-free, does MCO even allow changing it without a reboot?

leo8a: I see, thanks for commenting.

will /hold this and try full flow in a 2e2 env to see what could go wrong (if any) after the pivot.

main/main.go
leo8a: thanks for spotting this!

main/main.go
```diff
@@ -356,6 +359,32 @@ func initSeedGen(ctx context.Context, c client.Client, log *logr.Logger) error {
...
+		return fmt.Errorf("failed to read .dockerconfigjson file: %w", err)
+	}
+
+	pullSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.PullSecretName,
+			Namespace: "openshift-config",
+		},
+		Data: map[string][]byte{
+			".dockerconfigjson": []byte(base64.StdEncoding.EncodeToString(dockerConfigJSON)),
```
tsorya, line 377: I actually sure you don't need to encode it, it works without it too, secret is automatically encoded

tsorya:
@leo8a @eranco74 so i found the way that works and is pretty ok as for me.
We should create empty json pull-secret (or some bad data) and not to delete it, wait till mco will apply it on the system and add this file to archive as before, something like 
`func (s *SeedCreator) cleanupPS() error {
	s.log.Infof("Creating empty pull secret manifest")
	cleanPS := "{}"
	ps := corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      common.PullSecretName,
			Namespace: common.OpenshiftConfigNamespace,
		},
		Data: map[string][]byte{corev1.DockerConfigJsonKey: []byte(cleanPS)},
		Type: corev1.SecretTypeDockerConfigJson,
	}
	typeMeta, err := utils.TypeMetaForObject(s.client.Scheme(), &ps)
	if err != nil {
		return fmt.Errorf("failed to create typeMetafor pull secret, err: %w", err)
	}
	ps.TypeMeta = *typeMeta

	if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
		return runtime.IgnoreAlreadyExists(s.client.Update(context.TODO(), &ps))
	}); err != nil {
		return fmt.Errorf("failed to create pull-secret in seed cluster: %w", err)
	}

	_ = wait.PollUntilContextCancel(context.TODO(), 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
		s.log.Info("Waiting till mco will override pull secret file")
		dockerConfigJSON, err := os.ReadFile(common.ImageRegistryAuthFile)
		if err != nil {
			s.log.Infof("Failed to read %s with error %s, will retry",
				common.ImageRegistryAuthFile, err.Error())
			return false, nil
		}
		s.log.Infof("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON)))
		return strings.TrimSpace(string(dockerConfigJSON)) == cleanPS, nil
	})

	return nil
}`

On post-pivot operation i will copy this file aside create new one, with relevant pull secret. Then i will wait till api is up (or maybe till mco is actually up) and then copy it back, This seems to make mco happy, after that i will override pull-secret while applying pull-secret manifest.
Tested the scenario and it works, WDYT?

eranco74:
> @leo8a @eranco74 so i found the way that works and is pretty ok as for me. We should create empty json pull-secret (or some bad data) and not to delete it, wait till mco will apply it on the system and add this file to archive as before, something like `func (s *SeedCreator) cleanupPS() error { s.log.Infof("Creating empty pull secret manifest") cleanPS := "{}" ps := corev1.Secret{ ObjectMeta: metav1.ObjectMeta{ Name: common.PullSecretName, Namespace: common.OpenshiftConfigNamespace, }, Data: map[string][]byte{corev1.DockerConfigJsonKey: []byte(cleanPS)}, Type: corev1.SecretTypeDockerConfigJson, } typeMeta, err := utils.TypeMetaForObject(s.client.Scheme(), &ps) if err != nil { return fmt.Errorf("failed to create typeMetafor pull secret, err: %w", err) } ps.TypeMeta = *typeMeta
> 
> ```
> if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
> 	return runtime.IgnoreAlreadyExists(s.client.Update(context.TODO(), &ps))
> }); err != nil {
> 	return fmt.Errorf("failed to create pull-secret in seed cluster: %w", err)
> }
> 
> _ = wait.PollUntilContextCancel(context.TODO(), 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
> 	s.log.Info("Waiting till mco will override pull secret file")
> 	dockerConfigJSON, err := os.ReadFile(common.ImageRegistryAuthFile)
> 	if err != nil {
> 		s.log.Infof("Failed to read %s with error %s, will retry",
> 			common.ImageRegistryAuthFile, err.Error())
> 		return false, nil
> 	}
> 	s.log.Infof("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON)))
> 	return strings.TrimSpace(string(dockerConfigJSON)) == cleanPS, nil
> })
> 
> return nil
> ```
> 
> }`
> 
> On post-pivot operation i will copy this file aside create new one, with relevant pull secret. Then i will wait till api is up (or maybe till mco is actually up) and then copy it back, This seems to make mco happy, after that i will override pull-secret while applying pull-secret manifest. Tested the scenario and it works, WDYT?

Do we really need to go through all that?
won't it just work if you apply the pull secret that match the file on the node once the API is up?

tsorya:
> > @leo8a @eranco74 so i found the way that works and is pretty ok as for me. We should create empty json pull-secret (or some bad data) and not to delete it, wait till mco will apply it on the system and add this file to archive as before, something like `func (s *SeedCreator) cleanupPS() error { s.log.Infof("Creating empty pull secret manifest") cleanPS := "{}" ps := corev1.Secret{ ObjectMeta: metav1.ObjectMeta{ Name: common.PullSecretName, Namespace: common.OpenshiftConfigNamespace, }, Data: map[string][]byte{corev1.DockerConfigJsonKey: []byte(cleanPS)}, Type: corev1.SecretTypeDockerConfigJson, } typeMeta, err := utils.TypeMetaForObject(s.client.Scheme(), &ps) if err != nil { return fmt.Errorf("failed to create typeMetafor pull secret, err: %w", err) } ps.TypeMeta = *typeMeta
> > ```
> > if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
> > 	return runtime.IgnoreAlreadyExists(s.client.Update(context.TODO(), &ps))
> > }); err != nil {
> > 	return fmt.Errorf("failed to create pull-secret in seed cluster: %w", err)
> > }
> > 
> > _ = wait.PollUntilContextCancel(context.TODO(), 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
> > 	s.log.Info("Waiting till mco will override pull secret file")
> > 	dockerConfigJSON, err := os.ReadFile(common.ImageRegistryAuthFile)
> > 	if err != nil {
> > 		s.log.Infof("Failed to read %s with error %s, will retry",
> > 			common.ImageRegistryAuthFile, err.Error())
> > 		return false, nil
> > 	}
> > 	s.log.Infof("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON)))
> > 	return strings.TrimSpace(string(dockerConfigJSON)) == cleanPS, nil
> > })
> > 
> > return nil
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > }`
> > On post-pivot operation i will copy this file aside create new one, with relevant pull secret. Then i will wait till api is up (or maybe till mco is actually up) and then copy it back, This seems to make mco happy, after that i will override pull-secret while applying pull-secret manifest. Tested the scenario and it works, WDYT?
> 
> Do we really need to go through all that? won't it just work if you apply the pull secret that match the file on the node once the API is up?

Nope in this case we finish with 
`NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT   AGE
master   rendered-master-5a7822611b2c9726da2763f84edf8a70   False     True       True       1              0                   0                     1                      6d19h
worker   rendered-worker-58695e86b0362d8f6bc7a337ea102bd8   True      False      False      0              0                   0                     0                      6d19h
`

tsorya:
> Do we understand whether it adds another rollout? If so, lets wait with merging until it's being solved

We already change pull secret now so this logic just changes the way we do it but if there are rollouts then they are already there. Will check it.
I assume we can have another option to make recert change pull secret in machine-config and on FS but i think it will be even more vodoo to those that have no idea on recert

eranco74:
/hold per @tsorya request.
It seems that deleting the PS might result in degraded MCP, @tsorya is trying out different workarounds

leo8a:
I also agree, we shouldn't delete the `/var/lib/kubelet/config.json` file entirely. In a disconnected environment, this file is also used at an early stage (during post-pivot) by `recert_etcd` container to pull that image, if it doesn't exist it will fail.

tsorya:
@leo8a found another option, we can override /etc/machine-config-daemon/currentconfig

tsorya:
It seems like a bug in MCO but deleting pull secret does nothing cause of https://github.com/openshift/machine-config-operator/blob/87268155896563f95fa21defbb271210ec913e8e/lib/resourcemerge/machineconfig.go#L133
so we need to set empty value or some bad ps and wait till MCO updates file system.
Code example
`if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
	return runtime.IgnoreAlreadyExists(s.client.Update(context.TODO(), &ps))
}); err != nil {
	return fmt.Errorf("failed to create pull-secret in seed cluster: %w", err)
}

_ = wait.PollUntilContextCancel(context.TODO(), 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
	s.log.Info("Waiting till mco will override pull secret file")
	dockerConfigJSON, err := os.ReadFile(common.ImageRegistryAuthFile)
	if err != nil {
		s.log.Infof("Failed to read %s with error %s, will retry",
			common.ImageRegistryAuthFile, err.Error())
		return false, nil
	}
	s.log.Infof("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON)))
	return strings.TrimSpace(string(dockerConfigJSON)) == cleanPS, nil
})`

controllers/seedgen_controller.go
```diff
@@ -256,6 +257,40 @@ func (r *SeedGeneratorReconciler) currentAcmCrds(ctx context.Context) (acmCrdLis
...
+		},
+		Data: map[string][]byte{
+			".dockerconfigjson": []byte(common.PullSecretEmptyData),
+		},
+		Type: corev1.SecretTypeDockerConfigJson,
+	}
+
+	// Sanitize cluster's pull-secret from sensitive data
+	if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
+		return client.IgnoreAlreadyExists(r.Client.Update(ctx, sanitizedPullSecret))
```
tsorya, line 274: we are not creating it anymore so maybe no reason for IgnoreAlreadyExists ?

controllers/seedgen_controller.go
```diff
@@ -256,6 +257,40 @@ func (r *SeedGeneratorReconciler) currentAcmCrds(ctx context.Context) (acmCrdLis
...
+	}); err != nil {
+		return fmt.Errorf("failed to update pull-secret in seed cluster: %w", err)
+	}
+
+	_ = wait.PollUntilContextCancel(ctx, 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
```
tsorya, line 279: maybe we should set timeout for it?

main/main.go
```diff
@@ -374,6 +375,32 @@ func initSeedGen(ctx context.Context, c client.Client, log *logr.Logger) error {
...
+	// Restore original pull-secret after seed creation.
+	// During seed creation, the pull-secret was removed; it needs to be restored back
+	// to allow the cluster operators to fully recover in the seed cluster.
+	log.Info("Restore original pull-secret after seed creation")
+	dockerConfigJSON, err := os.ReadFile(common.PathOutsideChroot(utils.StoredPullSecret))
+	if err != nil {
+		return fmt.Errorf("failed to read .dockerconfigjson file: %w", err)
+	}
+
+	pullSecret := &corev1.Secret{
```
tsorya, line 387: Seems to be the same part in both cases just with different data, lets add it to utils? I assume it will be used in post-pivot too in this case. WDYT?

main/main.go
```diff
@@ -374,6 +375,32 @@ func initSeedGen(ctx context.Context, c client.Client, log *logr.Logger) error {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      common.PullSecretName,
+			Namespace: "openshift-config",
```
tsorya, line 390: there is consts for it i think

controllers/seedgen_controller.go
leo8a: agree

main/main.go
leo8a: great idea, imo

controllers/seedgen_controller.go
```diff
@@ -646,6 +679,17 @@ func (r *SeedGeneratorReconciler) generateSeedImage(ctx context.Context, seedgen
 		r.Log.Info(fmt.Sprintf("No hubKubeconfig found in secret %s. Skipping hub interaction", utils.SeedGenSecretName))
 	}
 
+	// Get the cluster's pull-secret
+	clusterPullSecret := &corev1.Secret{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: common.PullSecretName, Namespace: common.OpenshiftConfigNamespace}, clusterPullSecret); err != nil {
```
tsorya, line 684: We have pretty nice function in util to GetSecretData i think , better to use it

utils/utils.go
```diff
@@ -302,3 +302,16 @@ func MoveFileIfExists(source, dest string) error {
 	}
 	return nil
 }
+
+func NewPullSecretFromDockerConfig(dockerConfigJSON []byte) *corev1.Secret {
```
tsorya, line 306: why not to add 
`if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
		return c.Update(ctx, originalPullSecret)
	}); err != nil {
		return fmt.Errorf("failed to restore original pull-secret in seed cluster: %w", err)
	}`
here? It seems common to all places no?

main/main.go
```diff
@@ -374,6 +374,23 @@ func initSeedGen(ctx context.Context, c client.Client, log *logr.Logger) error {
 		return err
 	}
 
+	// Restore original pull-secret after seed creation.
+	// During seed creation, the pull-secret was removed; it needs to be restored back
+	// to allow the cluster operators to fully recover in the seed cluster.
+	log.Info("Restore original pull-secret after seed creation")
+	dockerConfigJSON, err := os.ReadFile(common.PathOutsideChroot(utils.StoredPullSecret))
```
tsorya, line 381: If file doesn't exist, why to fail? Maybe we failed to create it before it? I think we can just skip this part in case utils.StoredPullSecret doesn't exists.

eranco74: How can it be OK?
If you end up without a pull secret it means that something went wrong.

controllers/seedgen_controller.go
```diff
@@ -256,6 +257,33 @@ func (r *SeedGeneratorReconciler) currentAcmCrds(ctx context.Context) (acmCrdLis
...
+	_ = wait.PollUntilContextCancel(deadlineCtx, 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
+		r.Log.Info("Waiting for MCO to override pull-secret file")
+		dockerConfigJSON, err := os.ReadFile(filepath.Join(common.Host, common.ImageRegistryAuthFile))
+		if err != nil {
+			r.Log.Info(fmt.Sprintf("Failed to read %s file with error %s, will retry",
+				common.ImageRegistryAuthFile, err))
+			return false, nil
+		}
+		r.Log.Info(fmt.Sprintf("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON))))
+		return strings.TrimSpace(string(dockerConfigJSON)) == string(sanitizedPullSecret.Data[".dockerconfigjson"]), nil
```
eranco74, line 281: Consider using https://pkg.go.dev/github.com/stretchr/testify/require#JSONEqf instead of comparing strings

eranco74 (commented):
/approve
just a nit

main/main.go
tsorya: Lets say seed creation failed before removing PS, now you will fail to restore it and cleanup phase will get an error.
I assume if file doesn't exists it means seed creator didn't add it and it should be fine to validate if it exists before

controllers/seedgen_controller.go
leo8a: added #292 to align with this function

main/main.go
leo8a: discussed by chat - this PR doesn't remove the pull-secret, it just updates the secret in the cluster and waits for MCO to update it in the filesystem. There should always be a pull-secret file in the node's filesystem, hence we prefer to fail if the operator doesn't find it during this restoration.

PR #249, opened 2024-01-25
Add IBU doc.

browsell:
Co-authored by: Jun Chen <jun@redhat.com

PR #257, opened 2024-01-26
CNF-10559: Configurable automatic rollbacks on upgrade failures

donpenney:
This update introduces automatic rollbacks when a failure occurs during an IBU, as well as adding an init monitor that triggers a rollback should the configurable timeout expire before the upgrade completes.

An automatic rollback can occur in these cases::
- post-reboot config service-unit failure
- Upgrade stage completion failure
- init-monitor timeout before Upgrade completes

Upon completion of the Upgrade stage, LCA will shutdown and disable the init-monitor. If the init-monitor times out prior to this, it will trigger a rollback.

Details on configuration of the automatic rollbacks can be found in HACKING.md


lca-cli/postpivot/postpivot.go
```diff
@@ -117,6 +117,11 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
 		return err
 	}
 
+	// Inject a failure for testing
```
omertuc, line 120: Is this committed on purpose?

lca-cli/ops/execute.go
```diff
@@ -57,6 +65,10 @@ func NewRegularExecutor(logger *logrus.Logger, verbose bool) Execute {
 }
 
 func (e *regularExecutor) Execute(command string, args ...string) (string, error) {
+	if stringslices.Contains(args, "|") {
+		// The command includes a pipe, so run it with bash -c
+		e.executor.executeBash(nil, "", command, args...)
```
omertuc, line 70: This makes me very uneasy from a security perspective. Nothing concrete but shells are injection paradise. If we need a shell, we better be very explicit about it in the calling function

lca-cli/ops/execute.go
```diff
@@ -116,17 +128,10 @@ func NewChrootExecutor(logger *logrus.Logger, verbose bool, root string) Execute
...
-func (e *chrootExecutor) baseExecute(writer io.Writer, command string, args ...string) (string, error) {
-	commandBase := "/usr/bin/env"
-	args = append([]string{command}, args...)
-	arguments := []string{"--", "bash", "-c", strings.Join(args, " ")}
-	return e.executor.execute(writer, e.root, commandBase, arguments...)
-}
-
 func (e *chrootExecutor) Execute(command string, args ...string) (string, error) {
-	return e.baseExecute(nil, command, args...)
+	return e.executor.executeBash(nil, e.root, command, args...)
```
omertuc, line 132: (continuing previous comment) Actually, I didn't realize all of our executions are always through bash. This is not good. We should change that

omertuc: Created #258

api/v1alpha1/types.go
```diff
@@ -76,6 +77,13 @@ type SeedImageRef struct {
 	PullSecretRef *PullSecretRef `json:"pullSecretRef,omitempty"`
 }
 
+type AutoRollbackOnFailure struct {
+	DisabledForPostRebootConfig  bool `json:"disabledForPostRebootConfig,omitempty"`
+	DisabledForUpgradeCompletion bool `json:"disabledForUpgradeCompletion,omitempty"`
+	DisabledInitMonitor          bool `json:"disabledInitMonitor,omitempty"`
+	InitMonitorTimeoutSeconds    int  `json:"initMonitorTimeoutSeconds,omitempty"`
```
omertuc, line 84: Add // for every field to explain what it does?

controllers/prep_handlers.go
```diff
@@ -234,6 +235,10 @@ func (r *ImageBasedUpgradeReconciler) SetupStateroot(ctx context.Context, ibu *l
 		return fmt.Errorf("failed rpm-ostree cleanup -b: %w", err)
 	}
 
+	if err := reboot.WriteIBUAutoRollbackConfigFile(ibu, r.OstreeClient); err != nil {
```
omertuc, line 238: Should we consider including all this configuration in the `SeedReconfiguration` struct? I feel the pre-pivot LCA and post-pivot LCA communications are getting a bit scattered

omertuc: On the other hand SeedReconfiguration is shared with IBI, which doesn't care about rollbacks, so you could consider it to be polluting the struct. 

wdyt?

internal/reboot/reboot.go
```diff
@@ -2,25 +2,159 @@ package reboot
...
+		EnabledComponents:  make(map[string]bool),
+		LcaTestVars:        make(map[string]string),
+	}
+
+	rollbackCfg.EnabledComponents["config"] = !ibu.Spec.AutoRollbackOnFailure.DisabledForPostRebootConfig
+	rollbackCfg.EnabledComponents["postpivot"] = !ibu.Spec.AutoRollbackOnFailure.DisabledForPostRebootConfig
+
+	// Check environ for LCA_TEST_* vars and add them
+	envFileContent := ""
+	re := regexp.MustCompile(`^LCA_TEST_`)
```
omertuc, line 80: `^LCA_TEST_` prefix has multiple appearances, deserves a const

internal/reboot/reboot.go
```diff
@@ -2,25 +2,159 @@ package reboot
...
+	for _, envVar := range os.Environ() {
+		if re.MatchString(envVar) {
+			pair := strings.SplitN(envVar, "=", 2)
+			rollbackCfg.LcaTestVars[pair[0]] = pair[1]
+			envFileContent += envVar + "\n"
+		}
+	}
+
+	if envFileContent != "" {
```
omertuc, line 89: Condition seems unnecessary, let's write the file despite it being empty, for more consistent behavior

controllers/prep_handlers.go
donpenney: Yeah, I was trying to keep out of the IBI area as much as possible, hence the separate config file. In IBI, the file won't be present, so it would be skipped.

internal/reboot/reboot.go
```diff
@@ -2,25 +2,159 @@ package reboot
...
+}
+
+func DisableInitMonitor(log logr.Logger, e ops.Execute) error {
+	if _, err := e.Execute("systemctl", "is-active", common.IBUInitMonitorService); err == nil {
+		if _, err := e.Execute("systemctl", "stop", common.IBUInitMonitorService); err != nil {
+			return fmt.Errorf("failed to stop %s: %w", common.IBUInitMonitorService, err)
+		}
+	}
+
+	if _, err := e.Execute("systemctl", "is-enabled", common.IBUInitMonitorService); err == nil {
```
omertuc, line 135: I think this one as well

lca-cli/initmonitor/initmonitor.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		m.log.Error(err, "failed to read rollback config file")
+		return err
+	}
+
+	if !rollbackCfg.InitMonitorEnabled || rollbackCfg.InitMonitorTimeout <= 0 {
+		return nil
+	}
+
```
omertuc, line 55: A "Will initiate rollback if not stopped by `<timestamp>`" log could be nice here

lca-cli/initmonitor/initmonitor.go
```diff
@@ -0,0 +1,107 @@
...
+		if os.IsNotExist(err) {
+			return nil
+		}
+
+		m.log.Error(err, "failed to read rollback config file")
+		return err
+	}
+
+	if !rollbackCfg.InitMonitorEnabled || rollbackCfg.InitMonitorTimeout <= 0 {
+		return nil
```
omertuc, line 53: Negative timeout should be error IMO

Also a log here explaining why the service will not run could be nice for troubleshooting

lca-cli/initmonitor/initmonitor.go
```diff
@@ -0,0 +1,107 @@
...
+		rpmOstreeClient:      rpmostreeclient.NewClient("initmonitor", hostCommandsExecutor),
+		ostreeClient:         ostreeclient.NewClient(hostCommandsExecutor, false),
+	}
+}
+
+func (m *InitMonitor) RunInitMonitor() error {
+	rollbackCfg, err := reboot.ReadIBUAutoRollbackConfigFile()
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
```
omertuc, line 45: A log here as well, saying it will not run because the config did not exist

donpenney: I avoided that so that it wouldn't generate logs in the IBI scenario

omertuc: Logs are virtually free, if it helps troubleshooting in case the file doesn't exist because <some bug> in IBU, I think it's worth it

lca-cli/ops/execute.go
donpenney: I think originally the bash execs were for the chroot'ed unshare "remount and run command" steps, and maybe ended up as the default exec when the executor code was created?

lca-cli/initmonitor/initmonitor.go
donpenney: If you don't mind the noise in the IBI journal logs, I can add them in, sure.

internal/reboot/reboot.go
```diff
@@ -2,25 +2,159 @@ package reboot
...
+	if rollbackCfg, err := ReadIBUAutoRollbackConfigFile(); err == nil {
+		tag := "LCA_TEST_inject_failure_" + component
+		return (rollbackCfg.LcaTestVars[tag] == "yes")
+	}
+
+	return false
+}
+
+func DisableInitMonitor(log logr.Logger, e ops.Execute) error {
+	if _, err := e.Execute("systemctl", "is-active", common.IBUInitMonitorService); err == nil {
```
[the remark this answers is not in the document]
donpenney: If the service-unit doesn't exist for whatever reason (ie. wasn't in the seed image, manually deleted, or some other reason), then these checks avoid errors trying to disable or stop it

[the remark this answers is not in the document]
omertuc: I see. This is not apparent from the code, so this code is at risk of being removed in the future by someone who is none-the-wiser about which purpose it really serves.

So perhaps some code comment explaining its purpose can prevent it being removed

lca-cli/initmonitor/initmonitor.go
browsell: If the log is not relevant in IBI then should not be generated, just creates confusion

controllers/upgrade_handlers.go
```diff
@@ -211,13 +218,44 @@ func exportForUncontrolledRollback(ibu *lcav1alpha1.ImageBasedUpgrade) error {
...
+
 var getStaterootVarPath = func(stateroot string) string {
 	return common.PathOutsideChroot(filepath.Join(common.GetStaterootPath(stateroot), "/var"))
 }
 
 // CheckHealth helper func to call HealthChecks
 var CheckHealth = healthcheck.HealthChecks
 
+// DisableInitMonitor function pointer, allowing UT to replace it
+var DisableInitMonitor = reboot.DisableInitMonitor
```
tsorya, line 233: Can we make it interface and use it with mock please

lca-cli/ops/execute.go
```diff
@@ -48,6 +49,13 @@ func (e *executor) execute(liveLogger io.Writer, root, command string, args ...s
 	return stdoutBytesTrimmed, nil
 }
 
+func (e *executor) executeBash(writer io.Writer, root, command string, args ...string) (string, error) {
```
tsorya, line 52: Why we need it? Can you use function from ops?

internal/prep/prep.go
```diff
@@ -208,28 +201,14 @@ func SetupStateroot(log logr.Logger, ops ops.Ops, ostreeClient ostreeclient.ICli
...
-	} else {
-		deploymentID, err := rpmOstreeClient.GetDeploymentID(osname)
-		if err != nil {
-			return fmt.Errorf("failed to get deploymentID: %w", err)
-		}
-		deployment, err = getDeploymentFromDeploymentID(deploymentID)
-		if err != nil {
-			return err
-		}
+	deploymentDir, err := ostreeClient.GetDeploymentDir(osname)
```
tsorya, line 204: before pushing we should verify it doesn't break ibi

internal/reboot/reboot.go
tsorya, line 80: Can't we has those envs as params? I mean we should have something like here
https://github.com/openshift/assisted-service/blob/master/cmd/main.go#L145
This way is pretty dirty as for me

internal/reboot/reboot.go
```diff
@@ -2,25 +2,159 @@ package reboot
...
 )
 
+type IBUAutoRollbackConfig struct {
+	InitMonitorEnabled bool              `json:"monitor_enabled,omitempty"`
+	InitMonitorTimeout int               `json:"monitor_timeout,omitempty"`
+	EnabledComponents  map[string]bool   `json:"enabled_components,omitempty"`
+	LcaTestVars        map[string]string `json:"lca_test_vars,omitempty"`
+}
+
+func writeInstallationConfigurationEnvFile(ostreeClient ostreeclient.IClient, stateroot, content string) error {
```
tsorya, line 33: Can we provide parameters as some struct written to file?

lca-cli/installation_configuration_files/scripts/prepare-installation-configuration.sh
```diff
@@ -68,3 +68,9 @@ systemctl disable prepare-installation-configuration.service
 if [[ -n "${DEVICE}" ]]; then
     umount_config "${DEVICE}"
 fi
+
+if [ "${LCA_TEST_inject_failure_config:-no}" = "yes" ]; then
```
tsorya, line 72: All this is done for test? 
In anycase i am going to remove this file in nearest future so not sure this is a good way to do it.
Code will be moved to post-pivot

lca-cli/installation_configuration_files/services/installation-configuration.service
```diff
@@ -6,6 +6,8 @@ After=prepare-installation-configuration.service NetworkManager-wait-online.serv
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/local/bin/lca-cli post-pivot
+ExecStopPost=-/usr/local/bin/lca-cli init-monitor --exec-stop-post postpivot
```
tsorya, line 9: Currently this service is built in a way that it should restart on failure and failure is possible. We can't add this line till post-pivot logic is not changed in case you want to handle roll-back on service failure.
From my personal opinion this logic should be added to post pivot itself. In case there is an error we can verify how much time we are running and run a rollback. Currently this change will break pretty many installations as for example "applying manifest can fail and it is ok" or "pulling image can fail and it is ok" .

tsorya:
I am pretty confused regarding monitor logic to say the truth, we sure should talk about it

lca-cli/ops/execute.go
donpenney: Removing

lca-cli/ops/execute.go
donpenney: We need the `/bin/env` for chroot, but I can get rid of the `bash -c` here, and instead only use it where absolutely necessary. Also adding a comment to explain the need for the `env`

internal/reboot/reboot.go
tsorya: We can provide those but lets do it nice and not to go trough all envs params, just look on link i added

lca-cli/installation_configuration_files/services/installation-configuration.service
tsorya: Talked about it, we can resolve the comment

lca-cli/ops/execute.go
donpenney: Removing the "bash -c by default" from the chroot altogether, leaving it to the caller to use it in specific cases when absolutely necessary.

donpenney:
The IBU end-user doc has been created with the merging of https://github.com/openshift-kni/lifecycle-agent/pull/249 .
I'll update that doc in a separate PR with auto-rollback information.

internal/ostreeclient/ostreeclient.go
```diff
@@ -91,12 +100,53 @@ func (c *Client) SetDefaultDeployment(index int) error {
...
+
+		if len(fields) < 2 {
+			continue
+		}
+
+		if fields[0] == stateroot {
+			// Return the deployment for the first matching stateroot
+			return fields[1], nil
+		}
+	}
```
omertuc, line 139: This parser could be expressed with regex, e.g.

```go
func getDeploymentHash(output, deployment string) (string, error) {
	matches := regexp.MustCompile(fmt.Sprintf(`(?m)^[\s\*]+%s\s+([a-f0-9]{64}.\d+)\s*$`, regexp.QuoteMeta(deployment))).FindStringSubmatch(output)

	if len(matches) > 2 {
		return "", fmt.Errorf("deployment %s found multiple times", deployment)
	} else if len(matches) == 0 {
		return "", fmt.Errorf("deployment %s not found", deployment)
	}

	return matches[1], nil
}
```

I personally find it easier to follow (especially if pasted to tools like regex101.com)

(feel free to ignore if you disagree)

omertuc (changes requested):
Some non-substantial comments

internal/ostreeclient/ostreeclient.go
donpenney: Given that this would only be parsing 4 lines at most, in practice, the regex doesn't buy us much in efficiency. And while I have a very long history of enjoying using regex in my code, there's an equally long history of others complaining about the readability of said regex 😃  I do like the suggested regex, though, so I'll think a bit

internal/common/utils.go
```diff
@@ -41,8 +42,23 @@ import (
 	"github.com/openshift-kni/lifecycle-agent/api/v1alpha1"
 )
 
-// TODO: Need a better way to change this but will require relatively big refactoring
-var OstreeDeployPathPrefix = ""
+var (
+	// TODO: Need a better way to change this but will require relatively big refactoring
+	OstreeDeployPathPrefix = ""
+
+	hostDirExists = true
```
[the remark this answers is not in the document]
donpenney: The intent here is to save doing a system call to check for the presence of /host every time we're determining a path. We should be able to check once at the beginning and cache the result. It's internal to the module, so would only be called once on init, and in this specific case should never change.

internal/common/utils.go
```diff
@@ -75,7 +91,10 @@ func GetConfigMaps(ctx context.Context, c client.Client, configMaps []v1alpha1.C
 
 // PathOutsideChroot returns filepath with host fs
 func PathOutsideChroot(filename string) string {
-	return filepath.Join(Host, filename)
+	if hostDirExists {
```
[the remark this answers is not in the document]
donpenney: It seemed slightly more efficient to just check the bool directly in this case, rather than having the added function call. The accessor function was added in case there was a need to check for /host outside this module

internal/ostreeclient/ostreeclient.go
```diff
@@ -91,12 +100,53 @@ func (c *Client) SetDefaultDeployment(index int) error {
...
+func (c *Client) GetDeployment(stateroot string) (string, error) {
 	args := []string{"admin", "status"}
 	if c.ibi {
-		args = append(args, "--sysroot", "/mnt")
+		args = append(args, "--sysroot", common.OstreeDeployPathPrefix)
+	}
+
+	output, err := c.executor.Execute("ostree", args...)
+	if err != nil {
+		return "", err
```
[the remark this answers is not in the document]
donpenney: missed this went I went back through my code. I specifically looked for "return err" after the thread on this, but didn't cast a wider net

internal/common/utils.go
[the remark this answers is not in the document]
omertuc: I realize and understand all of that, it doesn't counter my argument that it's very confusing to read and reason about it, and it's not worth it for such a cheap operation. Golang code performance is no concern to us, and likely will never be. Our operator does very simple things. When performance becomes a concern, we should do profiling and find out low hanging fruits where we can make dramatic improvements. This one just adds complexity for a performance difference that is likely hard to even measure, so I argue it's not worth it

internal/ostreeclient/ostreeclient.go
omertuc: >the regex doesn't buy us much in efficiency

Yeah this was mostly about simplicity and removing lines of code / logic, not about efficiency. But its admittedly subjective in this case, as you said, some people hate regex

internal/ostreeclient/ostreeclient.go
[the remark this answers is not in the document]
donpenney: Fixed up the unwrapped errors from my PR that I missed

internal/common/utils.go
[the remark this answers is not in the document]
donpenney: Removed the caching

internal/common/utils.go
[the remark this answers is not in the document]
donpenney: Removed the caching

PR #267, opened 2024-01-30
MGMT-16571 Add ib-cli binary to allow users to create IBI installation ISO

eranco74:


omertuc:
Where would the user get the kubeconfig to log into the cluster created from the ISO?

lca-cli/installationiso/installationiso.go
```diff
@@ -0,0 +1,184 @@
...
+func (r *InstallationIso) Create(seedImage, seedVersion, authFile, pullSecretFile, sshPublicKeyPath, lcaImage, rhcosLiveIsoUrl, installationDisk string) error {
+	r.log.Info("Creating IBI installation ISO")
+	err := r.validate()
+	if err != nil {
+		return err
+	}
+	err = r.renderIgnition(seedImage, seedVersion, authFile, pullSecretFile, sshPublicKeyPath, lcaImage, installationDisk)
+	if err != nil {
+		return err
+	}
```
leo8a, line 61: just a nit to your consideration... if so, mind the `err :=` for below conditionals
```suggestion
	if err := r.validate(); err != nil {
		return err
	}
	if err = r.renderIgnition(seedImage, seedVersion, authFile, pullSecretFile, sshPublicKeyPath, lcaImage, installationDisk); err != nil {
		return err
	}
```

eranco74:
> lca-cli is currently an internal binary, do we really want to expose it to users? I think we should have a separate binary for human users
Although it's getting executed by different code flows lca-cli was meant for human users (see the readme and the usage example).
Also, I prefer adding it here so the users can do whatever image-based task they need to with a single tool

lca-cli/installationiso/installationiso.go
eranco74: I think in this case the line gets too long, making it less readable

eranco74:
> Where would the user get the kubeconfig to log into the cluster created from the ISO?

From the image-based-install-operator of course. 
Note that the installation ISO only unpack the seed image to the disk, the cluster configuartion will happen at a later stage, currently it can be achived by the above mentioned operator or https://github.com/rh-ecosystem-edge/ib-orchestrate-vm/blob/master/Makefile.ibi

omertuc:
> > Where would the user get the kubeconfig to log into the cluster created from the ISO?
> 
> From the image-based-install-operator of course. Note that the installation ISO only unpacks the seed image to the disk, the cluster configuration will happen at a later stage. Currently, it can be achieved by the above-mentioned operator or [rh-ecosystem-edge/ib-orchestrate-vm@`master`/Makefile.ibi](https://github.com/rh-ecosystem-edge/ib-orchestrate-vm/blob/master/Makefile.ibi?rgh-link-date=2024-01-30T19%3A58%3A00Z)

Ah I see, OK

omertuc:
> > lca-cli is currently an internal binary, do we really want to expose it to users? I think we should have a separate binary for human users
>
> Although it's getting executed by different code flows lca-cli was meant for human users (see the readme and the usage example).

"lca-cli was meant for human users" - this is wrong to the best of my knowledge.

Before this PR, it wasn't meant for human users. Ever since we have SeedGenerator, there was no need for humans to run this tool.

The reason this README exists was because it was orignally created for the ibu-imager which was meant to be ran manually. ibu-imager was renamed to lca-cli, and we kept the README around, but there was no reason to. We should've deleted the README.

> Also, I prefer adding it here so the users can do whatever image-based task they need to with a single tool

"With a single tool" implies they things they could do with this tool in the first place, which I argue they didn't.

Now, with your new PR, the only manual use for lca-cli is your new PR use-case.

If we have this all be part of lca-cli, then users of lca-cli running `--help` will see into a bunch of weird subcommands they don't understand, have no use for, know nothing about, and will be confused by, because they are not meant for humans.

So this is why I suggest we create a new binary - to have a binary only with subcommands meant for manual human interaction

eranco74:
> Before this PR, it wasn't meant for human users. Ever since we have SeedGenerator, there was no need for humans to run this tool.
OK, I agree

> The reason this README exists was because it was orignally created for the ibu-imager which was meant to be ran manually. ibu-imager was renamed to lca-cli, it became automatic since SeedGeneartor, but we kept the README around, even though there was no reason to keep it. We should've deleted the README.
We can keep the README for documentation, we can add a note saying that this binary is used by the LCA-operator and that the functionality can be achieved with a CR.    

> > Also, I prefer adding it here so the users can do whatever image-based task they need to with a single tool
> 
> "With a single tool" implies there's more things to manually do with this tool, which I argue there aren't.
> Now, with your new PR, the only manual use for lca-cli is your new PR use-case.
True, though there's an ibi command which get used by the install-rhcos-and-restore-seed.sh (not a human but also not the LCA-opeator).
And @donpenney was thinking of adding the precaching subcommand (which can be usefull for a human user) to this binary as well.   

> If we have this all be part of lca-cli, then users of lca-cli running `--help` will see into a bunch of weird subcommands they don't understand, have no use for, know nothing about, and will be confused by, because they are not meant for humans.
> So this is why I suggest we create a new binary - to have a binary only with subcommands meant for manual human interaction
You are making a good point, especially because the current functionality isn't just supported by the SeedGenerator CR, it requires LCA to be installed on the SNO (though I'm not sure whether it's required for generating seed images for IBI, for the sake of alignment and single artifact lets say that it is), so there's no reason to use the lca-cli manually.
I'll bring it up on the slack channel just to hear more thoughts but overall another binary that is user facing sounds alright.

ib-cli/installationiso/installationiso.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if _, err := r.ops.RunInHostNamespace(command, args...); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *InstallationIso) renderButaneConfig(seedImage, seedVersion, authFile, pullSecretFile, sshPublicKeyPath, lcaImage, installationDisk string) error {
+	r.log.Debug("Generating butane config")
+	ignitionTemplate, err := folder.ReadFile(ibiButaneTemplateFilePath)
```
tsorya, line 149: You can use RenderTemplate from utils.go

eranco74: Fixed

PR #270, opened 2024-01-31
CNF-11136: Add handleFinalizeFailure and handleAbortFailure

sudomakeinstall2:
If finalize or abort fails, ibu trasitions into finalizeFailed and
abortFailed states respectively.
```
  status:
    conditions:
    - message: failed to cleanup backups. Perform cleanup manually then
add lca.openshift.io/manualCleanupDone
        annotation to ibu CR to transition back to Idle
      observedGeneration: 5
      reason: AbortFailed
      status: "False"
      type: Idle
```

User should perform cleanup manually in these states. After manual
cleanup user should add 'lca.openshift.io/manualCleanupDone' annotation
to IBU CR.

```
  kind: ImageBasedUpgrade
  metadata:
    annotations:
      lca.openshit.io/manualCleanupDone: ""

```
Upon observing this annotation IBU will remove the annotations and runs
abort/finalize. If successful ibu transitions back to idle state.

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -237,20 +237,18 @@ func (r *ImageBasedUpgradeReconciler) handleStage(ctx context.Context, ibu *lcav
 
 func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (nextReconcile ctrl.Result, err error) {
 	idleCondition := meta.FindStatusCondition(ibu.Status.Conditions, string(utils.ConditionTypes.Idle))
-	if idleCondition != nil && idleCondition.Status == metav1.ConditionFalse {
-		switch idleCondition.Reason {
-		case string(utils.ConditionReasons.Aborting):
-			nextReconcile, err = r.handleAbort(ctx, ibu)
-		case string(utils.ConditionReasons.AbortFailed):
-			nextReconcile, err = r.handleAbortFailure(ctx, ibu)
-		case string(utils.ConditionReasons.Finalizing):
-			nextReconcile, err = r.handleFinalize(ctx, ibu)
-		case string(utils.ConditionReasons.FinalizeFailed):
-			nextReconcile, err = r.handleFinalizeFailure(ctx, ibu)
-		}
-		if nextReconcile.Requeue == false {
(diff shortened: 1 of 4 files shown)


controllers/ibu_controller.go
```diff
@@ -242,11 +242,11 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 		case string(utils.ConditionReasons.Aborting):
 			nextReconcile, err = r.handleAbort(ctx, ibu)
 		case string(utils.ConditionReasons.AbortFailed):
-			nextReconcile, err = r.handleAbortFailure(ctx, ibu)
+			nextReconcile, err = r.handleIdleFailure(ctx, ibu)
 		case string(utils.ConditionReasons.Finalizing):
 			nextReconcile, err = r.handleFinalize(ctx, ibu)
```
Missxiaoguo, line 247: Should we now return immediate requeue in handleFinalize/handleAbort for failed cleanup case? otherwise I think the IBU status will be reset by the next `ResetStatusConditions` func.

sudomakeinstall2: Done. Also changed to condition to reset status to only when reason is idle

sudomakeinstall2:
/hold
adding verification after manual cleanup

controllers/ibu_controller.go
```diff
@@ -248,7 +248,7 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
 		case string(utils.ConditionReasons.FinalizeFailed):
 			nextReconcile, err = r.handleFinalizeFailure(ctx, ibu)
 		}
-		if nextReconcile.Requeue == false {
+		if nextReconcile.Requeue == false && idleCondition.Reason == string(utils.ConditionReasons.Idle) {
```
jc-rh, line 251: With the new condition, the old one can be removed? Then we don't have to requeue for failure handling? Annotation would be an edit that will trigger reconcile. Originally we were thinking about some sort of flag file which would require polling.

sudomakeinstall2: I am not sure what you mean by new and old conditions?

Adding annotation does not trigger reconcile on ibu.

sudomakeinstall2: removed the condition and add trigger to reconcile upon adding or removing this annotation

PR #271, opened 2024-01-31
Copy the kubeadmin password from original SNO

omertuc:
Solves [MGMT-15731](https://issues.redhat.com//browse/MGMT-15731)

Users of OCP expect the default kubeadmin password generated during
installation time to keep working even after an IBU.

To achieve that, LCA will have to copy the password hash from the
kube-system/kubeadmin secret in the original SNO into the
`SeedReconfiguration` struct so it can be set in the equivalent secret
in the new SNO.

Note that not all clusters that we will be upgrading will have such
password, as OCP encourages people, for security reasons, to disable
this password by deleting the secret as outlined here:
https://docs.openshift.com/container-platform/4.14/authentication/remove-kubeadmin.html

In case no such secret exists, LCA will ensure that the secret will
simply be deleted in the new SNO.

The secret hash replacement will be actually done using a new recert
feature (which is activated by simply setting the hash in the recert
config, or setting it to an empty string to signal to recert to delete
the secret). The reason for doing it using recert is that recert is
already modifying etcd secrets while the cluster is offline, and we
don't want to launch a cluster that might still have some old
compromised seed password, even for just a little while.

An alternative solution to doing it in recert would be to have the seed
generator set a dummy password hash that no one in the world knows the
true password of. We arbitrarily chose the recert solution instead.

Note that if the seed cluster for some reason doesn't have the secret,
and the original SNO does have the secret, then it will cause an error,
as recert only supports modifying an existing secret, not creating one
from scratch. We could modify recert to theoretically create the secret,
but there's not much point to do it - we expect all seed clusters to
still have the secret. Another thing worth noting that we can't simply
in this case create the secret after the cluster has already started,
because the password gets rejected by OCP if the creation timestamp of
the secret differs from the kube-system namespace's creation timestamp
by more than 1 hour.

### DIFF
--- a/internal/clusterconfig/clusterconfig_test.go
+++ b/internal/clusterconfig/clusterconfig_test.go
@@ -18,6 +18,7 @@ package clusterconfig
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -147,6 +148,14 @@ var (
 			InfrastructureName: "mysno-xsb4m",
 		},
 	}
+
+	kubeadminSecret = &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubeadmin",
+			Namespace: "kube-system",
+		},
+		Data: map[string][]byte{"kubeadmin": []byte(`$2a$10$20Q4iRLy7cWZkjn/D07bF.RZQZonKwstyRGH0qiYbYRkx5Pe4Ztyi`)},
+	}
 )
 
 func init() {
@@ -213,16 +222,17 @@ func TestClusterConfig(t *testing.T) {
 		},
 	}
 	testcases := []struct {
-		testCaseName   string
-		pullSecret     client.Object
-		caBundleCM     client.Object
-		clusterVersion client.Object
-		idms           client.Object
-		icsps          []client.Object
-		node           client.Object
-		proxy          client.Object
-		expectedErr    bool
-		validateFunc   func(t *testing.T, tempDir string, err error, ucc UpgradeClusterConfigGather)
+		testCaseName    string
+		pullSecret      client.Object
+		caBundleCM      client.Object
+		clusterVersion  client.Object
+		idms            client.Object
+		icsps           []client.Object
+		node            client.Object
+		proxy           client.Object
+		deleteKubeadmin bool
+		expectedErr     bool
+		validateFunc    func(t *testing.T, tempDir string, err error, ucc UpgradeClusterConfigGather)
 	}{
 		{
 			testCaseName:   "Validate success flow",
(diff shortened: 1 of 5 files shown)


internal/clusterconfig/clusterconfig_test.go
```diff
@@ -266,6 +274,7 @@ func TestClusterConfig(t *testing.T) {
 				if err := utils.ReadYamlOrJSONFile(filepath.Join(clusterConfigPath, common.SeedClusterInfoFileName), seedReconfig); err != nil {
 					t.Errorf("unexpected error: %v", err)
 				}
+				assert.Equal(t, "$2a$10$20Q4iRLy7cWZkjn/D07bF.RZQZonKwstyRGH0qiYbYRkx5Pe4Ztyi", seedReconfig.KubeadminPasswordHash)
```
eranco74, line 277: Add a test for kubeadmin secret not found

internal/recert/recert.go
```diff
@@ -20,43 +20,50 @@ const (
...
+	ExtendExpiration bool   `json:"extend_expiration,omitempty"`
+	ForceExpire      bool   `json:"force_expire,omitempty"`
+	EtcdEndpoint     string `json:"etcd_endpoint,omitempty"`
+	ClusterRename    string `json:"cluster_rename,omitempty"`
+	Hostname         string `json:"hostname,omitempty"`
+	// We intentionally don't omitEmpty this field because an empty string here
+	// means "delete the kubeadmin password secret" while a complete omission
+	// of the field means "don't touch the secret". We never want the latter,
+	// we either want to delete the secret or update it, never leave it as is.
+	KubeadminPasswordHash string   `json:"kubeadmin_password_hash"`
```
eranco74, line 33: If `don't touch the secret` isn't a valid option, you can interpret a complete omission of the filed as `delete the kubeadmin secret`.

internal/recert/recert.go
```diff
@@ -20,43 +20,50 @@ const (
...
+	CNSanReplaceRules     []string `json:"cn_san_replace_rules,omitempty"`
+	UseKeyRules           []string `json:"use_key_rules,omitempty"`
+	UseCertRules          []string `json:"use_cert_rules,omitempty"`
 }
 
 // CreateRecertConfigFile function to create recert config file
 // those params will be provided to an installation script after reboot
 // that will run recert command with them
-func CreateRecertConfigFile(clusterInfo *seedreconfig.SeedReconfiguration, seedClusterInfo *seedclusterinfo.SeedClusterInfo, cryptoDir, recertConfigFolder string) error {
+func CreateRecertConfigFile(seedReconfig *seedreconfig.SeedReconfiguration, seedClusterInfo *seedclusterinfo.SeedClusterInfo, cryptoDir, recertConfigFolder string) error {
```
eranco74, line 46: :+1:

internal/recert/recert.go
omertuc: Recert runs in more contexts other than just seed reconfiguration (e.g. seed expiration)

In those contexts, I don't want recert to do anything with this secret.

So recert supports:

- Don't touch the secret (omission of field)
- Modify the secret (field set to non-empty value)
- Delete the secret (field set to empty value)

omertuc:
Behavior changed, secret will be forcefully deleted, but still generally compatible

/unlabel cluster-config-api-changed

internal/clusterconfig/clusterconfig_test.go
omertuc: Created #287 before adding test

omertuc:
/unlabel cluster-config-api-changed

Behavior changed, secret will be forcefully deleted, but still generally compatible

PR #272, opened 2024-01-31
Add some context to reconciliation error messages

omertuc:
This helps track down the source of errors. Long term we're considering
having a linter for this


internal/extramanifest/extramanifest.go
```diff
@@ -131,26 +131,27 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
...
 
 	for i, policy := range policies {
 		objects, err := getConfigurationObjects(policy, objectLabels)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get configuration objects: %w", err)
 		}
 		for _, object := range objects {
 			fileName := fmt.Sprintf("%d_%s_%s.yaml", i, object.GetName(), object.GetNamespace())
-			filePath := filepath.Join(toDir, PolicyManifestPath, fileName)
```
leo8a, line 150: maybe removing `toDir` here accidentally?

omertuc: So, to answer your question, no.

But I'll revise the name

PR #284, opened 2024-02-05
Add recert IP arg to skip control-plane revision triggers

omertuc:
Similar to b6c5db194b205dd86e2227b630907b92f597b724 but for IP instead of hostname

This change leverages recert's latest OCP post-process IP feature, which makes OCP's control-plane cluster operators (i.e. etcd, kube-apiserver, authentication, etc) happy, so that they won't trigger additional revisions. Thus reducing the time OCP needs to stabilize after recert.

For more details please check here:
- https://github.com/rh-ecosystem-edge/recert/pull/89

### DIFF
--- a/internal/recert/recert.go
+++ b/internal/recert/recert.go
@@ -26,6 +26,7 @@ type RecertConfig struct {
 	EtcdEndpoint     string `json:"etcd_endpoint,omitempty"`
 	ClusterRename    string `json:"cluster_rename,omitempty"`
 	Hostname         string `json:"hostname,omitempty"`
+	IP               string `json:"ip,omitempty"`
 	// We intentionally don't omitEmpty this field because an empty string here
 	// means "delete the kubeadmin password secret" while a complete omission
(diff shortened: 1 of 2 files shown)


mresvanis:
Since recert replaces the previous IP with the current on, we should verify whether we still need [this](https://github.com/openshift-kni/lifecycle-agent/blob/main/lca-cli/postpivot/postpivot.go#L225-L239) code.

omertuc:
> Since recert replaces the previous IP with the current one, we should verify whether we still need [this](https://github.com/openshift-kni/lifecycle-agent/blob/main/lca-cli/postpivot/postpivot.go?rgh-link-date=2024-02-06T09%3A13%3A10Z#L225-L239) code.

https://github.com/rh-ecosystem-edge/recert/blob/88cb8c4811451b2e4b420dd62812bbe0e3c69701/src/ocp_postprocess/ip_rename.rs#L40-L46

eranco74 (commented):
/lgtm
/approve

PR #285, opened 2024-02-06
CNF-11184: remove seed image even if it is not mounted

sudomakeinstall2:
Remove seed image even if it is not mounted as part of `UnmountAndRemoveImage`

### DIFF
--- a/lca-cli/ops/ops.go
+++ b/lca-cli/ops/ops.go
@@ -253,19 +253,7 @@ func (o *ops) IsImageMounted(imgName string) (bool, error) {
 	return false, nil
 }
 
-func (o *ops) UnmountAndRemoveImage(img string) error {
-	mounted, err := o.IsImageMounted(img)
-	if err != nil {
-		return fmt.Errorf("failed to check if image is mounted: %w", err)
-	}
-	if !mounted {
-		return nil
-	}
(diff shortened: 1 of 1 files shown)


lca-cli/ops/ops.go
```diff
@@ -281,6 +269,22 @@ func (o *ops) UnmountAndRemoveImage(img string) error {
...
+	}
+	if !mounted {
+		return o.removeImage(img)
+	}
+	if _, err := o.hostCommandsExecutor.Execute(
+		"podman", "image", "umount", img,
+	); err != nil {
+		return fmt.Errorf("failed to unmount image: %w", err)
+	}
+	return o.removeImage(img)
```
omertuc, line 285: Would this be equivalent?

```suggestion
	mounted, err := o.IsImageMounted(img)
	if err != nil {
		return fmt.Errorf("failed to check if image is mounted: %w", err)
	}

	if mounted {
		if _, err := o.hostCommandsExecutor.Execute(
			"podman", "image", "umount", img,
		); err != nil {
			return fmt.Errorf("failed to unmount image: %w", err)
		}
	}

	return o.removeImage(img)
```

omertuc (changes requested):

jc-rh (approved):

PR #289, opened 2024-02-06
CNF-11185: Remove undeployed stateroots in finalize

sudomakeinstall2:
If a deployment fails there could be a stateroot directory under `/ostree/deploy`. This stateroot will not be listed in rpm-ostree. This commit will remove such stateroots.

### DIFF
--- a/controllers/idle_handlers.go
+++ b/controllers/idle_handlers.go
@@ -31,6 +31,8 @@ import (
 )
 
 var osStat = os.Stat
+var osReadDir = os.ReadDir
+var osRemoveAll = os.RemoveAll
 
 //nolint:unparam
 func (r *ImageBasedUpgradeReconciler) handleAbort(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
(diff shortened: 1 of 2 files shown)


omertuc:
What does "deployment fails" mean in this context? 

Could this directory being leftover be considered an ostree bug? If so, did we file one, so we can eventually remove this workaround to that bug?

sudomakeinstall2:
> What does "deployment fails" mean in this context?
> 
> Could this directory being leftover be considered an ostree bug? If so, did we file one, so we can eventually remove this workaround to that bug?

I am not sure since I haven't faced this issue myself.
@browsell

donpenney (approved):

donpenney:
> What does "deployment fails" mean in this context?
> 
> Could this directory being leftover be considered an ostree bug? If so, did we file one, so we can eventually remove this workaround to that bug?

@omertuc @sudomakeinstall2 @browsell This would generally mean there was some issue with the seed image that led to a failure creating the deployment. The existing cleanup code assumed that all stateroots have a deployment, but that isn't the case.

omertuc:
> > What does "deployment fails" mean in this context?
> > Could this directory being leftover be considered an ostree bug? If so, did we file one, so we can eventually remove this workaround to that bug?
> 
> @omertuc @sudomakeinstall2 @browsell This would generally mean there was some issue with the seed image that led to a failure creating the deployment. The existing cleanup code assumed that all stateroots have a deployment, but that isn't the case.

So again:

>led to a failure creating the deployment

What does that mean? Isn't "creating a deployment" an atomic operation that ostree is supposed to take care of end-to-end, and if it fails doing that, why doesn't ostree delete that directory?

donpenney:
> > > What does "deployment fails" mean in this context?
> > > Could this directory being leftover be considered an ostree bug? If so, did we file one, so we can eventually remove this workaround to that bug?
> > 
> > 
> > @omertuc @sudomakeinstall2 @browsell This would generally mean there was some issue with the seed image that led to a failure creating the deployment. The existing cleanup code assumed that all stateroots have a deployment, but that isn't the case.
> 
> So again:
> 
> > led to a failure creating the deployment
> 
> What does that mean? Isn't "creating a deployment" an atomic operation that ostree is supposed to take care of end-to-end, and if it fails doing that, why doesn't ostree delete that directory?

Creating the stateroot is a separate step.

PR #290, opened 2024-02-06
MGMT-16780: remove node deletion from seed image creation

mresvanis:
This change leaves the seed node object intact in the seed image, as [recert is going to delete it](https://github.com/rh-ecosystem-edge/recert/blob/v0/src/ocp_postprocess.rs#L399) as part of its OCP post-process flow. 

This resolves an etcd rollout, which is triggered due to the `etcd-all-certs` secret that is [emptied](https://github.com/openshift/cluster-etcd-operator/blob/551193bb82d1310bf75b04bd9727a18ad61051fc/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go#L178-L217) by the cluster etcd operator when the node is deleted via the k8s API. Since his secret is empty in the seed image, recert cannot adjust it on the target SNO and thus make the cluster etcd operator happy, so a new etcd rollout is triggered. 

By deleting the node object via recert and not via the k8s API, once kubelet is up and doesn't find a node object it re-registers the node. Since recert has already replaced all hostname and IP references in etcd and in the filesystem, including the `etcd-all-certs` secert, the etcd operator is kept happy and doesn't trigger a new rollout.

> No additional recert changes are needed, as recert:{v0,latest} already [deletes](https://github.com/rh-ecosystem-edge/recert/blob/v0/src/ocp_postprocess.rs#L399) the node objects from etcd.

### DIFF
--- a/lca-cli/seedcreator/seedcreator.go
+++ b/lca-cli/seedcreator/seedcreator.go
@@ -16,7 +16,6 @@ import (
 	cp "github.com/otiai10/copy"
 	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	runtime "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


lca-cli/seedcreator/seedcreator.go
```diff
@@ -108,10 +108,6 @@ func (s *SeedCreator) CreateSeedImage() error {
 		s.log.Info("Seed cluster certificates backed up successfully for recert tool")
 	}
 
-	if err := utils.RunOnce("delete_node", common.BackupChecksDir, s.log, s.deleteNode, ctx); err != nil {
-		return err
-	}
-
 	_ = utils.RunOnce("wait_for_ovn_to_go_down", common.BackupChecksDir, s.log, s.waitTillOvnKubeNodeIsDown, ctx)
```
tsorya, line 111: We should delete this one too and lets delete deleteNode function itself and waitTillOvnKubeNodeIsDown

eranco74: But we tested without deleting this line.
Are we sure it will still work?

mresvanis: So far we have been testing with all the [changes](https://github.com/openshift-kni/lifecycle-agent/pull/290/files) you currently see in this PR.

PR #291, opened 2024-02-07
enable gosec and new DEVELOPING.md

pixelsoccupied:


internal/precache/progress.go
```diff
@@ -61,7 +61,7 @@ func (p *Progress) Log() {
 
 func (p *Progress) Persist(filename string) {
 	data, _ := json.Marshal(p)
-	if err := os.WriteFile(filename, data, 0o700); err != nil {
+	if err := os.WriteFile(filename, data, 0o700); err != nil { //nolint:gosec
```
pixelsoccupied, line 64: @sakhoury suppressing it for now...but do you think we can make this `0o600`? I can make this change now if you would like.

DEVELOPING.md
```diff
@@ -0,0 +1,43 @@
...
+
+- Or run the following to allow `golangci-lint` for any auto fixes (if available by linter)
+  
+  ```shell
+  golangci-lint run --fix --config .golangci.yml
+  ```
+
+### Suppress `golangci-lint` errors
+
+If you would like to suppress a false positive, consider in-line syntax.
```
donpenney, line 23:
```suggestion
If you would like to suppress a false positive, consider in-line syntax. Please note that such suppressions should only be used in exceptional cases, however, rather than common practice.
```

DEVELOPING.md
```diff
@@ -0,0 +1,43 @@
...
+
+```shell
+lca-cli/ops/ops.go:162:17: G107: Potential HTTP request made with variable url (gosec)
+                        resp, err := http.Get(healthzEndpoint)
+```
+
+This can ignore with the following `//nolint:mylinter1,mylinter2`
+
+```go
+// ops.go L162 - suppress only gosec
```
donpenney, line 35: You're not suggesting a comment like this should be added, correct? Seems a little misleading to have this comment here

internal/precache/progress.go
donpenney: Why would it be executable? This is just persisting the progress file, right? Please make the change and run through an IBU

DEVELOPING.md
pixelsoccupied: Removed this comment from the code block

DEVELOPING.md
```diff
@@ -0,0 +1,41 @@
+# Development guides and tips  
+
+## Linting
+
+### Run `golangci-lint`
+
+Full list linters and config can be found in `.golangci.yml`
+
+- Consider running the linter locally before commit.
```
donpenney, line 9: Actually, I would suggest instead changing this to indicate a more stronger recommendation/requirement:

Developers are expected to run `make ci-job` prior to posting PRs. This will run the various linters, as well as required unit tests. The golangci-lint test can be run individually at any time during development, as well, with the `make golangci-lint` job to scan code changes.

pixelsoccupied:
/hold
running e2e with the updated perm

DEVELOPING.md
```diff
@@ -0,0 +1,51 @@
...
+  golangci-lint run --fix --config .golangci.yml
+  ```
+
+### Suppress `golangci-lint` errors
+
+If you would like to suppress a false positive, consider in-line syntax. Please note that such suppressions should only be used in exceptional cases, however, rather than common practice.
+
+An example error for gosec
+
+```shell
```
donpenney, line 37: very minor nit: Should be console, rather than shell

internal/precache/progress.go
pixelsoccupied: updated to 0o600

DEVELOPING.md
pixelsoccupied: updated

pixelsoccupied:
/unhold
precache completes with the updated file permission

donpenney (approved):

PR #292, opened 2024-02-07
Refactor code to use utils.GetSecretData

leo8a:


controllers/seedgen_controller.go
```diff
@@ -699,18 +699,18 @@ func (r *SeedGeneratorReconciler) generateSeedImage(ctx context.Context, seedgen
 
 func (r *SeedGeneratorReconciler) restoreManagedCluster(ctx context.Context, clusterName string) error {
 	// Get the seedgen secret
-	seedGenSecret := &corev1.Secret{}
-	if err := r.Client.Get(ctx, types.NamespacedName{Name: utils.SeedGenSecretName, Namespace: common.LcaNamespace}, seedGenSecret); err != nil {
+	hubKubeconfig, err := lcautils.GetSecretData(ctx, utils.SeedGenSecretName, common.LcaNamespace, "hubKubeconfig", r.Client)
```
tsorya, line 702: I don't think you can do it here this way
GetSecretData returns error in case key doesn't exist and in this case it is seems to be ok that key is not there

donpenney: Agreed. I think you can leave this as-is. Support for the hubKubeconfig will likely be getting removed anyway

leo8a: Ok

leo8a:
Rebased code

PR #294, opened 2024-02-08
Add recert image to containers.list only if not exists

leo8a:
Removes the duplication of the recert image from the `containers.list` in seed images. As a result, it also reduces an extra operation while pre-cachingin IBU.

The recert image was already added to the node (and hence included in the `containers.list`) before by the seedgen CR.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -13,6 +13,7 @@ import (
 	"regexp"
 	"text/template"
 
+	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -323,3 +323,12 @@ func UpdatePullSecretFromDockerConfig(ctx context.Context, c client.Client, dock
 
 	return newPullSecret, nil
 }
+
+func AppendToListIfNotExists(list []string, value string) []string {
+	for _, v := range list {
+		if v == value {
+			return list
+		}
+	}
```
omertuc, line 332:
```suggestion
	if lo.Contains(list, value) {
		return list
	}
```

omertuc (changes requested):
Small chnage

utils/utils.go
leo8a: nice

PR #300, opened 2024-02-11
MGMT-16652: make configuration label part of seed reconfiguration api package as consts

tsorya:
[MGMT-16652](https://issues.redhat.com//browse/MGMT-16652): make configuration label part of seed reconfiguration api package as consts


eranco74 (commented):
/lgtm

PR #302, opened 2024-02-11
MGMT-16764: bug fix - removing shorthand from precache params

tsorya:
Shorthand should have only one ascii param but in this case it doesn't worth it and 2 letters fails to run


eranco74 (commented):
/lgtm
/approve

PR #305, opened 2024-02-12
Seed restoration will also restore seed password hash

omertuc:
(draft because untested yet)

Resolves [MGMT-16801](https://issues.redhat.com//browse/MGMT-16801)

We don't want the seed cluster's kube admin password to be valid in the seed image, so we must overwrite its hash with a random junk hash (we had this, except it was not random, just a constant broken invalid hash, which was good enough, but a valid hash of a random unknown password is more "right")

We also want to undo this operation, so that the seed cluster remains usable after being turned into a seed image.

This commit does both

* Extra unrelated change - rename mentions of seed cleanup from "cleanup" to "restoration" which is more accurate


utils/crypto_dir.go
```diff
@@ -115,5 +116,40 @@ func BackupKubeconfigCrypto(ctx context.Context, client runtimeclient.Client, cr
...
 		return err
 	}
+
 	return nil
 }
+
+func BackupKubeadminPasswordHash(ctx context.Context, client runtimeclient.Client, cryptoDir string) (bool, error) {
+	password, err := GetSecretData(ctx, "kubeadmin", "kube-system", "kubeadmin", client)
+	if err != nil {
+		if runtimeclient.IgnoreNotFound(err) != nil {
```
eranco74, line 126: A "not found" error (err == nil) is a valid case and should return nil error
```suggestion
		if runtimeclient.IgnoreNotFound(err) == nil {
```

eranco74 (commented):
looks good

utils/crypto_dir.go
omertuc: Thanks

eranco74:
Tried testing this PR
After generating the seed image the password isn't restored on the seed 
```
# make seed-image-create SEED_IMAGE=$SEED_IMAGE SEED_VERSION=$SEED_VERSION BACKUP_SECRET=$BACKUP_SECRET
Triggering seed image creation
secret/seedgen configured
seedgenerator.lca.openshift.io/seedimage created
Waiting for seed image to be completed
seedgenerator.lca.openshift.io/seedimage condition met
[root@rdu-infra-edge-12 ib-orchestrate-vm]# oc get secret -n kube-system  kubeadmin -o jsonpath='{.data.kubeadmin}'
Error from server (NotFound): secrets "kubeadmin" not found
```
The upgrade fail:
```
Feb 14 17:47:06 recipient recert[1805]: 2024-02-14 17:47:06 - INFO - src/ocp_postprocess.rs:105: setting kubeadmin password hash
Feb 14 17:47:06 recipient recert[1805]: Error: finalizing
Feb 14 17:47:06 recipient recert[1805]: 
Feb 14 17:47:06 recipient recert[1805]: Caused by:
Feb 14 17:47:06 recipient recert[1805]:     0: performing ocp specific post-processing
Feb 14 17:47:06 recipient recert[1805]:     1: setting kubeadmin password hash
Feb 14 17:47:06 recipient recert[1805]:     2: couldn't find Secret/kube-system:kubeadmin
```

PR #307, opened 2024-02-12
Improve pull-secret sanitization

leo8a:


controllers/seedgen_controller.go
```diff
@@ -553,6 +547,13 @@ func (r *SeedGeneratorReconciler) validateSystem(ctx context.Context) (msg strin
 		return
 	}
 
+	// Ensure cluster's pull-secret is not sanitized
+	dockerConfigJSON, _ := os.ReadFile(filepath.Join(common.Host, common.ImageRegistryAuthFile))
+	if strings.TrimSpace(string(dockerConfigJSON)) == strings.TrimSpace(common.PullSecretEmptyData) {
+		msg = "Rejected due to cluster's pull-secret is sanitized (i.e., empty credentials)"
```
donpenney, line 553:
```suggestion
		msg = "Rejected due to invalid cluster pull-secret (previously sanitized without proper restore)"
```

donpenney: Actually, I think in some cases, the empty pull secret is going to mean the health checks will fail, due to the current config in the OADP deployment (it could be stuck trying to pull images, causing out health checks to fail). Given that this is checking the file directly and does not rely on the kube-apiserver, I think we could have this as a separate check done ahead of the health check in the Reconcile function, so that we could catch this particular scenario early with this clearer failure.

This could be tested by deleting the saved pull-secret (better to rename it, in order to easily restore it) while the imager is running. When LCA is restored, it won't restore the pull-secret, and the OADP CSV (I believe) will get stuck in a non-Successful state

donpenney: Thinking about it more, you can leave this check here. I think I need to reorganize a little actually. I believe we have some changes coming to how we run the health checks, so I can rearrange some things then.

PR #288, opened 2024-02-06
enable wrapcheck linter

pixelsoccupied:
wrapcheck linter is now enabled to enforce additional err logs for easy debugging

this builds on https://github.com/openshift-kni/lifecycle-agent/pull/272

PR #310, opened 2024-02-12
enable wrapcheck

pixelsoccupied:
This should fix all the issues seen in https://github.com/openshift-kni/lifecycle-agent/pull/288 

Manually tested seedgen + all happy path stages 

/cc @donpenney

### DIFF
--- a/internal/precache/workload/pullImages.go
+++ b/internal/precache/workload/pullImages.go
@@ -62,8 +62,10 @@ func podmanImgPull(image, authFile string) error {
 	if authFile != "" {
 		args = append(args, []string{"--authfile", authFile}...)
 	}
-	_, err := Executor.ExecuteWithLiveLogger("podman", args...)
-	return err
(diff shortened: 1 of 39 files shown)


internal/precache/workload/pullImages.go
```diff
@@ -98,8 +100,7 @@ func GetAuthFile() (string, error) {
 
 	// Check if authFile exists
 	if _, err := os.Stat(authFile); os.IsNotExist(err) {
-		log.Errorf("Missing auth file for podman")
-		return "", err
+		return "", fmt.Errorf("failed to get authfile for podman: %w", err)
```
browsell, line 103: Missing nil check

pixelsoccupied: L102 actually doing the init statement ...so in this case `os.IsNotExist(err)` doing the check

pixelsoccupied:
/hold

will do another full run after the rebase

pixelsoccupied:
/unhold

sanity check looks good

donpenney (approved):

PR #317, opened 2024-02-15
CNF-11143: Health check enhancements and associated reconciler changes

donpenney:
This update includes changes to the LCA system health checks and how they are handled by the Upgrade and SeedGen reconcilers.

The health checks have been updated to remove the existing polling of background jobs to perform the health checks, replacing this with performing each check in turn when HealthCheck is called from the Upgrade and SeedGen reconcilers. Rather than a blocking call, the reconcilers run the checks and requeue on failure. This provides multiple benefits:
- If the health checks are continuing to fail post-reboot in an upgrade, the user can now trigger a Rollback at any time, rather than waiting for the checks to time out. The reconciler is no longer blocked waiting for the set of health checks to pass, which means it is able to immediately process the Rollback request.
- The IBU and SeedGen CRs are now updated with progress messages to indicate a) that the health checks are in-progress and b) which checks are failing.

The SeedGen reconciler has been updated to allow for the health checks to run and requeue on failure. In addition, the progress status reporting has been improved to provide better information to the user in case certain steps take more time, rather than simply seeing "In progress".

/cc @browsell @jc-rh


donpenney:
Updated for minor rebase to pick up latest merge

controllers/seedgen_controller.go
```diff
@@ -78,6 +78,18 @@ var (
 
 const (
 	EnvSkipRecert = "SEEDGEN_SKIP_RECERT"
+
+	// SeedGen reconciler phases
+	phaseInitial    = "initial"    // SeedGen hasn't started yet
+	phaseGenerating = "generating" // SeedGen is in the first phase of work, ending with the launch of the imager
+	phaseFinalizing = "finalizing" // SeedGen has previously launched the imager and is in the final phase of seed generation
+	phaseCompleted  = "completed"  // SeedGen has successfully completed
+	phaseFailed     = "failed"     // SeedGen has failed
```
pixelsoccupied, line 87: could make the them typed and include them in the seedgen apis for consistency? e.g https://github.com/openshift-kni/lifecycle-agent/blob/main/api/v1alpha1/types.go#L45-L46

controllers/seedgen_controller.go
```diff
@@ -861,56 +921,65 @@ func (r *SeedGeneratorReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return
 	}
 
-	if isSeedGenFailed(seedgen) {
+	phase := getPhase(seedgen)
+
+	if phase == phaseFailed {
```
pixelsoccupied, line 926: probably can benefit from switch statement clarity and performance since we have 4 or more cases?

controllers/upgrade_handlers.go
```diff
@@ -273,44 +273,58 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedU
 	u.Log.Info("Starting health check for different components")
 	err := CheckHealth(u.Client, u.Log)
 	if err != nil {
-		utils.SetUpgradeStatusFailed(ibu, err.Error())
-		u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to health check failure: %s", err))
-		return doNotRequeue(), nil
+		utils.SetUpgradeStatusInProgress(ibu, fmt.Sprintf("System health checks in progress: %s", err.Error()))
+		return requeueWithShortInterval(), nil
```
pixelsoccupied, line 277: if the there's an err...it means health is in progress. Maybe a bool or typed string can be used to indicate in progress would be better?

controllers/upgrade_handlers.go
```diff
@@ -319,10 +333,12 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedU
 			u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to restore failure: %s", err))
 			return doNotRequeue(), nil
 		}
+		utils.SetUpgradeStatusInProgress(ibu, fmt.Sprintf("Restoring OADP Data: Failure occurred: %s", err))
```
pixelsoccupied, line 336: glad we are making use of the status msg more. but i bet we (i) will forget to update the status right before returning with an error. Is there something we can do like in prep where updating the msg is sort of handled implicitly? (dont have to update this now...just a thought)

controllers/seedgen_controller.go
donpenney: I can certainly make it typed, but this is an internal state that won't be exposed in the API file

controllers/seedgen_controller.go
donpenney: Will do... That's what I wanted to do, but was looking up the wrong go function, and ended up just doing the if-checks :)

controllers/upgrade_handlers.go
donpenney: Yes, if any of the health checks fail for whatever reason, I'd consider that an overall health check failure, which would in turn mean the health checks are "in progress" and will retry once requeued.

controllers/upgrade_handlers.go
donpenney: The Prep handler status updates need work, as well, as they generally lag behind what's actually happening, sometimes quite a bit, but at least part of that would be due to the structure of the Prep handler where tasks are run in the background. That's not the case in the Upgrade handler, where the status can be updated directly as needed.

controllers/upgrade_handlers.go
```diff
@@ -273,44 +273,58 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedU
...
+			u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to failure applying config manifests: %s", err))
 			return doNotRequeue(), nil
 		}
-		return requeueWithError(fmt.Errorf("error while applying extra manifests: %w", err))
+		utils.SetUpgradeStatusInProgress(ibu, fmt.Sprintf("Applying Config Manifests: Failure occurred: %s", err.Error()))
+		return requeueWithError(fmt.Errorf("error while applying config manifests: %w", err))
 	}
 
 	// Recovering OADP configuration
+	utils.SetUpgradeStatusInProgress(ibu, "Restoring OADP Configuration")
```
browsell, line 310: nit, can we change the message to "Restoring Application Configuration"

controllers/upgrade_handlers.go
pixelsoccupied: Yeah i guess in this case there's no possibility of random errors like "missing CRDs" and any get/list api call fails would be implicitly retried this way. Alright sounds good.

controllers/seedgen_controller.go
pixelsoccupied: Ah yeah just typed work in that case

jc-rh (approved):

controllers/ibu_controller.go
```diff
@@ -194,26 +204,40 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-	} else {
-		inProgressStage := utils.GetInProgressStage(ibu)
-		if inProgressStage != "" {
-			nextReconcile, err = r.handleStage(ctx, ibu, inProgressStage)
-			if err != nil {
-				_ = utils.UpdateIBUStatus(ctx, r.Client, ibu)
-				return
+			ibu.Status.ValidNextStages = getValidNextStageList(ibu, isAfterPivot)
+			// Update status
+			if updateErr := utils.UpdateIBUStatus(ctx, r.Client, ibu); updateErr != nil {
```
jc-rh, line 216: Not sure if continuing reconciliation is the best option when there is an update err

controllers/ibu_controller.go
```diff
@@ -183,6 +190,9 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return
 	}
 
+	// Make sure "next stage" list is initialized
+	ibu.Status.ValidNextStages = getValidNextStageList(ibu, isAfterPivot)
```
jc-rh, line 194: This case should be covered by line 234?

controllers/ibu_controller.go
```diff
@@ -194,26 +204,40 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 					return
 				}
 				if !isValid {
-					err = utils.UpdateIBUStatus(ctx, r.Client, ibu)
+					ibu.Status.ValidNextStages = getValidNextStageList(ibu, isAfterPivot)
```
jc-rh, line 207: I can't think of a case where invalid transition would change the next stages.

controllers/ibu_controller.go
```diff
@@ -194,26 +204,40 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			}
+			return
 		}
 	}
 
+	ibu.Status.ValidNextStages = getValidNextStageList(ibu, isAfterPivot)
+
 	// Update status
-	err = utils.UpdateIBUStatus(ctx, r.Client, ibu)
+	if updateErr := utils.UpdateIBUStatus(ctx, r.Client, ibu); updateErr != nil {
```
jc-rh, line 237: I am more concerned with this one, comparing to the update error in the middle. Are we sure that it's good idea to hide the error from the controller? Normally this is entitled for another reconcile.

controllers/ibu_controller.go
donpenney: I was originally just ignoring the interim status update failures, but figured it would be better to log them at least. I think all the DB interactions should be reviewed/revised, as the current Update potentially masks issues.

PR #323, opened 2024-02-16
CNF-11236: Apply extra manifests in the order specified by the `apply-wave` annotation

sudomakeinstall2:
Sort, group by and apply extra manifests with respect to `apply-wave`
annotation

Refactor sort, group by and load grouped manifests to use generics to
remove code duplications

Signed-off-by: Saeid Askari <saskari@redhat.com>

### DIFF
--- a/internal/backuprestore/common.go
+++ b/internal/backuprestore/common.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math"
 	"os"
 	"path/filepath"
 
@@ -57,11 +56,9 @@ import (
 // +kubebuilder:rbac:groups=oadp.openshift.io,resources=dataprotectionapplications,verbs=get;list;create;update;watch
 
 const (
-	applyWaveAnn     = "lca.openshift.io/apply-wave"
-	applyLabelAnn    = "lca.openshift.io/apply-label"
-	backupLabel      = "lca.openshift.io/backup"
-	clusterIDLabel   = "config.openshift.io/clusterID" // label for backups applied by lifecycle agent
-	defaultApplyWave = math.MaxInt32                   // 2147483647, an enough large number
+	applyLabelAnn  = "lca.openshift.io/apply-label"
+	backupLabel    = "lca.openshift.io/backup"
+	clusterIDLabel = "config.openshift.io/clusterID" // label for backups applied by lifecycle agent
 
 	OadpPath        = "/opt/OADP"
 	OadpRestorePath = OadpPath + "/veleroRestore"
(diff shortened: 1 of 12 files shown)


internal/extramanifest/extramanifest.go
```diff
@@ -264,24 +265,30 @@ func (h *EMHandler) ExportExtraManifestToDir(ctx context.Context, extraManifestC
...
-	for i, manifest := range manifests {
-		gvk := manifest.GroupVersionKind()
-
-		fileName := fmt.Sprintf("%d_%s_%s_%s.yaml", i, gvk.Kind, manifest.GetName(), manifest.GetNamespace())
-		filePath := filepath.Join(toDir, ExtraManifestPath, fileName)
-		err = utils.MarshalToYamlFile(manifest, filePath)
-		if err != nil {
-			return fmt.Errorf("failed to marshal manifest %s to yaml: %w", manifest.GetName(), err)
+		for j, manifest := range manifestGroup {
+			fileName := fmt.Sprintf("%d_%s_%s.yaml", j+1, manifest.GetName(), manifest.GetNamespace())
```
Missxiaoguo, line 285: gvk.Kind is removed from filename. Is it intentional?

internal/backuprestore/common.go
```diff
@@ -399,11 +393,19 @@ func (h *BRHandler) ValidateOadpConfigmaps(ctx context.Context, content []lcav1a
...
-	restores, err := h.extractRestoreFromConfigmaps(ctx, configmaps)
+
+	for _, backup := range backups {
+		err := h.createBackupWithDryRun(ctx, backup)
+		if err != nil {
+			return err
+		}
+	}
+
+	restores, err := common.ExtractResourcesFromConfigmaps[*velerov1.Restore](ctx, configmaps, common.RestoreGvk)
```
Missxiaoguo, line 408: Missing dryrun for restore CRs?

internal/common/manifest_helpers.go
```diff
@@ -0,0 +1,97 @@
...
+			decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewBufferString(value), 4096)
+			for {
+				resource := unstructured.Unstructured{}
+				err := decoder.Decode(&resource)
+				if err != nil {
+					if errors.Is(err, io.EOF) {
+						// Reach the end of the data, exit the loop
+						break
+					}
+					return nil, fmt.Errorf("")
```
Missxiaoguo, line 81: Should raise error with a message? i.e., failed to decode yaml

Missxiaoguo (commented):
Thanks for the refactoring! Nice work!

internal/backuprestore/common.go
sudomakeinstall2: updated

internal/extramanifest/extramanifest.go
Missxiaoguo: Probably not a problem now as it's sorted file starting with number. It was added to ensure unique filename. But I still prefer to add it which helps to know the resource type by just looking at the name. The linux file length limit is long enough, so we don't need to worry about.

sudomakeinstall2: updated

PR #326, opened 2024-02-16
CNF-11228: Add extra manifests configmap validation

Missxiaoguo:
- validations for extra manifests configmap
   - fail the prep stage if the configmap is not present
   - fail the prep stage if any syntax issues detected when parsing/decoding manifest yamls
   - fail the prep stage if create/update with dryrun failed due to invalid fields and webhook errors
   - continue with prep stage if create/update with dryrun failed due to missing CRD, a warning message is added in the prep status condition for user to verify before moving to upgrade stage

- move all IBU spec validation to the prep handler
- move dynamicclient creation to utils
- pass the existing dynamicclient to EMhandler at the initialization stage and use the restmapper provided from the default client
- change the exported manifest file name with resource group_version_kind_name_namespace.yaml to ensure identical name

/cc @browsell @donpenney @sudomakeinstall2 @jc-rh


leo8a:
This is looking great!

Can we also have an update on the docs (where appropriate/needed)? Or this is planned in a follow-up PR?

sudomakeinstall2 (commented):
nit: other log messages in this repo does not use `.` at the end. Maybe we should keep it consistent.

Missxiaoguo:
> This is looking great!
> 
> Can we also have an update on the docs (where appropriate/needed)? Or this is planned in a follow-up PR?

Sure! Will have a follow-up PR for the docs update

internal/extramanifest/extramanifest.go
```diff
@@ -81,52 +100,180 @@ func IsEMFailedError(err error) bool {
...
-	configmaps, err := common.GetConfigMaps(ctx, h.Client, extraManifestCMs)
+func IsEMUnknownCRDError(err error) bool {
+	var emErr *EMStatusError
+	if errors.As(err, &emErr) {
+		return emErr.Reason == "UnknownCRD"
+	}
+	return false
+}
+
+func (h *EMHandler) applyExtraManifest(ctx context.Context, manifest *unstructured.Unstructured, isRryRun bool) error {
```
sudomakeinstall2, line 119: nit: type isRyRun

Missxiaoguo: ah

internal/extramanifest/extramanifest.go
```diff
@@ -81,52 +100,180 @@ func IsEMFailedError(err error) bool {
...
+
+		manifest.SetResourceVersion(existingManifest.GetResourceVersion())
+		if _, err := resource.Update(ctx, manifest, opts); err != nil {
+			return fmt.Errorf("failed to update manifest %s called %s: %w", manifest.GetKind(), manifest.GetName(), err)
+		}
+	}
+	return nil
+}
+
+func (h *EMHandler) ValidateExtraManifestConfigmap(ctx context.Context, content []lcav1alpha1.ConfigMapRef) error {
```
jc-rh, line 162: Nit, ValidateExtraManifestConfigmap**s**

internal/extramanifest/extramanifest.go
```diff
@@ -81,52 +100,180 @@ func IsEMFailedError(err error) bool {
...
+	// Create the directory for the extra manifests
+	exMDirPath := filepath.Join(toDir, ExtraManifestPath)
+	if err := os.MkdirAll(exMDirPath, 0o700); err != nil {
+		return fmt.Errorf("failed to create directory for extra manifests in %s: %w", exMDirPath, err)
+	}
+
+	for i, manifest := range manifests {
+		gvk := manifest.GroupVersionKind()
+
+		fileName := fmt.Sprintf("%d_%s_%s_%s_%s_%s.yaml", i, gvk.Group, gvk.Version, gvk.Kind, manifest.GetName(), manifest.GetNamespace())
```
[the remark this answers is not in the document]
Missxiaoguo: Yes, the filename length limit is 255 bytes.  Kustomize generated files are in this format, so I assume it should be okay. I also tested all manifests from DU profile, and everything is working fine. 
But I guess there is still have a risk of exceeding the limit. Perhaps I will just pre-append the Kind for now and update if neccessary later on.

PR #332, opened 2024-02-19
MGMT-16968: Remove old rendered machine-configs from seed

tsorya:
Removing machine-configs on seedcreation that are not used by mcp


omertuc:
Does this serve any purpose?

tsorya:
> Does this serve any purpose?

actually yes, it brings up cleaner cluster without old machine configs. This cleanups old pull secrets for example from them (though recert will override them at the end)

PR #339, opened 2024-02-21
[doc] Extends apply-label documentation and removes secret from klusterlet backup example

alosadagrande:
Hi,

I found not clear the documentation around the apply-label. Concretely the one related to filtering resources, did not know that even the same resources or other specifically added in the spec are not backed up if they are not included in the apply-label annotation.

The other commit focuses on removing a pull-secret from open-cluster-management-agent ns that I do not see on my installations and causes the klusterlet backup example to fail.

cc/ @Missxiaoguo @leo8a

### DIFF
--- a/controllers/seedgen_controller.go
+++ b/controllers/seedgen_controller.go
@@ -339,7 +339,14 @@ func (r *SeedGeneratorReconciler) waitForPullSecretOverride(ctx context.Context,
 			return false, nil
 		}
 		r.Log.Info(fmt.Sprintf("%s data is %s", common.ImageRegistryAuthFile, strings.TrimSpace(string(dockerConfigJSON))))
-		return strings.TrimSpace(string(dockerConfigJSON)) == string(updatedPullSecret.Data[".dockerconfigjson"]), nil
+		if strings.TrimSpace(string(dockerConfigJSON)) != string(updatedPullSecret.Data[".dockerconfigjson"]) {
+			return false, nil
+		}
+		if err := healthcheck.AreMachineConfigPoolsReady(deadlineCtx, r.NoncachedClient, r.Log); err != nil {
+			r.Log.Info(fmt.Sprintf("Waiting for MCP: %s", err.Error()))
+			return false, nil
+		}
+		return true, nil
 	})
 	if err != nil {
 		return fmt.Errorf("timed out waiting for MCO to override pull-secret file: %w", err)


docs/backuprestore-with-oadp.md
```diff
@@ -334,14 +336,14 @@ W2RlZmF1bHRdCmF3c19hY2Nlc3Nfa2V5X2lkPTxBV1NfQUNDRVNTX0tFWV9JRD4KYXdzX3NlY3JldF9h
...
+    lca.openshift.io/apply-label: "apps/v1/deployments/open-cluster-management-agent/klusterlet,v1/secrets/open-cluster-management-agent/bootstrap-hub-kubeconfig,rbac.authorization.k8s.io/v1/clusterroles/klusterlet,v1/serviceaccounts/open-cluster-management-agent/klusterlet,rbac.authorization.k8s.io/v1/clusterroles/open-cluster-management:klusterlet-admin-aggregate-clusterrole,rbac.authorization.k8s.io/v1/clusterrolebindings/klusterlet,operator.open-cluster-management.io/v1/klusterlets/klusterlet,apiextensions.k8s.io/v1/customresourcedefinitions/klusterlets.operator.open-cluster-management.io"
```
donpenney, line 346: We need to make the distinction here that this `v1/secrets/open-cluster-management-agent/open-cluster-management-image-pull-credentials` is present on systems where the multiclusterhub CR has `.spec.imagePullSecret`. I don't think we can drop it completely

docs/backuprestore-with-oadp.md
```diff
@@ -64,7 +64,7 @@ spec:
    - deployments
 ```
 
-By adding this annotation, LCA will limit the scope exclusively to those resources. LCA will parse the annotations and apply `lca.openshift.io/backup: <backup-name>` label to those resources. LCA then adds this labelSelector when creating the backup CR:
+By adding this annotation, LCA will limit the scope exclusively to those resources. This means that no other resources, even if they are specifically added in the spec field, will be backed up. However, notice that in order to use the apply-label annotation to back up specific objects, the resource name must be included in the backup. LCA will parse the annotations and apply `lca.openshift.io/backup: <backup-name>` label to those resources. LCA then adds this labelSelector when creating the backup CR:
```
Missxiaoguo, line 70: Suggest to add this note in line74.

>[!IMPORTANT]
> Please note that to use the wave-apply annotation for backing up specific resources, the resources listed in the annotation should also be properly included in the spec. Additionally, if the wave-apply annotation is used in the backup CR, only the resources listed in the annotation will be backed up, regardless of whether other resource types are specified in the spec or not.

Missxiaoguo: /cc @sudomakeinstall2

sudomakeinstall2: makes sense to me.

PR #348, opened 2024-02-22
OCPBUGS-29788: Introduce optional SriovNetworkNodeState health check

donpenney:
If the sriov-network-operator is not ready to process configuration updates, it can cause upgrade failures when the post-reboot Upgrade handler attempts to apply extra-manifests with SRIOV config data. While there is a CSV health check that verifies the operator is installed and running, it does not ensure the necessary resources have been created.

The SriovNetworkNodeState CR is created by the sriov-network-operator when its reconcilers first run. If the CR has a Succeeded state, it indicates the operator should be ready to process the configuration updates from the extra-manifests.

However, there is a race condition after rebooting to the upgraded release, where sriov-network-operator can be launched and run its reconcilers before kube-apiserver is able to service its requests. As a result, the first time its reconcilers run, it may not see its config data at all, and does not create the necessary CRs until the next time the reconciler runs, 5 minutes later.

As a workaround, the upgrade handler checks the results of a failed health check, adding an annotation to the default SriovNetworkNodePolicy CR if the failure indicates the SRIOV resources have not yet been created, kicking the corresponding reconciler to run immediately, rather than wait for the periodic requeue, which could be minutes later.


controllers/upgrade_handlers.go
```diff
@@ -268,14 +271,80 @@ func (u *UpgHandler) autoRollbackIfEnabled(ibu *lcav1alpha1.ImageBasedUpgrade, m
...
+	// creating the CRs, which triggers the workaround to give the reconciler a little kick before requeuing. The health
+	// check should then be fine on the next pass of the IBU reconciler, as there will have been enough time for the CRs
+	// to have been created.
+	//
+
+	if !strings.Contains(msg, healthcheck.SriovNetworkNodeStateNotPresentMsg) {
+		return nil
+	}
+
+	policies := &sriovv1.SriovNetworkNodePolicyList{}
```
pixelsoccupied, line 313: You could consider leveraging  `Unstructured` instead of typed. Seems like too much import just for the def? (though I'm team typed)

https://github.com/kubernetes/client-go/tree/master/examples/dynamic-create-update-delete-deployment
https://github.com/openshift-kni/lifecycle-agent/blob/main/internal/extramanifest/extramanifest.go#L212-L215

controllers/upgrade_handlers.go
```diff
@@ -268,14 +271,80 @@ func (u *UpgHandler) autoRollbackIfEnabled(ibu *lcav1alpha1.ImageBasedUpgrade, m
...
+	// If the health check failure message includes the message from the SriovNetworkNodeState health check that there
+	// were no SriovNetworkNodeState CRs, it is an indication that the sriov-network-operator reconciler ran without
+	// creating the CRs, which triggers the workaround to give the reconciler a little kick before requeuing. The health
+	// check should then be fine on the next pass of the IBU reconciler, as there will have been enough time for the CRs
+	// to have been created.
+	//
+
+	if !strings.Contains(msg, healthcheck.SriovNetworkNodeStateNotPresentMsg) {
+		return nil
+	}
```
pixelsoccupied, line 311: Would be good use for custom err? we can just do something like 
```
if errors.Is(err, SriovWorkaround) {
```

controllers/upgrade_handlers.go
```diff
@@ -268,14 +271,80 @@ func (u *UpgHandler) autoRollbackIfEnabled(ibu *lcav1alpha1.ImageBasedUpgrade, m
...
 //
 // Note: All decisions, including reconciles and failures, should be made within this function.
 // The caller will simply return what this function returns.
 func (u *UpgHandler) PostPivot(ctx context.Context, ibu *lcav1alpha1.ImageBasedUpgrade) (ctrl.Result, error) {
 	u.Log.Info("Starting health check for different components")
-	err := CheckHealth(u.NoncachedClient, u.Log)
+	err := CheckHealth(ctx, u.NoncachedClient, u.Log)
 	if err != nil {
+		// Trigger sriov workaround, if necessary
+		u.sriovNetworkNodeStateWorkaround(ctx, err.Error())
```
pixelsoccupied, line 346: At this point we can confirm the error and just launch a go routine to shave a tiny bit of time?

pixelsoccupied (commented):
Just some tiny comments but i like modifying the annotation way to trigger the loop. In the end when moving to Idle do we do a cleanup of the annotations? 

LGTM otherwise

internal/healthcheck/healthcheck.go
```diff
@@ -233,3 +255,34 @@ func getNodeStatusCondition(conditions []corev1.NodeCondition, conditionType cor
...
+	}
+
+	nodeStates := &sriovv1.SriovNetworkNodeStateList{}
+
+	if err := c.List(ctx, nodeStates); err != nil {
+		return fmt.Errorf("failed to get SriovNetworkNodeStates: %w", err)
+	}
+
+	if len(nodeStates.Items) == 0 {
+		return fmt.Errorf(SriovNetworkNodeStateNotPresentMsg)
```
pixelsoccupied, line 277: Worth logging that we are doing the special workaround in case devs (I) fall for this (i had crd but no resources!)

controllers/upgrade_handlers.go
donpenney: The error returned by the HealthCheck function is an aggregate, with an error message generated using a list of error messages returned by individual checks, in order to report a summary of the errors, essentially

controllers/upgrade_handlers.go
donpenney: Why?

controllers/upgrade_handlers.go
donpenney: But we'd want the type for the health check. While we could likely use unstructured here in the workaround, we're accessing the status field. So we'd still be importing the same API. Unless you're suggesting there's a way we can drop typed queries for the health checks and still access the status fields?

donpenney:
> Just some tiny comments but i like modifying the annotation way to trigger the loop. In the end when moving to Idle do we do a cleanup of the annotations?
> 
> LGTM otherwise

I suppose we could look for the annotation and remove it, but it doesn't seem necessary.

controllers/upgrade_handlers.go
pixelsoccupied: Yep exactly you could do something like `unstructured.NestedMap(u.UnstructuredContent(), "status")` to key out just the status or even the full with `u.UnstructuredContent()`...they both return map[string]interface{} so it's a bit awkward to use. 

Either way (typed vs Unstructured) it's good with me though so up to you

donpenney: Could be an interesting exercise, maybe we can reduce imports overall, as long as the resulting code doesn't end up ugly/unmaintainable

internal/healthcheck/healthcheck.go
```diff
@@ -233,3 +255,34 @@ func getNodeStatusCondition(conditions []corev1.NodeCondition, conditionType cor
...
+	}
+
+	for _, nodeState := range nodeStates.Items {
+		if nodeState.Status.SyncStatus != "Succeeded" {
+			return fmt.Errorf("sriovNodeNetworkState not ready")
```
Missxiaoguo, line 282: nit: sriovNetworkNodeState

PR #350, opened 2024-02-23
OCPBUGS-29892: seedgen: Add MCP health check to pull-secret sanitization

donpenney:
The pull-secret sanitization function waits until the file has been updated by the MCO. This update also adds a call to the MCP health check to ensure MCO has completed handling the rendered machine-config before continuing with seed generation.


leo8a:
/lgtm

thanks for adding this!

PR #353, opened 2024-02-26
Move OADP secrets and DPA restoration to installation-configuration service

Missxiaoguo:
[CNF-11310](https://issues.redhat.com//browse/CNF-11310)
It takes around 30s for OADP to establish connection with the storage backends after recovering DPA. By relocating the DPA restore to an eariler stage within post-pivot installation-configuration service, the storage backends should be ready when initiating the application restore, so improving the overall efficiency.

[CNF-11312](https://issues.redhat.com//browse/CNF-11312)
Add upgrade status message for pre-pivot

/cc @browsell @jc-rh @donpenney


Missxiaoguo:
Rebased

PR #355, opened 2024-02-26
MGMT-16687: IBI use separate partition for /var/lib/containers

javipolo:
Set as draft until https://github.com/openshift/assisted-service/pull/6005 or https://github.com/openshift/assisted-installer/pull/787 get to an openshift release

### DIFF
--- a/ib-cli/README.md
+++ b/ib-cli/README.md
@@ -82,9 +82,16 @@ export AUTH_FILE=/path/to/seed-image-pull-secret.json
 export PS_FILE=/path/to/release-pull-secret.json
 export SSH_PUBLIC_KEY=~/.ssh/id_rsa.pub
 export IBI_INSTALLATION_DISK=/dev/sda
+# Start of the /var/lib/containers partition. Free space before it will be allocated to system partition
+# It can be one of the following:
+#     - Positive number: partition will start at position 120Gb of the disk and extend to the end of the disk. Example: 120Gb
+#     - Negative number: partition will be of that precise size. Example: -40Gb
+#     - If the flag is ommited, no new partition will be created, and a directory /sysroot/containers will be created and used instead
+export EXTRA_PARTITION_START=-40G
 
 
 ib-cli create-iso --installation-disk ${IBI_INSTALLATION_DISK} \
+  --extra-partition-start ${EXTRA_PARTITION_START} \
   --lca-image ${LCA_IMAGE} \
   --seed-image ${SEED_IMAGE} \
   --seed-version ${SEED_VERSION} \
--- a/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
+++ b/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
@@ -6,27 +6,42 @@ seed_image=${1:-$SEED_IMAGE}
 seed_version=${2:-$SEED_VERSION}
 installation_disk=${3:-$INSTALLATION_DISK}
 lca_image=${4:-$LCA_IMAGE}
+extra_partition_start=${5:-$EXTRA_PARTITION_START}
+extra_partition_number=5
+extra_partition_label=varlibcontainers
+create_extra_partition=true
+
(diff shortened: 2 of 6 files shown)


ib-cli/README.md
```diff
@@ -82,9 +82,14 @@ export AUTH_FILE=/path/to/seed-image-pull-secret.json
 export PS_FILE=/path/to/release-pull-secret.json
 export SSH_PUBLIC_KEY=~/.ssh/id_rsa.pub
 export IBI_INSTALLATION_DISK=/dev/sda
+# Start point of the /var/lib/containers partition. Free space before it will be allocated to system partition
+# It can be specified in a positive manner: 120G , meaning that new partition will start at 120Gb of disk, and extend to the end of the disk
```
eranco74, line 86: While it's nice do users care whether the extra partition is at the end or at the start of the disk?
If not perhaps asking for the partition size is more simple

ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
```diff
@@ -6,28 +6,26 @@ seed_image=${1:-$SEED_IMAGE}
...
+extra_partition_label=varlibcontainers
 
 authfile=${AUTH_FILE:-"/var/tmp/backup-secret.json"}
 pull_secret=${PULL_SECRET_FILE:-"/var/tmp/pull-secret.json"}
 
 coreos-installer install ${installation_disk}
 
+# Create new partition for /var/lib/containers
+sfdisk ${installation_disk} <<< write
+sgdisk --new 5:$extra_partition_start --change-name 5:$extra_partition_label ${installation_disk}
```
eranco74, line 19: Why there's `5` everywhere?

ib-cli/README.md
javipolo: I like the "give me the size" approach, but it differs from the way you define it for the seed

While creating the seed cluster, you specify the start of the partition. and AFAIK you cannot say "I want a partition of 5Gb at the end of the disk"

In IBI case, since we use sgdisk, we can just provide a negative number as the start of the partition, and it counts backwards since the end of the disk, so effectively is the size of the partition

My prefered approach is to just specify the desired size of the partition, and prepend it by - , but I thought better to give context to the user, since partitioning in the end "is not part" of LCA and lands more on user side ...

ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
javipolo: There are 4 standard partitions for RHCOS, we are creating a new one, and so it will be the 5th one. Thus the 5

eranco74: got it, create a var and use it just to make the code clear
e.g.:
`PRECACHCING_PARTITION_NUMBER=5`

javipolo: done :)

eranco74 (commented):
/lgtm

ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
```diff
@@ -6,27 +6,42 @@ seed_image=${1:-$SEED_IMAGE}
 seed_version=${2:-$SEED_VERSION}
 installation_disk=${3:-$INSTALLATION_DISK}
 lca_image=${4:-$LCA_IMAGE}
+extra_partition_start=${5:-$EXTRA_PARTITION_START}
```
eranco74, line 9: Why there's a 5 here? Should be "use directory" no?

javipolo: It's the 5th argument from the command line in case you call the install-rhcos-and-restore-seed.sh script :sweat_smile: 

In the systemd unit we feed the data through env variables, but for testing it is useful :)

eranco74 (commented):
/lgtm

PR #360, opened 2024-02-27
OCPBUGS-29483: Ensure initMonitorTimeoutSeconds positive values only

leo8a:
If the user provides a negative value for initMonitorTimeoutSeconds, it will be rejected when validating the IBU spec.


donpenney:
Please leave 0 as "use default".

PR #362, opened 2024-02-27
CNF-11187: Reconcile when precache job updates

pixelsoccupied:
Currently when a precache job is launched, reconcile is requested after an arbitrary amount of time. During the reconcile wait time the system could be in any state and we can not let the user know about it immediately if an event occurs such as when the job is done.

This is fixed by taking ownership of precache job using IBU cr (OwnerReferences). Now reconcile will be triggered as soon as job updates and system is checked to see if desired state has been reached to complete prep

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openshift-kni/lifecycle-agent/internal/backuprestore"
 	"github.com/openshift-kni/lifecycle-agent/internal/extramanifest"
 	"github.com/openshift-kni/lifecycle-agent/internal/reboot"
+	kbatch "k8s.io/api/batch/v1"
 
 	"github.com/go-logr/logr"
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
@@ -75,11 +76,12 @@ type ImageBasedUpgradeReconciler struct {
 
 // Task contains objects for executing a group of serial tasks asynchronously
 type Task struct {
-	Active   bool
-	Success  bool
-	Cancel   context.CancelFunc
-	Progress string
-	done     chan struct{}
+	Active             bool
+	Success            bool
+	Cancel             context.CancelFunc
+	Progress           string
+	AdditionalComplete string // additional completion msg
+	done               chan struct{}
 }
 
 // Reset Re-initialize the Task variables to initial values
@@ -88,11 +90,13 @@ func (c *Task) Reset() {
 	c.Success = false
 	c.Cancel = nil
 	c.Progress = ""
+	c.AdditionalComplete = ""
 	select {
 	case _, open := <-c.done:
 		if open {
 			close(c.done)
 		}
+	case <-time.After(30 * time.Second): // max wait timeout in case c.done is still empty
 	}
 }
 
@@ -138,16 +142,11 @@ func requeueWithCustomInterval(interval time.Duration) ctrl.Result {
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheusrules,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=batch,resources=jobs/status,verbs=get
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the ImageBasedUpgrade object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
 func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (nextReconcile ctrl.Result, err error) {
 	if r.Mux != nil {
 		r.Mux.Lock()
@@ -506,6 +505,7 @@ func (r *ImageBasedUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return false
 			},
 		})).
+		Owns(&kbatch.Job{}). // note: job resource watched is restricted further using cache.Options during NewManager
 		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
 		Complete(r)
 }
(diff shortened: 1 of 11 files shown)


pixelsoccupied:
/hold 
looking into reduce the scope of watch for better cache management

pixelsoccupied:
/unhold 
scope reduced with cache.Options [feature](https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/cache_options.md#cache-options)

controllers/prep_handlers.go
```diff
@@ -393,24 +379,9 @@ func (r *ImageBasedUpgradeReconciler) prepStageWorker(ctx context.Context, ibu *
 			}
 			r.Log.Info("Successfully created precaching job")
 			r.PrepTask.Progress = "Successfully created precaching job"
+			r.Precache.PrecacheRequested = true
```
Missxiaoguo, line 382: I wonder if `r.Precahe.PrecacheRequested` is really needed? since if it reaches the end, it means the job was created successfully. If any error is raised during precache job creation, the prep will fail.

controllers/prep_handlers.go
```diff
@@ -442,37 +413,86 @@ func (r *ImageBasedUpgradeReconciler) handlePrep(ctx context.Context, ibu *lcav1
...
+				close(r.PrepTask.done)
+				return
+			}
+			if r.Precache.PrecacheRequested {
+				// this indicates we can now safely check for precache job status during reconcile
+				r.PrepTask.Progress = "Prep in progress, watching precache job"
+				r.Log.Info("Precache requested")
 			} else {
-				r.Log.Info("Prep stage completed successfully!")
+				r.Log.Info("No precache requested")
```
Missxiaoguo, line 442: If `prepStageWorker` func is executed successfully, a precache job must have been created. If not, the prep will just fail. It seems we don't have a "No precache requested" case, should we remove it? It looks like r.Preache.PreacheRequested adds the complexity.

pixelsoccupied:
/hold 
making some code review changes

controllers/prep_handlers.go
pixelsoccupied: Yeah! So it once the go routine is launched we may reconcile before the job is created. So it was just make sure we can safely check the job status only once the job is launched. 

But you're right...this is pretty ugly. Will refactor this

internal/precache/precache.go
```diff
@@ -193,13 +193,12 @@ func (h *PHandler) QueryJobStatus(ctx context.Context) (*Status, error) {
 
 // Cleanup deletes the ConfigMap and Job precaching resources
 func (h *PHandler) Cleanup(ctx context.Context) error {
```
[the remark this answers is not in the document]
pixelsoccupied: > it's no longer applicable

This is mostly used for just logging and nothing to do with this PR/ checking precache status actually. I think think we can keep it as it is.

internal/precache/precache.go
```diff
@@ -193,13 +193,12 @@ func (h *PHandler) QueryJobStatus(ctx context.Context) (*Status, error) {
...
-	// Delete Job
-	if err := deleteJob(ctx, h.Client, LcaPrecacheJobName, common.LcaNamespace); err != nil {
+	if err := deleteJob(ctx, h.Client); err != nil {
```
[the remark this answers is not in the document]
pixelsoccupied: Updated comment

controllers/prep_handlers.go
pixelsoccupied: Dropped `PreacheRequested`

controllers/prep_handlers.go
pixelsoccupied: Dropped PreacheRequested. Should be a lot cleaner now

internal/precache/helper.go
```diff
@@ -243,6 +246,12 @@ func renderJob(config *Config, log logr.Logger) (*batchv1.Job, error) {
 		},
 	}
 
+	// set reference
+	log.Info("Set ibu as the owner")
```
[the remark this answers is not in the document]
pixelsoccupied: Removed

controllers/prep_handlers.go
```diff
@@ -481,6 +450,51 @@ func (r *ImageBasedUpgradeReconciler) handlePrep(ctx context.Context, ibu *lcav1
...
+func (r *ImageBasedUpgradeReconciler) updatePrepTaskForPrecacheJob(ctx context.Context) {
+	if _, err := precache.GetJob(ctx, r.Precache.Client); err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("Precache job not launched yet")
+		} else {
+			r.failPrepTask(err, "prep stage failed, unable to get precache job")
+		}
+		return
+	}
+
```
Missxiaoguo, line 463: This should be handled in the following precache job status check already as it needs to get the job first. It seems like a duplicate?

pixelsoccupied: Yeah like the getJob call was buried deep and with ton of error wraps...wanted to keep it as it is. But yeah makes sense to just refactor!

pixelsoccupied: Removed the extra call and instead bubbling up the error

controllers/prep_handlers.go
```diff
@@ -481,6 +450,51 @@ func (r *ImageBasedUpgradeReconciler) handlePrep(ctx context.Context, ibu *lcav1
...
+}
+
+// successPrepTask set PrepTask to success
+func (r *ImageBasedUpgradeReconciler) successPrepTask(msg string) {
+	r.PrepTask.Progress = msg
+	r.PrepTask.Success = true
+	close(r.PrepTask.done)
+}
+
+// successPrepTask set PrepTask to fail
```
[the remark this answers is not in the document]
pixelsoccupied: Removed

pixelsoccupied:
/hold
Need to handle a case where user deletes the job randomly

pixelsoccupied:
/unhold
a finalizer is now added during job render. Allows us to check if deletion is requested and fail it asap.

controllers/ibu_controller.go
```diff
@@ -506,6 +505,7 @@ func (r *ImageBasedUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return false
 			},
 		})).
+		Owns(&kbatch.Job{}). // note: job resource watched is restricted further using cache.Options during NewManager
```
donpenney, line 508: Cool!

controllers/idle_handlers.go
```diff
@@ -126,26 +126,32 @@ func (r *ImageBasedUpgradeReconciler) cleanup(
 		r.Log.Error(err, msg)
 		errorMessage += msg + " "
 	}
-	// Terminate precaching worker thread
+
+	r.Log.Info("Terminating precaching worker thread, will wait up to 30 seconds")
```
donpenney, line 130: I love the extra logs. I think we need to be more verbose in general, as the logs will be all we have if we need to debug external issues

pixelsoccupied: Yeah putting these logs helped me find an issue without the needing to bring out the debugger/breakpoints enabled build. Wish we had a `.Debug` be have less visual clutter but then again `info` and `debug` sometimes are interchangeable

donpenney: Plus, it's harder to go back and get an additional level of logs later, especially if it's a one-off or rare issue / race condition

PR #363, opened 2024-02-28
CNF-11488: Allow spec changes when transitioning to idle

sudomakeinstall2:
spec changes are blocked when ibu is not idle

we should allow spec changes when we transition to idle to be able to reset ibu and go idle in one step

for this to work handle abort should not rely on image version on spec so handle abort will now remove all unbootes stateroots


donpenney:
Does this account for an invalid transition to Idle? Like if I change the spec to set stage to Idle and delete the seedImageRef bits?

donpenney:
Is this the right Jira?

sudomakeinstall2:
> Is this the right Jira?

no i fixed it.

sudomakeinstall2:
> Does this account for an invalid transition to Idle? Like if I change the spec to set stage to Idle and delete the seedImageRef bits?

yes the point of the PR is to be able to delete spec fields and set stage to idle in one step.
This does not affect invalid transition you still can't transition to idle after pivot, you will get an invalid transition condition.

donpenney:
> > Does this account for an invalid transition to Idle? Like if I change the spec to set stage to Idle and delete the seedImageRef bits?
> 
> yes the point of the PR is to be able to delete spec fields and set stage to idle in one step. This does not affect invalid transition you still can't transition to idle after pivot, you will get an invalid transition condition.

Sorry... just to clarify my question:
* I'm post-pivot, upgrade failed
* I edit the spec to set the stage to Idle and delete the seedImageRef fields
* This is an invalid transition, but I will now be able to do this, leaving me in an invalid state with no seedImageRef?

sudomakeinstall2:
> > > Does this account for an invalid transition to Idle? Like if I change the spec to set stage to Idle and delete the seedImageRef bits?
> > 
> > 
> > yes the point of the PR is to be able to delete spec fields and set stage to idle in one step. This does not affect invalid transition you still can't transition to idle after pivot, you will get an invalid transition condition.
> 
> Sorry... just to clarify my question:
> 
> * I'm post-pivot, upgrade failed
> * I edit the spec to set the stage to Idle and delete the seedImageRef fields
> * This is an invalid transition, but I will now be able to do this, leaving me in an invalid state with no seedImageRef?

That's a good point. if this happens you can set the `seedImageRef` again.
But if we don't want to allow this to happen, we can limit this to only when `UpgradeCompleted=True` and before upgrade starts. This way invalid transition won't happen, but you can't go idle and clear ibu after upgrade is started until it is finished.

@jc-rh

jc-rh:
> > > > Does this account for an invalid transition to Idle? Like if I change the spec to set stage to Idle and delete the seedImageRef bits?
> > > 
> > > 
> > > yes the point of the PR is to be able to delete spec fields and set stage to idle in one step. This does not affect invalid transition you still can't transition to idle after pivot, you will get an invalid transition condition.
> > 
> > 
> > Sorry... just to clarify my question:
> > 
> > * I'm post-pivot, upgrade failed
> > * I edit the spec to set the stage to Idle and delete the seedImageRef fields
> > * This is an invalid transition, but I will now be able to do this, leaving me in an invalid state with no seedImageRef?
> 
> That's a good point. if this happens you can set the `seedImageRef` again. But if we don't want to allow this to happen, we can limit this to only when `UpgradeCompleted=True` and before upgrade starts. This way invalid transition won't happen, but you can't go idle and clear ibu after upgrade is started until it is finished.
> 
> @jc-rh

Do we want to allow reset during abort?

donpenney (commented):
/lgtm

PR #364, opened 2024-02-28
MGMT-16515: Set the node IP hint file according to the node ip in the manifest

tsorya:
[MGMT-16515](https://issues.redhat.com//browse/MGMT-16515): Set the node IP hint file according to the node ip in the manifest

1. In case ip was provided set cidr that matches this ip in nodeip hint file
2. Adding one Network manager restart as part of physical network configuration

### DIFF
--- a/lca-cli/postpivot/postpivot.go
+++ b/lca-cli/postpivot/postpivot.go
@@ -58,6 +58,7 @@ var (
 	hostnameFile       = "/etc/hostname"
 	nmConnectionFolder = common.NMConnectionFolder
 	nodeIpFile         = "/run/nodeip-configuration/primary-ip"
+	nodeIPHintFile     = "/etc/default/nodeip-configuration"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


lca-cli/postpivot/postpivot.go
```diff
@@ -478,6 +479,47 @@ func (p *PostPivot) setNodeIPIfNotProvided(ctx context.Context,
...
+	// network online and ip probably will not be configured on first run
+	_ = wait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (done bool, err error) {
+		addresses, err := p.ops.ListNodeAddresses()
+		if err != nil {
+			p.log.Warnf("failed to list node addresses, will retry. err: %s", err.Error())
+			return false, nil
+		}
+
+		for _, addr := range addresses {
+			ip, cidr, _ := net.ParseCIDR(addr.String())
```
eranco74, line 500:
```suggestion
			ip, subnet, _ := net.ParseCIDR(addr.String())
```

lca-cli/postpivot/postpivot.go
```diff
@@ -478,6 +479,47 @@ func (p *PostPivot) setNodeIPIfNotProvided(ctx context.Context,
 	return nil
 }
 
+func (p *PostPivot) setNodeIPHint(ctx context.Context, nodeIp, nodeIPConfFile string) error {
+	if nodeIp == "" {
+		// TODO: maybe better to delete nodeip hint file in this case?
+		p.log.Info("Node ip was not provided, skipping node ip hint changes")
+		return nil
+	}
+	var hintSubnet net.IP
```
eranco74, line 488: I wonder if we care about the subnet?
We know the IP, why not just write the IP with CIDR 32?
192.168.127.10/32, won't this work?

lca-cli/postpivot/postpivot.go
```diff
@@ -649,21 +691,40 @@ func (p *PostPivot) copyNMConnectionFiles(source, dest string) error {
 	return nil
 }
 
+// physicalNetworkConfigurations is on charge of configuring host network prior installation process.
```
eranco74, line 694:
```suggestion
// physicalNetworkConfigurations is in charge of configuring the host network prior installation process.
```

lca-cli/postpivot/postpivot.go
```diff
@@ -649,21 +691,40 @@ func (p *PostPivot) copyNMConnectionFiles(source, dest string) error {
...
+		return err
+	}
+
+	if _, err := p.ops.SystemctlAction("restart", nmService); err != nil {
+		return fmt.Errorf("failed to restart network manager service, err %w", err)
+	}
+	return nil
+}
+
 // networkConfiguration is on charge of configuring network prior installation process.
```
eranco74, line 714:
```suggestion
// networkConfiguration is in charge of configuring the network prior installation process.
```

lca-cli/postpivot/postpivot.go
```diff
@@ -649,21 +691,40 @@ func (p *PostPivot) copyNMConnectionFiles(source, dest string) error {
...
 // 5. Restart NM and dnsmasq in order to apply provided configurations
 func (p *PostPivot) networkConfiguration(ctx context.Context, seedReconfiguration *clusterconfig_api.SeedReconfiguration) error {
-	if err := p.copyNMConnectionFiles(
-		path.Join(p.workingDir, common.NetworkDir, "system-connections"), nmConnectionFolder); err != nil {
+	if err := p.physicalNetworkConfigurations(seedReconfiguration); err != nil {
 		return err
 	}
 
-	if err := p.applyNMStateConfiguration(seedReconfiguration); err != nil {
+	if err := p.setNodeIPHint(ctx, seedReconfiguration.NodeIP, nodeIPHintFile); err != nil {
```
eranco74, line 727: Perhaps we should do this after executing `setNodeIPIfNotProvided` ?

tsorya: they actually mutually exclusive, one will run only of ip was provided and another one will only if not but sure we can change

lca-cli/postpivot/postpivot.go
tsorya: actually it is cidr so not sure why you want it as subnet. I believe better to stick to the same naming as in assisted

lca-cli/postpivot/postpivot.go
tsorya: nope, nodeip-hint expects to get like 192.168.127.0 , first ip in subnet

lca-cli/postpivot/postpivot.go
tsorya: we can actually in this case ip will always exists and we will just update nodeip hint with right value.

PR #365, opened 2024-02-29
OCPBUGS-30063: Restore fix for IBU with no rhcos delta

donpenney:
In an IBU where both releases have the same underlying rhcos image, the parent commit of the deployment has unique commit IDs (due to import from seed), but the same checksum. This previously caused a failure after creating the deployment, which was resolved with: https://github.com/openshift-kni/lifecycle-agent/pull/203/commits

The fix required two changes:
- Add the "--no-prune" option to the "ostree admin deploy" command
- Add a call to "rpm-ostree cleanup -b" to update the base refs

Code restructuring in a later update moved the "rpm-ostree cleanup -b" call later in the workflow, inadvertently leading to the no-delta IBU failure reoccurring.

This update moves the cleanup step into the function calling "ostree admin deploy", with a clarifying comment explaining that it must immediately follow the deploy command.

### DIFF
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -253,10 +253,6 @@ func (r *ImageBasedUpgradeReconciler) SetupStateroot(ctx context.Context, ibu *l
 		return fmt.Errorf("failed to setup stateroot: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


eranco74 (commented):
/lgtm

PR #374, opened 2024-03-04
MGMT-16610: add proxy to seed reconfiguration api

tsorya:
Part 1: adding proxy to api


tsorya:
waiting for pr from Omer

PR #375, opened 2024-03-04
MGMT-17115: Save the custom catalog sources as extra manifests

eranco74:
LCA should preserve the user-defined catalog sources 
This change will fetch the catalog sources pre-reboot, filter out the default catalog sources, and add the user-defined catalog sources as as extra manifests that will get applied post-reboot (before the application restore starts)

### DIFF
--- a/internal/clusterconfig/clusterconfig.go
+++ b/internal/clusterconfig/clusterconfig.go
@@ -6,15 +6,18 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/go-logr/logr"
-	v1 "github.com/openshift/api/config/v1"
-	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/strings/slices"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/go-logr/logr"
+	v1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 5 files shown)


internal/clusterconfig/clusterconfig.go
```diff
@@ -427,3 +452,39 @@ func (r *UpgradeClusterConfigGather) fetchNetworkConfig(ostreeDir string) error
...
+
+func (r *UpgradeClusterConfigGather) getCustomCatalogSources(ctx context.Context) (operatorsv1alpha1.CatalogSourceList, error) {
+	catalogSources := &operatorsv1alpha1.CatalogSourceList{}
+	customCatalogSources := &operatorsv1alpha1.CatalogSourceList{}
+	allNamespaces := client.ListOptions{Namespace: metav1.NamespaceAll}
+	if err := r.Client.List(ctx, catalogSources, &allNamespaces); err != nil {
+		return operatorsv1alpha1.CatalogSourceList{}, fmt.Errorf("failed to list all catalogueSources %w", err)
+	}
+
+	defaultCatalogRecources := []string{"certified-operators", "community-operators", "redhat-marketplace", "redhat-operators"}
```
omertuc, line 464: Maybe align with https://github.com/openshift-kni/lifecycle-agent/blob/e3bd855206b7ec61ec93822217fdd92158b23dc8/lca-cli/seedcreator/seedcreator.go#L478-L485

omertuc (changes requested):

internal/clusterconfig/clusterconfig.go
eranco74: Hmm, I think that one is filtering the images, but let me check with @tsorya

tsorya: yep, this regex matches image names

eranco74: So I think for this case, we just want to filter out the default catalog sources and we achieve that by filtering the catalog source name.

eranco74:
/hold
testing it...

eranco74:
/unhold
The catalogsources looks good post reboot

PR #376, opened 2024-03-04
MGMT-16807: setting required no proxy values

tsorya:
First commit is adding proxy api though i assume we will take from @omertuc proxy work, 
will leave it draft till his changes or other decision.
Second commit adds no proxy values exactly the same as installer 
Third commit removes possible duplicated in no proxy string


eranco74:
LGTM

PR #377, opened 2024-03-05
post-pivot - Cleanup seed's SriovFecNodeConfig CR if exists

leo8a:
In Telco deployments, the SRIOV-FEC operator may be deployed on the seed, and hence the seed's SriovFecNodeConfig CR would need to be cleaned up after pivoting.

### DIFF
--- a/lca-cli/postpivot/postpivot.go
+++ b/lca-cli/postpivot/postpivot.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


leo8a:
/jira refresh
/hold 

testing it...

browsell:
This is embedding speicif knowledge of an operator into LCA which is not desireable.

leo8a:
closing as the SR-IOV operator is the one who should remove nodeState objects from non existing / nodes without a config-daemon

we can open again in the future if we finally decide to implement such cleaner in LCA (post-pivot)

PR #381, opened 2024-03-07
MGMT-16819: LCA should use the same recert image across seed generation and post-pivot

eranco74:


lca-cli/seedcreator/seedcreator.go
```diff
@@ -96,6 +97,7 @@ func (s *SeedCreator) CreateSeedImage() error {
 	if err := utils.RunOnce("gather_cluster_info", common.BackupChecksDir, s.log, s.gatherClusterInfo, ctx); err != nil {
 		return fmt.Errorf("failed to run once gather_cluster_info: %w", err)
 	}
+	s.ensureRecertImageWithDigestFormat()
```
omertuc, line 100: Too late, `gatherClusterInfo` already [establishes](https://github.com/openshift-kni/lifecycle-agent/blob/99c6a4cb46939629e793dbaa832cc2bd2d1c562d/lca-cli/seedcreator/seedcreator.go#L189) the seed cluster info (including which recert image seed uses) one line above:

https://github.com/openshift-kni/lifecycle-agent/blob/99c6a4cb46939629e793dbaa832cc2bd2d1c562d/lca-cli/seedcreator/seedcreator.go#L97-L99

omertuc (changes requested):
Bug

lca-cli/seedcreator/seedcreator.go
eranco74: Thanks, fixedd

eranco74:
Seems to work:

The recert image in the SeedGenerator CR:
```
  recertImage: quay.io/edge-infrastructure/recert:v0
```
After generating the seed:
```
sudo cat /var/seed_data/manifest.json 
{"seed_cluster_ocp_version":"4.15.0","base_domain":"ibo0.redhat.com","cluster_name":"seed","node_ip":"192.168.126.10","release_registry":"quay.io","sno_hostname":"seed","recert_image_pull_spec":"quay.io/edge-infrastructure/recert@sha256:4e30af484263497eb652723de40ff99169a88d6ac6605497b1fd041152dbbe0b"}[
```

omertuc:
> If we do this, though, won't seed images used during dev become invalidated the next time the recert image is updated? The exact image used during seedgen would no longer be available in the registry once the tag is updated, unless it is explicitly retagged to save the image

Good point. Possible solutions:

* Make sure our infra tags all dev images also with a dummy tag that is simply just the first n letters of their hash

* Seedgen CR / LCA parameter to override this pinning behavior

eranco74:
Thanks @donpenney.
@omertuc I think we can do both options.
One note though, essentially the 2nd option already exists today, when creating the seedgen CR the user can provide the recert image by digest, I'm unsure whether a parameter in the CR that triggers the conversion from the image tag to digest is  better than instructing the user to provide the recert image by digest.

On another note, given the overhead this change creates (keeping multiple recert images we can't clean) perhaps recert should be backward compatible, meaning if an older recert image invalidated the seed image a newer recert image should be able to recertify it.

donpenney:
In the production environment, we'll soon be using the RELATED_IMAGE_RECERT_IMAGE env var from the deployment as the "default" recert image, which would be a SHA reference where the image is protected from deletion. Once that's in place, dev envs would likely be the only cases using a tagged image. So maybe it's sufficient to provide the warning in the designer docs about using a tagged recert image? Could recommend using a mirrored tag if someone would want to be able to use a seed image for longer periods and ensure the same recert image is used. I've done that myself for certain test scenarios, certainly - make a timestamped tag in my own repo and use that in the seedgen CR.

But as noted in another comment, the seedgen can use a SHA reference today, if a user wants to have that explicit image used.

I think we can probably just cover this in the MD files, with warnings/recommendations. Mirroring the image is certainly an easy way to ensure you have the same recert image being used for longer tests

eranco74:
I tend to agree with @donpenney.
In addition, today dev users can benefit from recert improvements as they merge without rebuilding the seed image.

Note that I looked into pushing recert images by tag and git revision (in order to allow this change without losing old revisions) while it's straight forwarded for the CI registry it's the case for mirroring the images to quay.io.

eranco74:
Closed in favor of https://github.com/openshift-kni/lifecycle-agent/pull/401

PR #316, opened 2024-02-14
Add documentation for Finalize/Abort failed

sudomakeinstall2:

PR #387, opened 2024-03-12
Add automatic cleanup for DeleteBackupRequests CRs when manualCleanUp annotation is set

leo8a:
When facing network issues with the S3 storage backed, the user might need to also have automatic cleanup for the DeleteBackupRequests CRs that are leftover.

### DIFF
--- a/internal/backuprestore/backup.go
+++ b/internal/backuprestore/backup.go
@@ -506,6 +506,7 @@ func (h *BRHandler) CleanupBackups(ctx context.Context) (bool, error) {
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      backup.Name,
 				Namespace: backup.Namespace,
+				Labels:    map[string]string{clusterIDLabel: clusterID},
 			},
 			Spec: velerov1.DeleteBackupRequestSpec{
 				BackupName: backup.Name,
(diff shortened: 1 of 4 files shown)


leo8a:
> Is there a related Jira for this update?

not that I'm aware of, the main context for this update is contained in [this discussion](https://github.com/openshift-kni/lifecycle-agent/pull/316#discussion_r1511900881)

internal/backuprestore/backup.go
```diff
@@ -565,3 +565,26 @@ func (h *BRHandler) ensureBackupsDeleted(ctx context.Context, backups []velerov1
 	}
 	return true, nil
 }
+
+// CleanupDeleteBackupRequests deletes all DeleteBackupRequest for this cluster from object storage
+// returns: true if all DeleteBackupRequest have been deleted, error
+func (h *BRHandler) CleanupDeleteBackupRequests(ctx context.Context) (bool, error) {
```
Missxiaoguo, line 571: I think only one return variable err is needed.
```suggestion
func (h *BRHandler) CleanupDeleteBackupRequests(ctx context.Context) error {
```

internal/backuprestore/backup.go
```diff
@@ -565,3 +565,26 @@ func (h *BRHandler) ensureBackupsDeleted(ctx context.Context, backups []velerov1
...
+
+	// List all DeleteBackupRequest CRs created for this cluster
+	deleteBackupRequestList := &velerov1.DeleteBackupRequestList{}
+	if err := h.List(ctx, deleteBackupRequestList); err != nil {
+		return false, fmt.Errorf("failed to list DeleteBackupRequest CRs: %w", err)
```
Missxiaoguo, line 576: Could we assgin a specific label to the deleteBackupRequest CRs to ensure that we are not deleting other CRs not created by us?
We could use the same label `clusterIDLabel: clusterID` of the backup CR

Missxiaoguo:
@leo8a Thanks for the PR!

leo8a:
code rebased, PTAL

PR #402, opened 2024-03-17
OCPBUGS-30190: Cleanup stale Backups during Prep stage

leo8a:
If the target cluster is reinstalled before cleaning up any previous Backups from the object storage, OADP may sync those stale Backup CRs back to this cluster, even though they are labeled with a different `clusterID`.

Handling those stale Backups early during the `Prep` stage would save the time required to delete those (via `DeleteBackupRequest` CR with OADP) from the `Upgrade` stage.

/cc @Missxiaoguo @jc-rh @browsell


leo8a:
should be merged after #387 

will also add checks to ensure `Backups` / `Restore` resources use the proper `clusterID` label during the `Upgrade` stage in a follow-up PR

leo8a:
rebased code to trigger a /!retest

PR #403, opened 2024-03-18
OCPBUGS-30190: Handle stale Backups during Upgrade stage

leo8a:
If the target cluster is reinstalled before cleaning up any previous `Backups` from the object storage, OADP may sync those stale `Backup` CRs back to this cluster, even though they are labeled with a different `clusterID`.

This piece handles those stale backups during the `Upgrade` phase, basically we added a check to use only those `Backup` / `Restore` resources that matches the `clusterID` of the target cluster. Additionally, we also incorporated the `CleanupStaleBackups` function during this stage (for the case when any `Backup` is arbitrarily / unexpectedly changed between `Prep` and `Upgrade` stages).

Signed-off-by: Leonardo Ochoa-Aday [lochoa@redhat.com](mailto:lochoa@redhat.com)

/cc @Missxiaoguo @jc-rh @browsell

### DIFF
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -477,6 +477,12 @@ func (u *UpgHandler) HandleBackup(ctx context.Context, ibu *lcav1alpha1.ImageBas
 	// trigger and track each group
 	for index, backups := range sortedBackupGroups {
 		u.Log.Info("Processing backup", "groupIndex", index+1, "totalGroups", len(sortedBackupGroups))
+
+		// check for any stale backup in the group
+		if err := u.BackupRestore.CleanupStaleBackups(ctx, backups); err != nil {
+			return requeueWithError(fmt.Errorf("failed to cleanup stale Backups: %w", err))
+		}
+
 		backupTracker, err := u.BackupRestore.StartOrTrackBackup(ctx, backups)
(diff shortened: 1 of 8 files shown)


leo8a:
/hold

waiting for #402 to be merged first

leo8a:
rebased code, PTAL

controllers/upgrade_handlers.go
```diff
@@ -477,6 +477,14 @@ func (u *UpgHandler) HandleBackup(ctx context.Context, ibu *lcav1alpha1.ImageBas
 	// trigger and track each group
 	for index, backups := range sortedBackupGroups {
 		u.Log.Info("Processing backup", "groupIndex", index+1, "totalGroups", len(sortedBackupGroups))
+
+		// check for any stale backup in the group
+		if allStaleBackupsRemoved, err := u.BackupRestore.CleanupStaleBackups(ctx, backups); err != nil {
```
Missxiaoguo, line 482: The function CleanupStaleBackups was changed in [PR404](https://github.com/openshift-kni/lifecycle-agent/pull/404/files) to return only one variable.

Missxiaoguo (changes requested):

leo8a:
code rebased, PTAL

PR #407, opened 2024-03-19
OCPBUGS-30190: Improve CleanupStaleBackups function

leo8a:
This fixes the `CleanupStaleBackups` function for when there aren't Backups in the cluster. If the `getBackup` call returns `nil`, it now continues to the next one, avoiding any possible panic.

/cc @browsell @Missxiaoguo


leo8a:
follow-up PR of #402

Missxiaoguo (approved):

PR #408, opened 2024-03-20
MGMT-17283 - IBI - Run restorecon after mounting /var/lib/containers

javipolo:
/cc @donpenney 
/cc @eranco74

### DIFF
--- a/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
+++ b/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
@@ -42,6 +42,7 @@ else
     chattr +i /mnt/
     mount -o bind /mnt/containers /var/lib/containers
 fi
+restorecon -R /var/lib/containers
 
 additional_flags=""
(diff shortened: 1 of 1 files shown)


eranco74 (commented):
/lgtm

javipolo:
It makes more sense to run restorecon in the live ISO

please @eranco74 review again :sweat_smile: 

/unhold

romfreiman:
Yea, lets understand first why is it need and not happens in IBU

javipolo:
in IBU the directory/filesystem is created by ignition, which should properly handle the SELinux relabeling
in IBI the directory/filesystem is created by our tool, so we need to handle this first SELinux labeling

once the main directory /var/lib/containers are properly labeled, newly created files will inherit those labels, so this only needs to be done right after we mount it for the first time

this PR now handles exactly that case, labeling the empty /var/lib/containers while in RHCOS live iso, and this code is only run the first time, just before using /var/lib/containers for the first time

eranco74 (commented):
/lgtm
/approve

PR #409, opened 2024-03-20
CNF-11502: Handle invalid stage transition condition

Missxiaoguo:
- add xvalidation rule to reject invalid stage transition based on the status.validNextStages list
- update the internal validation check(function validateStageTransition) to leverage the same logic to validate transition
- clear any invalid transition condition if exist when moving to the valid stage
- add the annotation "lca.openshift.io/triggerReconcile" to provide a backdoor for LCA reconcile if unexpected xvalidation rejects any spec changes
- doc updates

/cc @jc-rh @donpenney @browsell


controllers/ibu_controller.go
```diff
@@ -306,147 +315,90 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
...
+
+	// Check if the stage is in the valid next stage list
+	ok := lo.Contains(validStages, ibu.Spec.Stage)
+	if ok {
+		// The transition is valid, clear any invalid transition conditions if exist and set proper condition
+		switch ibu.Spec.Stage {
+		case lcav1alpha1.Stages.Prep:
+			utils.ClearInvalidTransitionStatusCondition(ibu, lcav1alpha1.Stages.Upgrade)
+			utils.ClearInvalidTransitionStatusCondition(ibu, lcav1alpha1.Stages.Rollback)
+			utils.SetIdleStatusInProgress(ibu, utils.ConditionReasons.InProgress, "In progress")
```
donpenney, line 353: This had me confused, admittedly. It threw me that we were setting two conditions to in-progress, and I thought it was a copy/paste error :) ... Maybe a comment here to explicitly note that this is to mark the overall IBU as "in progress", so I don't get confused again next time I see the code

controllers/ibu_controller.go
```diff
@@ -306,147 +315,90 @@ func (r *ImageBasedUpgradeReconciler) handleAbortOrFinalize(ctx context.Context,
...
+			utils.SetPrepStatusInProgress(ibu, "In progress")
+		case lcav1alpha1.Stages.Upgrade:
+			utils.ClearInvalidTransitionStatusCondition(ibu, lcav1alpha1.Stages.Rollback)
+			utils.SetUpgradeStatusInProgress(ibu, "In progress")
```
donpenney, line 357: nit: We should probably have a const in conditions.go for "In progress" string, since its used a number of times in the code - especially in ibu_controller_test.go

Missxiaoguo: Actually, I considered the same thing. I just ended up being lazy >_<

controllers/ibu_controller.go
Missxiaoguo: Actually, as we now use one logic, I realized that clearing doesn't necessarily have to be handled separatedly for each stage. As long as it transitions to a valid stage, we can clear any existing invalid conditions. It can be a single function in the "if ok" block. It doesn't hurt to check invalid transition in the cases where they are not expected, but it will make this function looking more clear.

controllers/ibu_controller_test.go
```diff
@@ -644,7 +753,7 @@ func TestValidateStageTransisions(t *testing.T) {
 					utils.ConditionTypes.UpgradeCompleted,
 					utils.ConditionReasons.Failed,
 					metav1.ConditionFalse,
-					"Rollback requested",
+					"Upgrade failed",
```
jc-rh, line 756: @donpenney I think you specifically asked for "rollback requested"?

donpenney: If it's a user-driven rollback, it's not a "failed upgrade" - it's an aborted upgrade. So I don't think we would want to set the message to "failed", as it could be misleading.

controllers/ibu_controller.go
```diff
@@ -474,6 +426,15 @@ func (r *ImageBasedUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 				}
 
+				// trigger reconcile upon adding or updating TriggerReconcileAnnotation
+				oldValue, oldExist := e.ObjectOld.GetAnnotations()[utils.TriggerReconcileAnnotation]
+				newValue, newExist := e.ObjectNew.GetAnnotations()[utils.TriggerReconcileAnnotation]
+				if oldExist && !newExist {
+					return false
+				} else if !oldExist && newExist || oldValue != newValue {
+					return true
+				}
```
pixelsoccupied, line 436: i wonder if this could be simplified a bit? 

I thinking return early with false if reflect.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations())

otherwise return oldExist != newExist?

pixelsoccupied: `or updating TriggerReconcileAnnotation` as in removing yeah?

Missxiaoguo: yeah, right, so adding/removing `ManualCleanupAnnotation` can trigger reconcile, but we want a dedicated annotation for this particular purpose when really bad things happen. And we want to limit the reconciliation to these two specific annotations only.

Getting annotations and comparing should be super pretty quick so I don't see the necessity to add additional check to `return early with false if reflect.DeepEqual(e.ObjectNew.GetAnnotations(), e.ObjectOld.GetAnnotations())`.  

I update the check for TriggerReconcileAnnotation to one line. It should be clearer.

controllers/utils/conditions.go
```diff
@@ -82,6 +82,20 @@ var ConditionReasons = struct {
 	InvalidTransition: "InvalidTransition",
 }
 
+// Condition messages
```
donpenney, line 85: Maybe note that this is just the set of common messages, not the complete set, as the message field is a freeform string that is also used for other status messages and relaying errors. ie. the message field can contain other values besides just this set

Missxiaoguo: Sure! I considered having these common messages in a conditionMessage struct but thought it might cause confusion as we have other custom messages. Updated the comment.

controllers/utils/constants.go
```diff
@@ -12,7 +12,8 @@ const (
 	IBUName     string = "upgrade"
 	IBUFilePath string = common.LCAConfigDir + "/ibu.json"
 
-	ManualCleanupAnnotation string = "lca.openshift.io/manualCleanupDone"
+	ManualCleanupAnnotation    string = "lca.openshift.io/manualCleanupDone"
+	TriggerReconcileAnnotation string = "lca.openshift.io/triggerReconcile"
```
pixelsoccupied, line 16: I wonder if we should start converging to the recommended syntax guide? https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

`lca.openshift.io/trigger-reconcile`?

donpenney: Is the use of camel case discouraged? It doesn't seem like these values are breaking the syntax described at that page. Capitals are allowed, for example

Missxiaoguo: Yeah, capticals are definitely allowed. 
See examples of k8s annotations using this format https://kubernetes.io/docs/reference/labels-annotations-taints/#scheduleralphakubernetesio-defaulttolerations.

I think the guide states that `The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.` means you could have "-"," _", "." in between but not mandatory. 
Even the example it provides doesn't have '-' in its name.
  annotations:
    imageregistry: "https://hub.docker.com/"

pixelsoccupied: For sure capital is allowed. 

Some more on this
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions 

```
Other advice regarding use of labels, annotations, taints, and other generic map keys by Kubernetes components and tools:

Key names should be all lowercase, with words separated by dashes instead of camelCase
For instance, prefer foo.kubernetes.io/foo-bar over foo.kubernetes.io/fooBar, prefer desired-replicas over DesiredReplicas
```

donpenney: I have no issue with using camelCase here, myself. It aligns with the variable naming conventions and doesn't break any rules. The note about using dashes instead of camel case is "advice", seeming to be suggested as possible guidelines for a subsystem, but that it's left to users to define their own rules. It seems odd to me to have a different convention for the annotation naming vs variables, almost opposing conventions. But I have no issues with whichever way people want to go.

PR #410, opened 2024-03-25
Use recert for proxy seed reconfiguration

omertuc:
# Overview

LCA will now use recert's new proxy feature. This will allow us to
support environments where the seed proxy is different than the upgraded
cluster's proxy. It will also allow us to perform a rollout-free seed
proxy reconfiguration.

Clusters with a proxy can only be upgraded using seeds that have a
proxy. Clusters without a proxy can only be upgraded using seeds that
don't have a proxy.

# Notes

Two new fields have been added to `SeedReconfiguration`, `Proxy` and
`StatusProxy` - see commit diff for an explanation of why we need both
and how they're determined in every scenario.

Added a new `HasProxy` field to the seed cluster info struct.

Seed images will now contain a `com.openshift.lifecycle-agent.seed_cluster_info`
OCI label which will contain a JSON serialized seed cluster info struct.
This label will allow LCA (and maybe IBIO in the future?) to make
informed decisions about the compatibility of a particular seed image
with the cluster being upgraded. For example, LCA will refuse to upgrade
a cluster that has a proxy with a seed that doesn't. This label helps
LCA knows whether the seed has a proxy or not (through the new
`HasProxy` field), and block the upgrade appropriately before the image
even gets pulled.


controllers/prep_handlers.go
```diff
@@ -97,28 +124,40 @@ func (r *ImageBasedUpgradeReconciler) checkSeedImageCompatibility(_ context.Cont
...
+// and the code constant is only manually bumped by developers when the image
+// format changes in a way that is incompatible with previous versions of the
+// lifecycle-agent.
+func checkSeedImageVersionCompatibility(labels map[string]string) error {
+	seedFormatLabelValue, ok := labels[common.SeedFormatOCILabel]
 	if !ok {
 		return fmt.Errorf(
-			"seed image %s is missing the %s label, please build a new image using the latest version of the lca-cli",
-			seedImageRef, common.SeedFormatOCILabel)
+			"seed image is missing the %s label, please build a new image using the latest version of the lca-cli",
```
donpenney, line 153: Not necessarily "latest version", but seed was generated from "incompatible version of LCA" maybe?

lca-cli/seedcreator/seedcreator.go
```diff
@@ -147,7 +147,13 @@ func (s *SeedCreator) CreateSeedImage() error {
...
 	}
 
-	if err := s.createAndPushSeedImage(); err != nil {
+	clusterInfoJSONSBytes, err := os.ReadFile(path.Join(s.backupDir, common.SeedClusterInfoFileName))
+	if err != nil {
+		return fmt.Errorf("failed to read cluster info file: %w", err)
+	}
+
+	clusterInfoJSON := string(clusterInfoJSONSBytes)
+	if err := s.createAndPushSeedImage(clusterInfoJSON); err != nil {
```
donpenney, line 156: Do we really want all this in a label? Why not just validate the data from the manifest.json after pulling the image? I get that it would save the pull, but is there some other benefit?

donpenney: That said, I do like the idea of this being an easy way to access the manifest.json data without pulling the image, for expanded image validation. Are there any cons to it? Like would we end up with restrictions on what could get added to the manifest.json? I think no, but I'm not sure

controllers/prep_handlers.go
omertuc: I agree, but unrelated to PR. Created #423

lca-cli/seedcreator/seedcreator.go
omertuc: Yeah exactly, it's nice to have. Couldn't think of any cons. It's information that's already in the seed, just now duplicated to different location as well (labels on top of inside a file in a layer), so I don't see any additional compatibility restrictions this would add.

>Like would we end up with restrictions on what could get added to the manifest.json? 

Perhaps in terms of size. I guess at that point in the future it would be an obvious problem and we could redact any future huge thing we put in `manifest.json` before we serialize it to put inside the label

leo8a (commented):
overall this looks great to me, thanks for adding it

api/seedreconfig/seedreconfig.go
```diff
@@ -92,6 +92,41 @@ type SeedReconfiguration struct {
...
+
+	// The install-config.yaml used to generate the cluster. This is used to
+	// populate the cluster-config-v1 configmaps in the cluster, which usually
+	// hold a slightly modified version of the user's installation
+	// install-config.yaml. Changing this is important because the Cluster
+	// Network Operator uses the information in this configmap to set the Proxy
+	// CR's no_proxy status field. In IBU, LCA simply copies it from the
+	// upgraded cluster. In IBI, a fake install-config should be generated by
+	// the IBIO. This parameter is required when the proxy parameters are set.
+	InstallConfig string `json:"install_config,omitempty"`
```
tsorya, line 129: What recert will use from install-config? 
It seems strange for me to provide full install-config but on the other hand we can use it as api + new ibi specific fields. 
Currently it seems like we are going to provide duplicate values and it will be sort of confusing

omertuc: See comment

image

omertuc: >What recert will use from install-config?

Recert doesn't use it itself, it only updates the entirety of the install-config configmaps to match the install-config value you give it

tsorya: Which info it uses, which rollout we see?
In IBU it really doesn't matter but in IBIO this api is not something that can be usable in my opinion at least

omertuc: https://github.com/openshift/cluster-network-operator/blob/ce9414f20ccab2151786bcc0dab9fd0f213e3f75/pkg/util/proxyconfig/no_proxy.go#L63-L78

omertuc: >which rollout we see?

Reboot, because the no_proxy settings are taken from the install-config

tsorya: > https://github.com/openshift/cluster-network-operator/blob/ce9414f20ccab2151786bcc0dab9fd0f213e3f75/pkg/util/proxyconfig/no_proxy.go#L63-L78

In this case we can't actually take install config as is from SNO as it can differ from the seed, not sure we should add new seed requirements

eranco74, line 129: OK, for now.
The only relevant information in the install-config is the machine network (added as part of https://github.com/openshift-kni/lifecycle-agent/pull/422).
For IBI IBIO can render the proxy status but I prefer if it won't have to provide a fake install-config, instead we should keep the existing install-config from the seed and use recert too update the machine network field.

controllers/prep_handlers.go
```diff
@@ -69,22 +69,49 @@ func (r *ImageBasedUpgradeReconciler) getSeedImage(
...
-// checkSeedImageCompatibility checks if the seed image is compatible with the
-// current version of the lifecycle-agent by inspecting the OCI image's labels
-// and checking if the specified format version equals the hard-coded one that
-// this version of the lifecycle agent expects. That format version is set by
-// the lca-cli during the image build process, and is only manually bumped by
-// developers when the image format changes in a way that is incompatible with
-// previous versions of the lifecycle-agent.
-func (r *ImageBasedUpgradeReconciler) checkSeedImageCompatibility(_ context.Context, seedImageRef string) error {
+func getSeedConfigFromLabel(labels map[string]string) (*seedclusterinfo.SeedClusterInfo, error) {
+	seedFormatLabelValue, ok := labels[common.SeedClusterInfoOCILabel]
```
eranco74, line 101: Is this change "beaking" all existing seed images?

api/seedreconfig/seedreconfig.go
omertuc: OK, I will add a separate machine network field in a separate PR, because this one has been taking too long, I want to unblock IBU

controllers/prep_handlers.go
omertuc: Yes. I will make it a bit more backwards compatible

donpenney (commented):
/lgtm

PR #431, opened 2024-04-08
MGMT-17463: Fix SELinux problems while unpacking tar files

javipolo:
/cc @donpenney 
/cc @eranco74

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"path"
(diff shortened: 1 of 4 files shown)


utils/utils.go
```diff
@@ -219,6 +220,38 @@ func CopyFileIfExists(source, dest string) error {
 	}})
 }
 
+func CopyToTempFile(sourceFileName, directory, pattern string) (string, error) {
```
tsorya, line 223: maybe you can use cp.Copy like in 
https://github.com/openshift-kni/lifecycle-agent/pull/431/commits/f8808ac7bb220896af3d7b960dce4a56039b3756#diff-0df82f8689d3e8d881be0653ab4e9f8221c5835ec95a14930d11669e9c7f07efR215
it will minimize this function no?

utils/utils.go
```diff
@@ -219,6 +220,38 @@ func CopyFileIfExists(source, dest string) error {
...
+		return "", fmt.Errorf("failed to open %s: %w", sourceFileName, err)
+	}
+	defer sourceFile.Close()
+
+	if _, err = io.Copy(destinationFile, sourceFile); err != nil {
+		return "", fmt.Errorf("failed to copy %s to temporary file %s: %w", sourceFileName, destinationFileName, err)
+	}
+
+	// Preserve permissions
+	sourceFileInfo, err := os.Stat(sourceFileName)
```
donpenney, line 242: Do we need to care about attributes, etc, like running `cp --preserve=all`? I guess not in this specific case

utils/utils.go
donpenney: nvm... I guess that wouldn't exist in IBI

utils/utils.go
javipolo: At the very least, we need to make the file executable

Since I was creating a more generic `CopyToTempFile` function, I thought that for the same price we could as well preserve the attributes

donpenney: That's what I was wondering. We're preserving the permissions. Just wondering if we'd need to preserve the extended attributes, for example. Certainly not needed here - maybe a function comment to note that the mode is preserved, but other file attributes are not? I'm not looking for the extended attributes to be implemented here, or attributes like owner/group/timestamp. Just a note in case anyone in the future decides to use this generic function

utils/utils.go
javipolo: Note added :+1:

utils/utils.go
javipolo: Commentary added on top of the function :)

PR #432, opened 2024-04-08
ib-cli: make extra-partition-start flag mandatory

javipolo:
/cc @eranco74


ib-cli/cmd/createiso.go
```diff
@@ -51,11 +51,12 @@ func addFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&lcaImage, "lca-image", "l", "quay.io/openshift-kni/lifecycle-agent-operator:4.16.0", "The lifecycle-agent image to use for generating the ISO.")
 	cmd.Flags().StringVarP(&rhcosLiveIso, "rhcos-live-iso", "r", "https://mirror.openshift.com/pub/openshift-v4/amd64/dependencies/rhcos/latest/rhcos-live.x86_64.iso", "The URL to the rhcos-live-iso for generating the ISO.")
 	cmd.Flags().StringVarP(&installationDisk, "installation-disk", "i", "", "The disk that will be used for the installation.")
-	cmd.Flags().StringVarP(&extraPartitionStart, "extra-partition-start", "e", "use_directory", "Start of extra partition used for /var/lib/containers. Partition will expand until the end of the disk. Uses sgdisk notation")
+	cmd.Flags().StringVarP(&extraPartitionStart, "extra-partition-start", "e", "", "Start of extra partition used for /var/lib/containers. Partition will expand until the end of the disk. Uses sgdisk notation")
```
tsorya, line 54: Why we need this flag? Shouldn't set it as default?

javipolo: If using an extra partition, the user needs to determine the size of that partition. That's not required for `directory` mode, but that method is not supported

After talking to a user, he was having problems because he did not set this, and then seed image was not booting correctly

I decided to make this flag mandatory to avoid user error. This way it will only be set to the usupported `directory` setting if manually specified

PR #433, opened 2024-04-08
CNF-11876: remove autoRollback config params from spec (except InitMonitorTimeoutSeconds) and use annotations instead

pixelsoccupied:
The autoRollback configs are now done using annotations and must only be configured for debug purposes.


pixelsoccupied:
/hold 
This is a breaking change

PR #435, opened 2024-04-09
OCPBUGS-31896: Moving set of sshkey before network manipulations

tsorya:
OCPBUGS-31896: Moving set of sshkey before network manipulations in order to allow ssh connection in case something will go wrong with network config but ip will exist

### DIFF
--- a/lca-cli/postpivot/postpivot.go
+++ b/lca-cli/postpivot/postpivot.go
@@ -98,13 +98,13 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
 		return fmt.Errorf("failed to get cluster info from %s, err: %w", "", err)
 	}
 
-	if err := p.networkConfiguration(ctx, seedReconfiguration); err != nil {
-		return fmt.Errorf("failed to configure networking, err: %w", err)
-	}
-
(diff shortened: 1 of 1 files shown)


eranco74 (commented):
/lgtm

leo8a:
ci-job is failing with the below nit, needs to be fixed before merging

(related to an empty line with indentation in it - should be removed)

```
git diff lca-cli/postpivot/postpivot.go
diff --git a/lca-cli/postpivot/postpivot.go b/lca-cli/postpivot/postpivot.go
index 00592ae1f..aea118500 100644
--- a/lca-cli/postpivot/postpivot.go
+++ b/lca-cli/postpivot/postpivot.go
@@ -106,7 +106,7 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
        if err := p.networkConfiguration(ctx, seedReconfiguration); err != nil {
                return fmt.Errorf("failed to configure networking, err: %w", err)
        }
-       
+
        if err := utils.RunOnce("pull-secret", p.workingDir, p.log, p.createPullSecretFile,
                seedReconfiguration.PullSecret, common.ImageRegistryAuthFile); err != nil {
                return fmt.Errorf("failed to run once pull-secret for post pivot: %w", err)
```

donpenney (approved):

PR #436, opened 2024-04-09
CNF-11930: Move oadp storage backend connection check to healthcheck

Missxiaoguo:
When oadp DPA is applied and reconciled, check the connectivity of the storage backend. Sometimes it may simply require more time to come up or be due to a temporary connection issue. Moving the storage backend check into healthcheck could help to ensure the connectivity of storage at each stage.

/cc @donpenney @browsell

### DIFF
--- a/internal/healthcheck/healthcheck.go
+++ b/internal/healthcheck/healthcheck.go
@@ -9,13 +9,16 @@ import (
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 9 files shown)


internal/healthcheck/healthcheck.go
```diff
@@ -95,6 +109,64 @@ func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
 	return nil
 }
 
+func IsDataProtentionApplicationReconciled(ctx context.Context, c client.Reader, l logr.Logger) (bool, error) {
```
donpenney, line 112: Typo:
```suggestion
func IsDataProtectionApplicationReconciled(ctx context.Context, c client.Reader, l logr.Logger) (bool, error) {
```

PR #438, opened 2024-04-10
Include command name in log

omertuc:
This log is missing the command name, you could only see the args


leo8a:
I guess the integration job needs to be overwritten (for now) in order to merge this one

omertuc:
I'll just wait for it to be fixed, no worries

PR #442, opened 2024-04-15
CNF-11754: Retry apply of manifest on retriable failure

Missxiaoguo:
Retry applying the manifest when it fails due to API server not responding. The initmonitor will eventually time out if it's not recovered.

/cc @browsell @donpenney


donpenney (commented):
/lgtm

PR #444, opened 2024-04-15
OCPBUGS-30327: allow restore only when IBU points to at least one OADPContent

pixelsoccupied:


lca-cli/postpivot/postpivot.go
```diff
@@ -160,11 +163,20 @@ func (p *PostPivot) PostPivotConfiguration(ctx context.Context) error {
...
+	}
+	if len(ibu.Spec.OADPContent) > 0 {
+		if err := p.restoreOadpSecrets(ctx, client); err != nil {
+			return fmt.Errorf("failed to restore OADP secrets: %w", err)
+		}
+		if err := p.restoreOadpDataProtectionApplication(ctx, client); err != nil {
+			return fmt.Errorf("failed to restore OADP DataProtectionApplication: %w", err)
+		}
+	} else {
+		p.log.Info("Locally stored IBU CR did not have any OADPContent")
```
Missxiaoguo, line 179: The check is not needed. If the directores don't exist or don't have files, it will just skip.

internal/backuprestore/restore.go
```diff
@@ -122,9 +122,14 @@ func (h *BRHandler) StartOrTrackRestore(ctx context.Context, restores []*velerov
 	return rt, nil
 }
 
-func (h *BRHandler) LoadRestoresFromOadpRestorePath() ([][]*velerov1.Restore, error) {
+func (h *BRHandler) LoadRestoresFromOadpRestorePath(ibu *lcav1alpha1.ImageBasedUpgrade) ([][]*velerov1.Restore, error) {
 	var sortedRestores [][]*velerov1.Restore
 
+	if len(ibu.Spec.OADPContent) == 0 {
```
Missxiaoguo, line 128: I don't think it's really needed since the restore dir won't be created if backup/restore is not needed.

Missxiaoguo (commented):
I think you just have to add the `ibu.spec.OADPContent` check in the pre-pivot when calling `ExportOadpConfigurationToDir` to ensure that the DPA and secret are exported only when backup/restore is needed. The post-pivot part should already be handled correctly.

lca-cli/postpivot/postpivot.go
pixelsoccupied: reverted

internal/backuprestore/restore.go
pixelsoccupied: reverted

PR #447, opened 2024-04-16
OCPBUGS-32194: pull precache images unconditionally

pixelsoccupied:
/cc @donpenney @browsell


internal/precache/helper.go
```diff
@@ -176,7 +176,7 @@ func renderJob(config *Config, log logr.Logger, ibu *v1alpha1.ImageBasedUpgrade,
 			Name:      LcaPrecacheJobName,
 			Namespace: common.LcaNamespace,
 			Annotations: map[string]string{
-				"app.kubernetes.io/name": "lifecyle-agent-precache",
+				"app.kubernetes.io/name": "lifecycle-agent-precache",
```
donpenney, line 179: This typo is unfortunately in multiple places, for the operator as well. It would probably be better to raise a new Jira and fix them all in a separate PR. 

Great catch!

donpenney (changes requested):

PR #448, opened 2024-04-16
OCPBUGS-32206: [IBU] LVMS volume contents not restored 

leo8a:
Adds support for LVMS in LCA.

The implemented workflow is as follows:
- pre-pivot: LCA patches to `Retain` all PVs created by LVMS and with a Reclaim Policy of `Delete`.
- post-pivot: LCA restores back to `Delete` the Reclaim Policy of all PVs created by LVMS.
- rollback -> idle: LCA restores back to `Delete` the Reclaim Policy of all PVs previously patched.

Also added docs and a couple of entries to the troubleshooting guide.

/cc @browsell @Missxiaoguo @donpenney @jakobmoellerdev

### DIFF
--- a/docs/backuprestore-with-oadp.md
+++ b/docs/backuprestore-with-oadp.md
@@ -228,6 +228,8 @@ metadata:
     ran.openshift.io/ztp-deploy-wave: "100"
 spec:
   configuration:
+    restic:
+      enable: false
     velero:
       defaultPlugins:
         - aws
@@ -255,6 +257,10 @@ status:
     status: "True"
     type: Reconciled
 ```
+
+> [!NOTE]
+> In IBU scenarios, the PV contents are kept on the SNO disk and reused after pivot, hence the CSI Integration (i.e.,
+> `.spec.configuration.restic`) should be always disabled in the `DataProtectionApplication`, see sample CR above.
 
 OadpSecret.yaml
 
@@ -431,7 +437,109 @@ spec:
 > 1. The examples provided are just for reference. Create your own backup and restore CRs based on your needs.
 > 2. The backup and restore CRs must be created in the same namespace where the OADP is installed which is `openshift-adp`.
 > 3. The OADP configmaps should only contain backup and restore CRs which are in YAML format or multiple document YAML format. Other type of CRs are unknown to the LCA and will be ignored.
->
+
+##### 1.1 Backup resource when LVMS is used
+
+When LVMS is used to provide dynamic storage in the cluster, we need to define the below Backup and Restore CRs, for
+the `LVMCluster` object, in the OADP ConfigMap. These objects will ensure the proper / expected functioning of this
+feature when using LVMS.
+
+> [!IMPORTANT]
+> Mind that on SNOs the persistent storage must be provided by either Logical Volume Manager Storage (LVMS) or Local
+> Storage Operator (LSO), not both.
+
+backup_lvmcluster.yaml
+
+```yaml
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  labels:
+    velero.io/storage-location: default
+  name: lvmcluster
+  namespace: openshift-adp
+spec:
+  includedNamespaces:
+    - openshift-storage
+  includedNamespaceScopedResources:
+    - lvmclusters
+    - lvmvolumegroups
+    - lvmvolumegroupnodestatuses
+```
+
+restore_lvmcluster.yaml
+
+```yaml
+apiVersion: velero.io/v1
+kind: Restore
+metadata:
+  name: lvmcluster
+  namespace: openshift-adp
+  labels:
+    velero.io/storage-location: default
+  annotations:
+    lca.openshift.io/apply-wave: "2"
+spec:
+  backupName:
+    lvmcluster
+```
+
+> [!IMPORTANT]
+> Mind that `apply-wave` annotation for the above object should be numerically lower than the app's one, in this way
+> when the app resources are restored, the `LVMCluster` operand will be already up and running in the cluster.
+
+Additionally, the Backup and Restore CR resources for the app should also include some extra fields in order to ensure
+that each PV content is maintained when performing an IBU.
+
+backup_app.yaml
+
+```yaml
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  labels:
+    velero.io/storage-location: default
+  name: small-app
+  namespace: openshift-adp
+spec:
+  includedNamespaces:
+  - test
+  includedNamespaceScopedResources:
+  - secrets
+  - persistentvolumeclaims
+  - deployments
+  - statefulsets
+  includedClusterScopedResources:
+  - persistentVolumes              # <- required field
+  - volumesnapshotcontents         # <- required field
+  - logicalvolumes.topolvm.io      # <- required field
+```
+
+> [!NOTE]
+> Given that the above Backup CR included cluster scoped resources (i.e., persistentVolumes, logicalvolumes, etc.)
+> for the small-app, there is no need to also include those in other Backup CRs. However, redundantly specifying those
+> in other Backup CRs would not affect its normal / expected operation.
+
+restore_app.yaml
+
+```yaml
+apiVersion: velero.io/v1
+kind: Restore
+metadata:
+  name: small-app
+  namespace: openshift-adp
+  labels:
+    velero.io/storage-location: default
+  annotations:
+    lca.openshift.io/apply-wave: "3"
+spec:
+  backupName:
+    small-app
+  restorePVs: true               # <- required field
+  restoreStatus:                 # <- required field
+    includedResources:           # <- required field
+      - logicalvolumes           # <- required field
+```
 
 #### 2. Build a configmap to include all CRs
 
--- a/docs/image-based-upgrade.md
+++ b/docs/image-based-upgrade.md
@@ -149,24 +149,24 @@ LCA will reject the stage transition if it is an invalid transition.
 
 This table presents a comprehensive list of conditions and valid next stages based on the current stage.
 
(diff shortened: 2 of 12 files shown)


docs/backuprestore-with-oadp.md
```diff
@@ -256,6 +258,10 @@ status:
     type: Reconciled
 ```
 
+> [!NOTE]
+> When LVMS is used to create PVs dynamically in the cluster, please ensure that the CSI Integration (i.e.,
```
browsell, line 262: We can probably do that by default, would not be needed by lso either

docs/backuprestore-with-oadp.md
```diff
@@ -431,7 +437,99 @@ spec:
 > 1. The examples provided are just for reference. Create your own backup and restore CRs based on your needs.
 > 2. The backup and restore CRs must be created in the same namespace where the OADP is installed which is `openshift-adp`.
 > 3. The OADP configmaps should only contain backup and restore CRs which are in YAML format or multiple document YAML format. Other type of CRs are unknown to the LCA and will be ignored.
->
+
+##### 1.1 Backup resource when LVMS is used
+
+When LVMS the selected provider to provide dynamic storage in the cluster, the backup and restore resources for the
```
browsell, line 443: nit, reword," When LVMS is used to provide dynamic storage in the cluster "

docs/backuprestore-with-oadp.md
```diff
@@ -431,7 +437,99 @@ spec:
...
+    velero.io/storage-location: default
+  annotations:
+    lca.openshift.io/apply-wave: "2"
+spec:
+  backupName:
+    lvmcluster
+```
+
+> [!IMPORTANT]
+> Mind that `apply-wave` annotation for below object should be numerically minor than the app's one, in this way when
```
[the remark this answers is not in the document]
leo8a: thanks

docs/backuprestore-with-oadp.md
leo8a: thanks

docs/backuprestore-with-oadp.md
```diff
@@ -431,7 +437,99 @@ spec:
...
+spec:
+  backupName:
+    small-app
+  restorePVs: true               # <- required field
+  restoreStatus:                 # <- required field
+    includedResources:           # <- required field
+      - logicalvolumes           # <- required field
+```
+
+As a complement, we also included below, sample resources to backup and restore the cluster's manifests required
```
jakobmoellerdev, line 491: Nit: I think it is important to note that this is always required, not restoring LVMCluster will break LVMS, so not sure if its a "complement"

docs/backuprestore-with-oadp.md
```diff
@@ -431,7 +437,99 @@ spec:
...
+  namespace: openshift-adp
+spec:
+  includedNamespaces:
+  - test
+  includedNamespaceScopedResources:
+  - secrets
+  - persistentvolumeclaims
+  - deployments
+  - statefulsets
+  includedClusterScopedResources:
```
jakobmoellerdev, line 464: Nit: it may be worth noting that this is only necessary exactly once per cluster restore since it is cluster scoped. (although specifying this per application is redundant, it also wont break anything)

docs/image-based-upgrade.md
```diff
@@ -149,24 +149,24 @@ LCA will reject the stage transition if it is an invalid transition.
...
-|             |                  |True  |Completed     |False      |Idle, Upgrade    |
-|`Upgrade`    |UpgradeInProgress |True  |InProgress    |False      |Idle             |
-|             |                  |True  |InProgress    |True       |Rollback         |
-|             |UpgradeCompleted  |False |Failed        |False      |Idle             |
-|             |                  |False |Failed        |True       |Rollback         |
-|             |                  |True  |Completed     |True       |Idle, Rollback   |
-|`Rollback`   |RollbackInProgress|True  |InProgress    |True       |N/A              |
-|             |RollbackCompleted |False |Failed        |True       |N/A              |
-|             |                  |True  |Completed     |True       |Idle             |
+| Current Stage | Condition          | Status | Reason         | After Pivot | Valid Next Stages |
```
jakobmoellerdev, line 152: Nit: Looks like an accidental format or did something change in the state table?

docs/backuprestore-with-oadp.md
leo8a: gotcha, maybe @jakobmoellerdev can help us confirm this

(keeping the note for now, if / when confirmed I may drop the note - if nothing against it)

jakobmoellerdev (commented):
Great resolution of Veleros issues with Backup/Restore of the PVs and double checked it against my OADP Restore POC which looks good. Just minor nits. Thanks for covering this in such detail!

docs/backuprestore-with-oadp.md
jakobmoellerdev: I think restic is only necessary in case we want to copy over the data of the PV into the backup itself. For IBU this should never be necessary since the data is kept on disk and reused after pivot.

docs/backuprestore-with-oadp.md
leo8a: totally agree, will move it up in the text and rephrase it to better highlight this, thanks for commenting it

docs/image-based-upgrade.md
leo8a: yea, nothing actually changed in the state table... was just my IDE complaining about "bad table formatting"

docs/backuprestore-with-oadp.md
leo8a: k, will drop a quick side note for that as well

docs/image-based-upgrade.md
```diff
@@ -324,13 +325,14 @@ Pivot
...
+- Once LCA starts, it will restore the saved IBU CR.
 - Restore the remaining platform configuration.
 - Wait for the platform to recover - Cluster/day2 operators and MCP are stable.
 - Apply extra manifests that were saved pre-pivot.
-- Apply any OADP restore CRs that were saved pre-pivot. Platform artifacts will be restored first including ACM artifacts if the system is managed by ACM.
+- Apply any OADP restore CRs that were saved pre-pivot. Platform artifacts will be restored first, including ACM artifacts if the system is managed by ACM.
+- If LVMS is used, automatically restores the `.spec.persistentVolumeReclaimPolicy` field of PVs to its original (pre-pivot) value.
 
-Upon completion, the condition will be updated to "Upgrade Completed".
+Upon completion, the condition will be updated to "Upgrade Completed."
```
donpenney, line 335: nit: I think the original would be correct. Both grammatically and because the condition string doesn't include a period. Mind you, the condition type also wouldn't include a space.

docs/troubleshooting.md
```diff
@@ -47,3 +47,120 @@ dataprotectionapplication-1   Available   33s              8d    true
...
+spec:
+  includedNamespaces:
+  - test
+  includedNamespaceScopedResources:
+  - secrets
+  - persistentvolumeclaims
+  - deployments
+  - statefulsets
+  includedClusterScopedResources:
+  - persistentVolumes              # <- required field
```
Missxiaoguo, line 97: I assume that all PVs will be backed up. If there are PVs created by both LSO and LVMS, will the PVs created by LSO be affected after restore in the post-pivot?

browsell: We need to update the documentation, only LSO or LVMS can be used, not both

leo8a: should we need to drop a note for that in this PR, or will it be included later on (in the official docs, for instance)?

browsell: We should do both

leo8a: Ok, just dropped [a note for it here](https://github.com/openshift-kni/lifecycle-agent/pull/448/commits/3e5e750549f3ca5cef80217ceb1938f75961a720#diff-d08037f5ce2c67f942ba8bf93d5d9c83711172f9d289737d2a44407044d1a4f8R447-R449)

/cc @amolnar-rh   (keeping you in the loop - for when adding LVMS support to official docs)

internal/backuprestore/backup.go
```diff
@@ -254,6 +254,70 @@ func (h *BRHandler) createNewBackupCr(ctx context.Context, backup *velerov1.Back
...
+		return fmt.Errorf("failed to list PersistentVolumes: %w", err)
+	}
+
+	if len(pvList.Items) == 0 {
+		h.Log.Info("No PersistentVolumes found in the cluster, skipping.")
+		return nil
+	}
+
+	for _, pv := range pvList.Items {
+		hasLvmsAnnotation := pv.GetAnnotations()[topolvmAnnotation] == topolvmValue
```
Missxiaoguo, line 273: The annotations could be nil if pv is manually created with `oc create -f  `.
It's safer to add check `if pv.GetAnnotations() != nil`

internal/backuprestore/backup.go
```diff
@@ -254,6 +254,70 @@ func (h *BRHandler) createNewBackupCr(ctx context.Context, backup *velerov1.Back
...
+		return fmt.Errorf("failed to list PersistentVolumes: %w", err)
+	}
+
+	if len(pvList.Items) == 0 {
+		h.Log.Info("No PersistentVolumes found in the cluster, skipping.")
+		return nil
+	}
+
+	for _, pv := range pvList.Items {
+		hasReclaimPolicyAnnotation := pv.GetAnnotations()[updatedReclaimPolicyAnnotation] == "true"
```
Missxiaoguo, line 304: same comment as above

Missxiaoguo (commented):
Others look good to me!

internal/backuprestore/backup.go
leo8a: nice, thanks

PR #449, opened 2024-04-17
OCPBUGS-32336: [IBU] Remove AdditionalImages from IBU CRD

leo8a:
/cc @browsell @donpenney @jc-rh

### DIFF
--- a/config/samples/lca_v1alpha1_imagebasedupgrade.yaml
+++ b/config/samples/lca_v1alpha1_imagebasedupgrade.yaml
@@ -7,9 +7,6 @@ spec:
   seedImageRef:
     version: 4.16.0
     image: quay.io/xyz
-  additionalImages:
-    name: ""
-    namespace: ""
   autoRollbackOnFailure: {}
   extraManifests:
   - name: sno-extramanifests
@@ -17,5 +14,3 @@ spec:
   oadpContent:
     - name: oadp-cm-sno-backup
       namespace: openshift-adp
-
-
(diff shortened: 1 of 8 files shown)


jc-rh:
There are two other unused fields in status. We will probably add comprehensive history info in the future instead of simple timestamps, if it ever becomes a requirement.

config/samples/lca_v1alpha1_imagebasedupgrade.yaml
```diff
@@ -3,19 +3,16 @@ kind: ImageBasedUpgrade
...
   - name: sno-extramanifests
     namespace: openshift-lifecycle-agent
   oadpContent:
     - name: oadp-cm-sno-backup
       namespace: openshift-adp
-
-
+  seedImageRef:
+    version: 4.16.0
+    image: quay.io/seed-image-pull-spec:latest
```
donpenney, line 15: This image ref must be left as quay.io/xyz. Otherwise, it causes an issue with the formal build, which includes an image ref scanner

donpenney (changes requested):

donpenney:
/cc @rauhersu 
Defining a list of reviewers in the initial PR description bypassed automatically adding Raul based on the CSV change

config/samples/lca_v1alpha1_imagebasedupgrade.yaml
leo8a: nice catch

PR #450, opened 2024-04-17
OCPBUGS-32360: fix all "lifecyle" spelling errors

pixelsoccupied:
/cc @donpenney


pixelsoccupied:
/cc @rauhersu 
Incase something you need to verify as well?

PR #454, opened 2024-04-18
Only use `summary_file_clean` recert option

omertuc:
# Background

Recert supports generating a summary of what it did. The summary
contains sensitive information. Recert supports both a regular summary
with sensitive info and a clean summary with sensitive info removed.

LCA must-gather collects recert summaries and includes them in the
must-gather

# Issue

LCA used the regular summary option instead of the clean one, this
caused sensitive information to appear in LCA must-gathers

The `summary_file` option leaks private keys, and should not be used
outside debugging / testing purposes.

# Solution

Change LCA to use the clean summary file option instead, so that
sensitive information does not get included in must-gathers

Solves OCPBUGS-32420


donpenney (commented):
/lgtm

PR #455, opened 2024-04-18
OCPBUGS-32377: [IBU] Remove SR-IOV workaround from upgrade handler

leo8a:
Removes the SR-IOV workaround from the upgrade handler.

I've kept the support for gathering info about SR-IOV operator in the must-gather thou, if this needs to be also removed, just let me know.

/cc @donpenney @browsell

### DIFF
--- a/bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
+++ b/bundle/manifests/lifecycle-agent.clusterserviceversion.yaml
@@ -508,15 +508,6 @@ spec:
         - apiGroups:
           - sriovnetwork.openshift.io
           resources:
-          - sriovnetworknodepolicies
-          verbs:
-          - get
-          - list
-          - patch
-          - watch
-        - apiGroups:
-          - sriovnetwork.openshift.io
-          resources:
           - sriovnetworknodestates
           verbs:
           - get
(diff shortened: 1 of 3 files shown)


internal/healthcheck/healthcheck.go
```diff
@@ -82,12 +80,6 @@ func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
 	}
 
 	if clusterOperatorsReady && clusterServiceVersionsReady {
-		// Only check SriovNetworkNodeState once cluster operators and CSVs are stable
-		if err := IsSriovNetworkNodeReady(ctx, c, l); err != nil {
```
donpenney, line 86: We still need the SRIOV network node ready health check. It's just the workaround in the upgrade_handlers.go that can be removed

donpenney (changes requested):

donpenney:
The only functionality to be removed here is the sriovNetworkNodeStateWorkaround function. The health check is still required

internal/healthcheck/healthcheck.go
leo8a: oh gotcha, I've shrunk the scope of the PR

donpenney (commented):
/lgtm

PR #463, opened 2024-04-19
CNF-11932: IBU: Add rollbackAvailabilityExpiration status field

donpenney:
From the original stateroot point of view, a rollback is effectively just a recovery from having the node out-of-service for a possibly extended period of time. Especially when running an IBU within the first 24 hours of deploying a cluster, this means the control plane certificates for the original release may be expired when the rollback is triggered, requiring a manual procedure in order to renew them and recover the cluster.

This update adds a new field to the IBU CR to provide the user with indication of when a rollback can be performed before the original release certificates have expired:
.status.rollbackAvailabilityExpiration


config/crd/bases/lca.openshift.io_imagebasedupgrades.yaml
```diff
@@ -204,6 +204,11 @@ spec:
               observedGeneration:
                 format: int64
                 type: integer
+              rollbackAvailabililtyExpiration:
+                description: RollbackAvailabililtyExpiration reflects the point at
+                  which rolling back may require manual recovery from expired control
+                  plane certficates.
+                type: string
```
jc-rh, line 211:
```suggestion
                format: date-time
                type: string
```

api/v1alpha1/types.go
```diff
@@ -135,6 +135,9 @@ type ImageBasedUpgradeStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Valid Next Stage"
 	ValidNextStages []ImageBasedUpgradeStage `json:"validNextStages,omitempty"`
+	// RollbackAvailabililtyExpiration reflects the point at which rolling back may require manual recovery from expired control plane certficates.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	RollbackAvailabililtyExpiration string `json:"rollbackAvailabililtyExpiration,omitempty"`
```
jc-rh, line 140: why isn't this metav1.time?

donpenney: Thanks @jc-rh ! Updated to use metav1.Time field

PR #466, opened 2024-04-23
OCPBUGS-32704: (seedgen) Avoid printing out seed cluster's pull-secret in LCA logs

leo8a:
Avoid printing out this info in the operator's logs in case the user may have configured an external aggregation log tool to extract those logs automatically through the network.


leo8a:
closing as it was fixed already by #453

PR #468, opened 2024-04-25
OCPBUGS-32939: [IBU] Add localVolume presence check in CleanupStaleBackups

leo8a:
Adds a check to the CleanupStaleBackups function to detect if a localVolume is specified in the resources in Backup CRs, and ignores its cleanup.

Also, refactors a bit the ValidateOadpConfigmaps function to reduce time complexity.

/cc @browsell @Missxiaoguo @donpenney


leo8a:
/test ibu-e2e-flow
```
 Error: Transaction test error:
  package libtool-ltdl-2.4.6-45.el9.x86_64 is already installed
```

PR #469, opened 2024-04-30
DRAFT: Cleanup disk

tsorya:
Moving ibi installation script into ibipreparation 
adding disk cleanup


lca-cli/ibi-preparation/ibipreparation.go
```diff
@@ -118,3 +139,109 @@ func (i *IBIPrepare) shutdownNode() error {
...
+	}
+
+	return nil
+}
+
+func (i *IBIPrepare) diskPreparation() error {
+	i.log.Info("Start preparing disk")
+
+	if err := i.cleanupDevice.CleanupInstallDevice(i.installationDisk); err != nil {
+		return fmt.Errorf("failed to cleanup installation disk %s, error : %w", i.installationDisk, err)
```
eranco74, line 208: Unsure if this should fail the install, need to check assisted logic.

lca-cli/ibi-preparation/ibipreparation.go
```diff
@@ -118,3 +139,109 @@ func (i *IBIPrepare) shutdownNode() error {
...
+		return fmt.Errorf("failed to grow root partition: %w", err)
+	}
+
+	return nil
+}
+
+func (i *IBIPrepare) diskPreparation() error {
+	i.log.Info("Start preparing disk")
+
+	if err := i.cleanupDevice.CleanupInstallDevice(i.installationDisk); err != nil {
```
eranco74, line 207: Allow skipping cleanup

eranco74 (commented):
I think this is better than bash since we can at least have some UT around it

PR #470, opened 2024-04-30
OCPBUGS-33012: [IBU] Add presence check for backup label in patchObj

leo8a:
Adds an error check for when the backup label is not found or any other temporary API issues. This aims to reduce the noise provoked when the label is tried to be removed and it doesn't exist. 

/cc @donpenney @browsell @Missxiaoguo

### DIFF
--- a/internal/backuprestore/backup.go
+++ b/internal/backuprestore/backup.go
@@ -223,16 +223,25 @@ func (h *BRHandler) cleanupBackupLabels(ctx context.Context, backup *velerov1.Ba
 	if err != nil {
 		return fmt.Errorf("failed to get objs from apply-label annotations: %w", err)
 	}
+
 	// json patch escape info: https://jsonpatch.com/#json-pointer
 	escaped := strings.Replace(backupLabel, "/", "~1", 1)
 	payload := []byte(fmt.Sprintf(`[{"op":"remove","path":"/metadata/labels/%s"}]`, escaped))
+
 	for _, obj := range objs {
-		err := patchObj(ctx, h.DynamicClient, &obj, false, payload) //nolint:gosec
-		if err != nil {
+		patchedObj := obj
+
+		if err := patchObj(ctx, h.DynamicClient, &patchedObj, false, payload); err != nil {
+			if k8serrors.IsNotFound(err) || k8serrors.IsInvalid(err) {
+				h.Log.Info("backup label doesn't exist, no patching needed, ignoring", "name", obj.Name,
+					"namespace", obj.Namespace, "resource", obj.Resource, "group", obj.Group, "version", obj.Version)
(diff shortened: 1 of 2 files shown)


leo8a:
/test ibu-e2e-flow

```
 Error: An rpm exception occurred: package not installed 
```

internal/backuprestore/backup.go
```diff
@@ -223,12 +223,20 @@ func (h *BRHandler) cleanupBackupLabels(ctx context.Context, backup *velerov1.Ba
...
 	// json patch escape info: https://jsonpatch.com/#json-pointer
 	escaped := strings.Replace(backupLabel, "/", "~1", 1)
 	payload := []byte(fmt.Sprintf(`[{"op":"remove","path":"/metadata/labels/%s"}]`, escaped))
+
 	for _, obj := range objs {
-		err := patchObj(ctx, h.DynamicClient, &obj, false, payload) //nolint:gosec
-		if err != nil {
+		patchedObj := obj
+
+		if hasBackupLabel, _ := searchBackupLabel(ctx, h.DynamicClient, &patchedObj); !hasBackupLabel {
```
Missxiaoguo, line 234: Looks like the deletion is skipped quitely when there is API call error. I think we should add an error log for errors other than "object not found"

internal/backuprestore/backup.go
```diff
@@ -223,12 +223,20 @@ func (h *BRHandler) cleanupBackupLabels(ctx context.Context, backup *velerov1.Ba
...
+			h.Log.Info("backup label doesn't exist, no patching needed, ignoring", "name", obj.Name,
+				"namespace", obj.Namespace, "resource", obj.Resource, "group", obj.Group, "version", obj.Version)
+			continue
+		}
+		if err := patchObj(ctx, h.DynamicClient, &patchedObj, false, payload); err != nil {
```
Missxiaoguo, line 239: Or another option would be check the error type(k8s.IsInvalid-label doesn't exist and k8s.IsNotFound) of `patch` instead of making another `get` request

leo8a: yea, I was trying to avoid changing the `patchObj` function. Also, if we check for `k8serrors.IsNotFound(err) || k8serrors.IsInvalid(err)` seems like we can't log (silently ignored) when the label is not found

leo8a: gave it another spin in my head and decided to go with the checking errors approach you suggested Angie, it seems simpler indeed, thanks!

Missxiaoguo (commented):
/lgtm

leo8a:
/test ibu-e2e-flow
```
 Error: An rpm exception occurred: package not installed 
 ```

PR #471, opened 2024-04-30
OCPBUGS-33065: deletionBackupRequests check update

Missxiaoguo:
An issue occurs during the prep and upgrade stages when cleaning up stale backup CRs. The check detects that the deleteBackupRequest is not found, but its associated backup CR still exists, causing the stage to fail. It can be easily reproduced in the prep stage. The fact is the backup has actually been deleted, but it still exists at the time  when querying the backup This is due to a very small race condition in the deleteBackupRequest check, where it queries the deleteBackupRequest CR immediately after its creation, and the updates haven't been synced in the cache yet. As a result, the deleteBackupRequest CR is not found on the first check, and LCA continues checking the backup CR, which is actually being deleted. Setting the immediate parameter to false for the poll function makes it wait for 1 second before performing the first check. Additionally, fix the issue where the request is not requeued on failure and finish reconcilation instead.

/cc @browsell @donpenney @leo8a


donpenney (commented):
/lgtm

PR #472, opened 2024-05-01
MGMT-16807: set correct no proxy and proxyStatus

tsorya:
[MGMT-16807](https://issues.redhat.com//browse/MGMT-16807): set correct no proxy and proxyStatus
We should create NoProxy and proxyStatus exactly same way as cluster network operator. 
Took CNO function and made it match lca. 
Added service and cluster networks to seedInfo

### DIFF
--- a/lca-cli/postpivot/postpivot.go
+++ b/lca-cli/postpivot/postpivot.go
@@ -25,6 +25,7 @@ import (
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/dynamic"
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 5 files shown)


lca-cli/postpivot/postpivot.go
```diff
@@ -225,6 +230,48 @@ func (p *PostPivot) recert(ctx context.Context, seedReconfiguration *clusterconf
 	return nil
 }
 
+func (p *PostPivot) setProxyAndProxyStatus(seedReconfig *clusterconfig_api.SeedReconfiguration, seedClusterInfo *seedclusterinfo.SeedClusterInfo) error {
+	if seedReconfig.Proxy != nil != seedClusterInfo.HasProxy {
+		return fmt.Errorf("seedReconfig and seedClusterInfo proxy configuration mismatch")
```
pixelsoccupied, line 235: LGTM from me! 

But on the topic of mismatch and proxy...

It turns turns out that simple `make deploy` during dev loses out on the proxy config with env (HTTP_PROXY, HTTPS_PROXY, and NO_PROXY) even though seed's `HasProxy` variable reports it correctly. This is a mismatch? 

@donpenney helped me figure it out but is this something we should document OR even better add a tiny warning somewhere to note this? 

This is obviously super dev specific (it's always a bundle for customer?) issue but would nice to link some docs explaining where these vars are inject from + this gotcha! wdyt?

donpenney: We should add some checks in a couple of places, in a separate PR... if proxy is configured but LCA doesn't have the corresponding env vars set, it would mean LCA isn't really running with proxy, presumably due to dev install method:
- Reject seedgen attempt outright, as a system validation check
- LCA init can generate an error log. Possibly a health check, but maybe reject IBU attempt completely as well

donpenney (commented):
/lgtm

PR #473, opened 2024-05-01
CNF-12459: seedgen: Exclude openvswitch and ovn db files from seed image

donpenney:
This update excludes the openvswitch and ovn database files from the seed image. Any data in these files from the seed SNO will likely end up needing to be pruned, with new config discovered and added. By excluding the files altogether, the databases are built from scratch with discovery, avoiding the need for any pruning or handling of stale info from the seed.

In addition, when creating the var.tgz, files from /var/lib in the MCD currentconfig file (ie. files managed by MCD) are now explicitly included ahead of processing any wildcard exclusions. This avoids risking the MCP being degraded due to missing files, such as the /var/lib/ovn-ic/etc/enable_dynamic_cpu_affinity file which is only present in certain configurations.


romfreiman:
Cc @tsorya ptal

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	github.com/coreos/go-semver v0.3.1
+	github.com/coreos/ignition/v2 v2.15.0
```
tsorya, line 7: if we are ok with pushing ignition package, it should be very helpful

lca-cli/seedcreator/seedcreator.go
```diff
@@ -302,6 +305,40 @@ func (s *SeedCreator) stopServices() error {
 	return nil
 }
 
+// getVarLibFilelist parses the MCD currentconfig to get the list of managed files under /var/lib
+func (s *SeedCreator) getVarLibFilelist() ([]string, error) {
+	var filelist []string
+	varlibRegex := regexp.MustCompile(`^/var/lib/`)
```
tsorya, line 311: why only var/lib ? i believe it is easier but same issue can be with etc/ too

tsorya:
@donpenney i really like the idea as i believe it can be an entrypoint to allow changing service and cluster networks, just interesting if it influences installation time or not?

lca-cli/seedcreator/seedcreator.go
donpenney: The etc tarball is based on the ostree admin config-diff and doesn't (currently) include any broad wildcard exclusions, so wouldn't face the same issue as the var tarball.

donpenney:
> @donpenney i really like the idea as i believe it can be an entrypoint to allow changing service and cluster networks, just interesting if it influences installation time or not?

I don't have any timing info right now for comparison, but it should be a slight improvement at least, since there would be nothing to prune. Unless the comparison was against a seed image built from the same node, such that there were no changes and nothing to prune? But the rebuild is fairly quick in any case.

donpenney:
> > @donpenney i really like the idea as i believe it can be an entrypoint to allow changing service and cluster networks, just interesting if it influences installation time or not?
> 
> I don't have any timing info right now for comparison, but it should be a slight improvement at least, since there would be nothing to prune. Unless the comparison was against a seed image built from the same node, such that there were no changes and nothing to prune? But the rebuild is fairly quick in any case.

@tsorya @browsell There's very little difference in timing. The pruning doesn't take very long, just from checking a couple of tests. Mostly, this will get rid of those stale seed references.

PR #476, opened 2024-05-02
OCPBUGS-32495: setup stateroot with a job and other prep stage refactoring

pixelsoccupied:
This PR aims to make the Prep stage error handling more robust and overall hopefully simpler. To achieve this...the following things were done (at a high level): 
- Bulk of the stateroot setup functions (e.g pull image, calling rpmostree) are now in a cli and are called as a k8s job. 
  - this allows the workload to be always "watched" by the job and any unexpected events (any pods deleted or system reboot) can be captured instantly. Currently any type of unexpected events will mark the stage as `failed` (no retries). 
- Precache job is treated as a generic Job and handled similar how we are orchestrating stateroot setup job, allowing for reduced code. 
-  As part of an effort to identify any data that needs to be persistent, the Prep stage will now annotate the IBU cr with a (NEW) warning if it detects that current cluster is missing certain CRDs during validation. The annotation with example value may look like this: 
  ```metadata:
    annotations:
      lca.openshift.io/warn-extramanifest-cm-missing-crd: '{"velero.io/v1, Kind=Backup":"acm-klusterlet2","velero.io/v1,
        Kind=Restore":"acm-klusterlet3"}'
  ```

Additionally note:

- There's a new function (with required rbacs) to grab running Job's pod log 
- AutoRollbackOnFailure is now a pointer so suppress false positive  `can not change spec.autoRollbackOnFailure while ibu is in progress`. Without it, the struct would auto initialize and wouldn't allow IBU to update with new annotation at runtime. edit: using patch (instead of update) should avoid auto init but regardless it's good have the struct as pointer. 
-  No more tasks/go routines during the prep stage. 
- Suppress successful exec podman call logs in preache to reduce noise 

/cc @Missxiaoguo @donpenney @jc-rh @browsell


internal/prep/prep_stateroot_setup_job.go
```diff
@@ -0,0 +1,167 @@
...
+							EnvFrom:         manager.EnvFrom,
+							SecurityContext: manager.SecurityContext, // this is needed for podman
+							VolumeMounts:    manager.VolumeMounts,
+							Resources:       manager.Resources,
+						},
+					},
+					HostPID:            lcaDeployment.Spec.Template.Spec.HostPID, // this is needed for rpmostree
+					ServiceAccountName: lcaDeployment.Spec.Template.Spec.ServiceAccountName,
+					RestartPolicy:      corev1.RestartPolicyNever,
+					Volumes:            lcaDeployment.Spec.Template.Spec.Volumes,
```
browsell, line 99: Need to add requests, see precache job, https://github.com/openshift-kni/lifecycle-agent/blob/main/internal/precache/helper.go#L213

pixelsoccupied: Yep it's right under the containers list (L93). We are grabbing everything from LCA basically....when rendered i see 
```
        resources:
          requests:
            cpu: 100m
            memory: 20Mi
```

lca-cli/cmd/staterootSetupIBU.go
```diff
@@ -0,0 +1,104 @@
+package cmd
```
donpenney, line 1: nit: suggest using IBU as a prefix, rather than suffix... ie. ibuStaterootSetup.go, ibuStaterootSetupCmd, etc

pixelsoccupied: Updated to `ibuStaterootSetup.go`

pixelsoccupied:
/hold 
will patch instead of update during cleanup as well

Missxiaoguo (commented):
Thanks for your great work!!
/lgtm

internal/extramanifest/extramanifest.go
```diff
@@ -188,26 +174,43 @@ func (h *EMHandler) ValidateExtraManifestConfigmaps(ctx context.Context, content
...
+			defaultErrMsg := fmt.Sprintf("failed to apply manifest with dryrun: %v", err.Error())
 			var groupDiscoveryErr *discovery.ErrGroupDiscoveryFailed
-			if errors.As(err, &groupDiscoveryErr) || meta.IsNoMatchError(err) {
-				return NewEMUnknownCRDError(errMsg)
+			switch {
+			case errors.As(err, &groupDiscoveryErr) || meta.IsNoMatchError(err): // Capture unknown CRD issue
+				key := manifest.GroupVersionKind().String()
+				if _, exists := unknownCRDs[key]; !exists {
+					unknownCRDs[key] = manifest.GetName()
+					h.Log.Info("unknown CRD", "CRD", key)
```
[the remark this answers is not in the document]
pixelsoccupied: Keeping unresolved! Will update in the next one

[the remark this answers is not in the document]
pixelsoccupied: https://github.com/openshift-kni/lifecycle-agent/pull/517

PR #478, opened 2024-05-02
Pass `fips=1` flag to new stateroot if MC is fips

omertuc:
LCA is responsible for constructing the kernel arguments for the new
stateroot. Right now it only copies the KernelArguments field of the
MachineConfig, but it should also look at the `FIPS` field and make
sure to add the `fips=1` kernel argument when it's true

Solves [OCPBUGS-29360](https://issues.redhat.com/browse/OCPBUGS-29360)

### DIFF
--- a/internal/prep/prep.go
+++ b/internal/prep/prep.go
@@ -68,6 +68,16 @@ func buildKernelArgumentsFromMCOFile(path string) ([]string, error) {
 			args[2*i+1] = string(val)
 		}
 	}
+
+	if mc.Spec.FIPS {
+		args = append(args,
+			"--karg-append", "fips=1",
+			// This is needed because /boot is on a separate partition https://access.redhat.com/solutions/137833
+			// TODO: Should we have this regardless of FIPS?
+			"--karg-append", "boot=LABEL=boot",
+		)
+	}
+
 	return args, nil
 }
 


omertuc:
Testing

internal/prep/prep.go
```diff
@@ -68,6 +68,15 @@ func buildKernelArgumentsFromMCOFile(path string) ([]string, error) {
 			args[2*i+1] = string(val)
 		}
 	}
+
+	if mc.Spec.FIPS {
+		args = append(args, "--karg-append", "fips=1")
+
+		// This is needed because /boot is on a separate partition https://access.redhat.com/solutions/137833
+		// TODO: Should we have this regardless of FIPS?
+		args = append(args, "--karg-append", "boot=LABEL=boot")
```
romfreiman, line 77: I wonder if we should copy some kargs by default. Not sure though how to identify what should be copied and what not

omertuc: Yeah, I feel like this FIPS thing is just a symptom of some bigger thing we're doing wrong. But I'm not familiar enough with ostree/MCO to say what exactly. We should probably consult with someone who knows

internal/prep/prep.go
```diff
@@ -68,6 +68,16 @@ func buildKernelArgumentsFromMCOFile(path string) ([]string, error) {
 			args[2*i+1] = string(val)
 		}
 	}
+
+	if mc.Spec.FIPS {
+		args = append(args,
+			"--karg-append", "fips=1",
+			// This is needed because /boot is on a separate partition https://access.redhat.com/solutions/137833
+			// TODO: Should we have this regardless of FIPS?
```
donpenney, line 76: Probably would be good to include always, for completeness at least?

PR #481, opened 2024-05-04
MGMT-16807: set correct no proxy and proxyStatus

tsorya:
We should create NoProxy and proxyStatus exactly same way as cluster network operator. Took CNO function and made it match lca. Added service and cluster networks to seedInfo


tsorya:
tested e2e

donpenney (approved):

PR #484, opened 2024-05-06
OCPBUGS-32939: Drop "Add check for localVolume presence in CleanupStaleBackups"

leo8a:
Dropping commit [094f0d5](https://github.com/openshift-kni/lifecycle-agent/commit/094f0d50e13302a4edcea62769ca520353196264) since it is not needed anymore. There was a race in the [ensureBackupsDeleted](https://github.com/openshift-kni/lifecycle-agent/blob/main/internal/backuprestore/backup.go#L566-L590) function - fixed in [PR#471](https://github.com/openshift-kni/lifecycle-agent/pull/471/files) - that was provoking [OCPBUGS-32939](https://issues.redhat.com//browse/OCPBUGS-32939).

Moreover, a check to ignore existing "localvolume" Backup CRs during `Prep` stage should not be added, given that LCA will try to create a Backup CR with the same name and this could provoke issues during the pre-pivot step.

```
{"name":"upgrade"}, "namespace": "", "name": "upgrade", "reconcileID": "c269c50a-e1b6-4f6d-adc6-026541173f50", "error": "failed to run pre pivots without errors: error while handling backup: error while starting or tracking backup: failed to create backup: backups.velero.io \"localvolume\" already exists"}
```

### DIFF
--- a/internal/backuprestore/backup.go
+++ b/internal/backuprestore/backup.go
@@ -504,11 +504,9 @@ func (h *BRHandler) CleanupStaleBackups(ctx context.Context, backups []*velerov1
 			continue
 		}
 
+		// Check cluster ID label of existingBackup
 		labels := existingBackup.GetLabels()
-		hasLocalVolumes := h.searchForLocalVolumes(backup)
-
-		// Check cluster ID label and localVolume presence in existingBackup CR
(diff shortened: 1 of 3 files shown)


donpenney (commented):
/lgtm

PR #487, opened 2024-05-06
OCPBUGS-30341: seedgen: Exclude /etc/hostname from seed image

donpenney:


tsorya:
why we should exclude it?

tsorya:
I think we are doing something wrong actually, from my point of view we set hostname in a wrong way and removing this file should just create strange hostname. 
IMO we should set hostname in postpivot operation with hostnamectl command and not just write it to file. 
https://github.com/openshift-kni/lifecycle-agent/blob/main/lca-cli/postpivot/postpivot.go#L743
should be change to 
hostnamectl set-hostname new-hostname
or we need to restart systemd-hostnamed service in order to pickup /etc/hostname file

tsorya:
Lets push this one and i will add another PR that setups hostname in a right way and not just set it in file

PR #488, opened 2024-05-06
`Dockerfile` should set `CGO_ENALBED=1` for FIPS compliance

omertuc:
As far as I can tell, there was no particular reason to set `CGO_ENABLED=0` in the first place, so I'm changing it to `CGO_ENABLED=1` to ensure FIPS compliance, according to https://access.redhat.com/articles/openshift_fips_compliance_faq .

Downstream Dockerfile changes might be required as well

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,11 @@ COPY internal internal
 COPY main main
 COPY utils utils
 
-# Build the binaries
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a -o build/manager main/main.go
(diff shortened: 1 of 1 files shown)


pixelsoccupied:
/cc @rauhersu 
Dockerfile change

omertuc:
```
/usr/local/bin/manager: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /usr/local/bin/manager)
/usr/local/bin/manager: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /usr/local/bin/manager)
```

omertuc:
Oh, it's a multi-stage Dockerfile and the second stage glibc version isn't compatible with the first stage

omertuc:
Should we even bother with fixing this upstream? I'm closing this PR, let's only do this downstream

PR #493, opened 2024-05-07
MGMT-17305: Wipe installation disk on IBI

tsorya:
[MGMT-17305](https://issues.redhat.com//browse/MGMT-17305): Wipe installation disk on IBI
1. Adding installation disk cleanup feature
2. In order to make it work as full flow in ibipreparation moved bash commands to it
3. Adding skip cleanup option


eranco74 (commented):
/lgtm
Please move all these options to configuration file and a config struct.
I guess it all started with few parameters but now it seems way too much parameters to pass around. 
/approve

PR #496, opened 2024-05-10
CNF-12652: bundle: Update csv annotations in LCA bundle

leo8a:
Sets `proxy-aware` and `fips-compliant` annotations to true in the LCA bundle, this will notify users installing LCA from the official catalogs about these operator's capabilities.

image

/cc @rauhersu @omertuc @jc-rh


jc-rh:
should `disconnected` be true too?

rauhersu:
/hold

We will need extra midstream changes to claim 'fips compliant'. I will let you know when ready. Additionally, I believe we should mimic those changes as part of the upstream build, to keep them both as symmetrical as we can.

leo8a:
> should `disconnected` be true too?

yes, we do support LCA deployments in `disconnected` envs (also tested by QE)

PR #497, opened 2024-05-10
Update api version to v1

jc-rh:


api/seedgenerator/v1/groupversion_info.go
```diff
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
pixelsoccupied, line 14: Probably best to do with a different PR...but do you think we should make the api dir a bit more uniform? e.g `./api/v1/seedgenerator` and `./api/v1/imagebasedupgrade`? That way under /v1

jc-rh: Probably makes more sense to have x/v1 and y/v1?

pixelsoccupied: > imagine we won't always upversion them together

Good point... x/v1 and y/v1 makes more sense.

> whole api under pkg

Yeah i think the pkg generally indicates "safe to import externally"?. But in this case IMO api is fine since it's more of [operator-sdk way](https://sdk.operatorframework.io/docs/overview/project-layout/#golang) and anyone new coming would probably look for api dir first?

jc-rh: Updated

PR #499, opened 2024-05-11
Use the new recert additional trust bundle options

omertuc:
# Background / Context

Recert recently added ([1], [2]) some options that allow changing the
cluster's trust bundle (it's recommended you read the PRs for more
background about this).

The lifecycle-agent doesn't make use of the new options added to recert

# Solution / Feature Overview

Change the lifecycle-agent to use the new options added to recert

# Implementation Details

Multiple new fields have been added.

- `AdditionalTrustBundle` in `SeedReconfiguration`. This represents the
  trust bundle to be used for seed-reconfiguration. This contains the
  user-ca-bundle contents, the proxy configmap name, and the proxy configmap
  contents.

- `AdditionalTrustBundle` in `SeedClusterInfo`. This represents the
  state of the trust bundle in the seed cluster. This is simply booleans
  indicating the presence or lack there-of of the user-ca-bundle and the
  proxy configmap name (only if it actually has contents, a configmap
  with no contents is considered invalid OCP configuration). This is
  useful for when we want to verify that the seed is compatible with
  our desired `SeedReconfiguration`.

- `RecertConfig` will now use the new `CryptoDirs` and `CryptoFiles` fields
  to specify the directories and files that should be considered part of
  the cluster's crypto material. Along with the `ClusterCustomizationDirs` and
  `ClusterCustomizationFiles` fields that specify the directories and files
  involved in cluster customization. Since these no longer overlap when
  it comes to customizing the trust bundle, we must use these new fields
  instead of the old common `StaticDirs` and `StaticFiles` fields.

[1] https://github.com/rh-ecosystem-edge/recert/pull/110
[2] https://github.com/rh-ecosystem-edge/recert/pull/140

### DIFF
--- a/api/seedreconfig/seedreconfig.go
+++ b/api/seedreconfig/seedreconfig.go
@@ -139,6 +139,8 @@ type SeedReconfiguration struct {
 	// In IBU case data will be taken from the upgraded cluster /etc/chrony.conf file.
 	// In IBI case data will be taken from the user provided configuration.
 	ChronyConfig string `json:"chrony_config,omitempty"`
+
+	AdditionalTrustBundle AdditionalTrustBundle `json:"additionalTrustBundle,omitempty"`
 }
 
 type KubeConfigCryptoRetention struct {
@@ -180,4 +182,16 @@ type Proxy struct {
 	// NoProxy is a comma-separated list of domains and CIDRs for which the proxy should not be used.
 	// +optional
 	NoProxy string `json:"noProxy,omitempty"`
+}
+
+type AdditionalTrustBundle struct {
+	// The contents of the "user-ca-bundle" configmap in the "openshift-config" namepace
+	UserCaBundle string `json:"userCaBundle"`
+
+	// The Proxy CR trustedCA configmap name
+	ProxyConfigmapName string `json:"proxyConfigmapName"`
+
+	// The contents of the ProxyConfigmapName configmap. Must equal
+	// UserCaBundle if ProxyConfigmapName is "user-ca-bundle"
+	ProxyConfigmapBundle string `json:"proxyConfigmapBundle"`
 }
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -124,6 +124,23 @@ func GetSeedImage(c client.Client, ctx context.Context, ibu *ibuv1.ImageBasedUpg
 		return fmt.Errorf("checking seed image compatibility: %w", err)
 	}
 
+	hasUserCaBundle, proxyConfigmapName, err := lcautils.GetClusterAdditionalTrustBundleState(ctx, c)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster additional trust bundle state: %w", err)
+	}
+
+	//nolint:staticcheck //lint:ignore SA9003 // else branch intentionally left empty
+	if seedInfo.AdditionalTrustBundle != nil {
+		if err := checkSeedImageAdditionalTrustBundleCompatibility(*seedInfo.AdditionalTrustBundle, hasUserCaBundle, proxyConfigmapName); err != nil {
+			return fmt.Errorf("checking seed image additional trust bundle compatibility: %w", err)
+		}
+	} else {
+		// For the sake of backwards compatibility, we allow older seed images
+		// that don't have information about the additional trust bundle. This
+		// means that upgrade will fail at the recert stage if there's a
+		// mismatch between the seed and the seed reconfiguration data.
+	}
+
 	return nil
 }
 
@@ -209,6 +226,31 @@ func checkSeedImageProxyCompatibility(seedHasProxy, hasProxy bool) error {
 
 	if !seedHasProxy && hasProxy {
 		return fmt.Errorf("seed image does not have a proxy but the cluster being upgraded does, this combination is not supported")
+	}
+
+	return nil
+}
+
+// checkSeedImageAdditionalTrustBundleCompatibility checks for proxy
+// configuration compatibility of the seed image vs the current cluster. If the
+// seed image has a proxy and the cluster being upgraded doesn't, we cannot
+// proceed as recert does not support proxy rename under those conditions.
+// Similarly, we cannot proceed if the cluster being upgraded has a proxy but
+// the seed image doesn't.
+func checkSeedImageAdditionalTrustBundleCompatibility(seedAdditionalTrustBundle seedclusterinfo.AdditionalTrustBundle, hasUserCaBundle bool, proxyConfigmapName string) error {
+	if seedAdditionalTrustBundle.HasUserCaBundle && !hasUserCaBundle {
+		return fmt.Errorf("seed image has an %s/%s configmap but the cluster being upgraded does not, this combination is not supported",
+			common.OpenshiftConfigNamespace, common.ClusterAdditionalTrustBundleName)
+	}
+
+	if !seedAdditionalTrustBundle.HasUserCaBundle && hasUserCaBundle {
+		return fmt.Errorf("seed image does not have an %s/%s configmap but the cluster being upgraded does, this combination is not supported",
+			common.OpenshiftConfigNamespace, common.ClusterAdditionalTrustBundleName)
+	}
+
+	if seedAdditionalTrustBundle.ProxyConfigmapName != proxyConfigmapName {
+		return fmt.Errorf("seed image's Proxy trustedCA configmap name %q (oc get proxy -oyaml) mismatches cluster's name %q, this combination is not supported",
+			seedAdditionalTrustBundle.ProxyConfigmapName, proxyConfigmapName)
 	}
 
 	return nil
--- a/internal/clusterconfig/clusterconfig.go
+++ b/internal/clusterconfig/clusterconfig.go
@@ -10,7 +10,6 @@ import (
 	v1 "github.com/openshift/api/config/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -34,15 +33,10 @@ import (
 const (
 	manifestDir = "manifests"
 
-	proxyName = "cluster"
-
 	pullSecretName = "pull-secret"
 
 	idmsFileName  = "image-digest-mirror-set.json"
 	icspsFileName = "image-content-source-policy-list.json"
-
-	caBundleCMName   = "user-ca-bundle"
-	caBundleFileName = caBundleCMName + ".json"
 
 	// ssh authorized keys file created by mco from ssh machine configs
 	sshKeyFile = "/home/core/.ssh/authorized_keys.d/ignition"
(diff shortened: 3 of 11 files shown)


api/seedreconfig/seedreconfig.go
```diff
@@ -174,3 +176,15 @@ type Proxy struct {
 	// +optional
 	NoProxy string `json:"noProxy,omitempty"`
 }
+
+type AdditionalTrustBundle struct {
+	// The contents of the "user-ca-bundle" configmap in the "openshift-config" namepace
+	UserCaBundle string `json:"userCaBundle"`
+
+	// The Proxy CR trustedCA configmap nam
```
mresvanis, line 184:
```suggestion
	// The Proxy CR trustedCA configmap name
```

internal/clusterconfig/clusterconfig.go
```diff
@@ -145,9 +136,73 @@ func (r *UpgradeClusterConfigGather) GetKubeadminPasswordHash(ctx context.Contex
 	return kubeadminPasswordHash, nil
 }
 
+type AdditionalTrustBundle struct {
+	// The contents of the "user-ca-bundle" configmap in the "openshift-config" namepace
```
leo8a, line 140: nit:
```suggestion
	// The contents of the "user-ca-bundle" configmap in the "openshift-config" namespace
```

utils/client_helper.go
```diff
@@ -334,6 +336,67 @@ func HasFIPS(ctx context.Context, client runtimeclient.Client) (bool, error) {
 	return machineConfig.Spec.FIPS, nil
 }
 
+func GetAdditionalTrustBundleFromConfigmap(ctx context.Context, client client.Client, configmapName string) (string, error) {
```
leo8a, line 339: `configmapName` not used within this func, any reason to maintain it?

utils/client_helper.go
```diff
@@ -334,6 +336,67 @@ func HasFIPS(ctx context.Context, client runtimeclient.Client) (bool, error) {
...
+
+	if userCaBundleConfigmap.Data[common.CaBundleDataKey] == "" {
+		return "", nil
+	}
+
+	return userCaBundleConfigmap.Data[common.CaBundleDataKey], nil
+}
+
+func GetClusterAdditionalTrustBundleState(ctx context.Context, client client.Client) (bool, string, error) {
+	clusterAdditionalTrustBundle, err := GetAdditionalTrustBundleFromConfigmap(ctx, client, common.ClusterAdditionalTrustBundleName)
```
leo8a, line 362: if `configmapName` removed from above suggestion, we may need to remove this one as well
```suggestion
	clusterAdditionalTrustBundle, err := GetAdditionalTrustBundleFromConfigmap(ctx, client)
```

utils/client_helper.go
```diff
@@ -334,6 +336,67 @@ func HasFIPS(ctx context.Context, client runtimeclient.Client) (bool, error) {
...
+	}
+
+	proxyCaBundle := ""
+	switch proxy.Spec.TrustedCA.Name {
+	case common.ClusterAdditionalTrustBundleName:
+		proxyCaBundle = clusterAdditionalTrustBundle
+	case "":
+		// No proxy trustedCA configmap is set, do nothing
+	default:
+		proxyCaBundle, err = GetAdditionalTrustBundleFromConfigmap(ctx, client, proxy.Spec.TrustedCA.Name)
```
leo8a, line 381: if `configmapName` removed from above suggestion, we may need to remove this one as well
```suggestion
		proxyCaBundle, err = GetAdditionalTrustBundleFromConfigmap(ctx, client)
```

lca-cli/seedcreator/seedcreator.go
```diff
@@ -193,6 +194,20 @@ func (s *SeedCreator) handleServices() error {
 	})
 }
 
+func GetSeedAdditionalTrustBundleState(ctx context.Context, client runtimeclient.Client) (*seedclusterinfo.AdditionalTrustBundle, error) {
```
leo8a, line 197: just wondering if the `GetSeedAdditionalTrustBundleState` funct could be moved under the utils pkg (in the client_helper.go) as well

leo8a (commented):
overall it looks good to me, thanks for adding it

(I've dropped some comments below for your consideration)

internal/recert/recert.go
```diff
@@ -59,6 +71,29 @@ func FormatRecertProxyFromSeedReconfigProxy(proxy, statusProxy *seedreconfig.Pro
...
+
+	if additionalTrustBundle.ProxyConfigmapName != "" {
+		if additionalTrustBundle.ProxyConfigmapBundle == "" {
+			return fmt.Errorf("proxy configmap bundle is empty while proxy configmap name is set")
+		}
+
+		switch additionalTrustBundle.ProxyConfigmapName {
+		case common.ClusterAdditionalTrustBundleName:
+			recertConfig.ProxyTrustedCaBundle = fmt.Sprintf("%s:", common.ClusterAdditionalTrustBundleName)
+		case "":
```
tsorya, line 87: Hmm, this case will not be reachable or i miss something ? You actually validate in line 79 that additionalTrustBundle.ProxyConfigmapName is not empty string?

lca-cli/postpivot/postpivot.go
```diff
@@ -464,14 +460,6 @@ func (p *PostPivot) restoreOadpDataProtectionApplication(ctx context.Context, cl
 	return nil
 }
 
-func (p *PostPivot) copyClusterConfigFiles() error {
-	if err := utils.CopyFileIfExists(path.Join(p.workingDir, common.ClusterConfigDir, path.Base(common.CABundleFilePath)),
```
tsorya, line 468: I believe there will not be new ca bundle files but we still need create it as in IBI we not always build on precaching and i believe we still need to set trusted bundle as possibly it will be required to pull recert image in case it was not precached for some reason

lca-cli/seedcreator/seedcreator.go
```diff
@@ -193,6 +194,20 @@ func (s *SeedCreator) handleServices() error {
...
+	HasUserCaBundle, ProxyConfigmapName, err := utils.GetClusterAdditionalTrustBundleState(ctx, client)
```
tsorya, line 198:
```suggestion
	hasUserCaBundle, proxyConfigmapName, err := utils.GetClusterAdditionalTrustBundleState(ctx, client)
```

utils/client_helper.go
```diff
@@ -334,6 +336,67 @@ func HasFIPS(ctx context.Context, client runtimeclient.Client) (bool, error) {
 	return machineConfig.Spec.FIPS, nil
 }
 
+func GetAdditionalTrustBundleFromConfigmap(ctx context.Context, client client.Client, configmapName string) (string, error) {
+	userCaBundleConfigmap := corev1.ConfigMap{}
+	if err := client.Get(ctx, types.NamespacedName{Name: common.ClusterAdditionalTrustBundleName,
```
tsorya, line 341:
```suggestion
	if err := client.Get(ctx, types.NamespacedName{Name: configmapName,
```

utils/client_helper.go
```diff
@@ -334,6 +336,67 @@ func HasFIPS(ctx context.Context, client runtimeclient.Client) (bool, error) {
...
+		proxyCaBundle = clusterAdditionalTrustBundle
+	case "":
+		// No proxy trustedCA configmap is set, do nothing
+	default:
+		proxyCaBundle, err = GetAdditionalTrustBundleFromConfigmap(ctx, client, proxy.Spec.TrustedCA.Name)
+		if err != nil {
+			return false, "", fmt.Errorf("failed to get additional trust bundle from configmap: %w", err)
+		}
+
+		if proxyCaBundle == "" {
```
tsorya, line 386: Why it actually matters for us? I mean if there is not ca bundle in proxy lets not get it? Or we need to provide better explanation in error we return another way not sure what user should do with it

utils/client_helper.go
omertuc: It should have been use, it not being used is a bug. Will fix

lca-cli/seedcreator/seedcreator.go
omertuc: Can't do that without creating an import loop

internal/recert/recert.go
omertuc: True, fixed

utils/client_helper.go
omertuc: This will probably never happen and I don't even want to spend the time thinking about what would happen if we just set it as empty and how we should behave exactly, so I simply block it altogether. This really doesn't matter, it's just to be safe.

lca-cli/postpivot/postpivot.go
omertuc: Right now `copyClusterConfigFiles` is only called [after](https://github.com/openshift-kni/lifecycle-agent/blob/f331b9d0d2154601f7799dc29660976a4599db1a/lca-cli/postpivot/postpivot.go#L126-L132) recert, is that a bug?

tsorya:
@omertuc how it will work with IBI? i wonder how we will be able to provide all the needed params?

api/seedreconfig/seedreconfig.go
```diff
@@ -181,3 +183,15 @@ type Proxy struct {
 	// +optional
 	NoProxy string `json:"noProxy,omitempty"`
 }
+
+type AdditionalTrustBundle struct {
```
tsorya, line 187: We need to handle it in IBI too as it will break current api and IBIO, need task for it

controllers/prep_handlers.go
```diff
@@ -214,6 +231,31 @@ func checkSeedImageProxyCompatibility(seedHasProxy, hasProxy bool) error {
...
+		return fmt.Errorf("seed image has an %s/%s configmap but the cluster being upgraded does not, this combination is not supported",
+			common.OpenshiftConfigNamespace, common.ClusterAdditionalTrustBundleName)
+	}
+
+	if !seedAdditionalTrustBundle.HasUserCaBundle && hasUserCaBundle {
+		return fmt.Errorf("seed image does not have an %s/%s configmap but the cluster being upgraded does, this combination is not supported",
+			common.OpenshiftConfigNamespace, common.ClusterAdditionalTrustBundleName)
+	}
+
+	if seedAdditionalTrustBundle.ProxyConfigmapName != proxyConfigmapName {
```
tsorya, line 251: Hmm, i wonder why we should not support it? Can't recert just get both names?

internal/clusterconfig/clusterconfig.go
```diff
@@ -162,9 +153,73 @@ func (r *UpgradeClusterConfigGather) GetKubeadminPasswordHash(ctx context.Contex
 	return kubeadminPasswordHash, nil
 }
 
+type AdditionalTrustBundle struct {
```
tsorya, line 156: why not to use seedreconfig.AdditionalTrustBundle ? Seems totally the same or i missed something?

controllers/prep_handlers.go
omertuc: There's already a lot of different cases we must support, supporting also another case where the name of the configmap holding the proxy trust bundle is different between the seed and the new cluster is just too much work (recert would have to delete the new configmap, create a new configmap with the correct name, and change the proxy reference to point to the new configmap) for something that's completely unnecessary. If we'll have users that have different names for this configmap across multiple clusters which they want to update with the same seed, we can consider supporting this, until then, I don't see the point.

api/seedreconfig/seedreconfig.go
```diff
@@ -181,3 +183,15 @@ type Proxy struct {
...
+	// The contents of the "user-ca-bundle" configmap in the "openshift-config" namepace
+	UserCaBundle string `json:"userCaBundle"`
+
+	// The Proxy CR trustedCA configmap name
+	ProxyConfigmapName string `json:"proxyConfigmapName"`
```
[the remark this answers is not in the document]
omertuc: See the commit message of https://github.com/rh-ecosystem-edge/recert/pull/140

PR #500, opened 2024-05-13
OCPBUGS-33536: [IBI] install-rhcos-and-restore-seed.service fails the first time it starts

tsorya:
OCPBUGS-33536: [IBI] install-rhcos-and-restore-seed.service fails the first time it starts
We are taking lca-cli from seed image and sometimes it can fail. Adding dependency for this service on network-online and adding retry to image pull part

### DIFF
--- a/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
+++ b/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
@@ -6,8 +6,10 @@ seed_image=${1:-$SEED_IMAGE}
 authfile=${AUTH_FILE:-"/var/tmp/backup-secret.json"}
 ibi_config=${IBI_CONFIGURATION_FILE:-"/var/tmp/ibi-configuration.json"}
 
-# Copy the lca-cli binary to the host
-podman create --authfile "${authfile}" --name lca-cli "${seed_image}" lca-cli
+# Copy the lca-cli binary to the host, pulling it seed image can sometimes fail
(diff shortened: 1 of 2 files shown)


ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh
```diff
@@ -6,8 +6,8 @@ seed_image=${1:-$SEED_IMAGE}
 authfile=${AUTH_FILE:-"/var/tmp/backup-secret.json"}
 ibi_config=${IBI_CONFIGURATION_FILE:-"/var/tmp/ibi-configuration.json"}
 
-# Copy the lca-cli binary to the host
-podman create --authfile "${authfile}" --name lca-cli "${seed_image}" lca-cli
+# Copy the lca-cli binary to the host, pulling it seed image can sometimes fail
+(while ! podman create --authfile "${authfile}" --name lca-cli "${seed_image}" lca-cli ; do exit;sleep 10;done)
```
donpenney, line 10: Why is this in a subprocess? Also, the `exit;sleep 10` syntax seems odd here. Won't this just exit from the subprocess when it fails?
```suggestion
while ! podman create --authfile "${authfile}" --name lca-cli "${seed_image}" lca-cli ; do sleep 10;done
```

javipolo, line 10: use `until` instead of `while !`

javipolo (approved):

javipolo (approved):

PR #501, opened 2024-05-13
OCPBUGS-33535: export associated storage classes when exporting localvolumes

Missxiaoguo:
The storage classes that are being used in the localvolumes should be exported and restored along with localvolumes during post-pivot as well.

### DIFF
--- a/internal/clusterconfig/lvmconfig.go
+++ b/internal/clusterconfig/lvmconfig.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
 	"github.com/openshift-kni/lifecycle-agent/utils"
 	cp "github.com/otiai10/copy"
+	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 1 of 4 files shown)


internal/clusterconfig/lvmconfig.go
```diff
@@ -94,17 +96,48 @@ func (r *UpgradeClusterConfigGather) fetchLocalVolumes(ctx context.Context, mani
...
+		lvSpec := lv.Object["spec"].(map[string]any)
+		if scDevices, exists := lvSpec["storageClassDevices"].([]any); exists {
+			for _, sc := range scDevices {
+				if scName, exists := sc.(map[string]any)["storageClassName"].(string); exists {
+					scNameSet[scName] = true
+				}
+			}
+		}
+	}
+
```
leo8a, line 123: I'd add a quick check `if len(scNameSet) == 0` then return before the `for` below, but for your consideration

donpenney: The for-loop would do nothing because of the range, correct? Is the length check before a range a common thing to do?

PR #520, opened 2024-05-16
CNF-12403: Introduce Subscription health check

donpenney:
# Background / Context

LCA runs system health checks during SeedGen and IBU to ensure that the cluster is healthy and stable before proceeding.

An unhealthy subscription means there may not be a corresponding CSV. The existing health checks validate the CSV status, but if the CSV doesn't exist, that health check passes. As a result, the user may generate a seed image, for example, without realizing that the desired operators have not actually been installed.

# Solution / Feature Overview

Introducing a Subscription health check gives added protection by ensuring all subscriptions are in a healthy state.

# Implementation Details

This update adds a Subscription check to the set of system health checks. The Subscription check ensures that each subscription CR has no Failed conditions set to true and CatalogSourcesUnhealthy is false, indicating the subscription is healthy.

This check is skipped in the postpivot Upgrade stage handler, as it is expected that the catalog sources will not have been updated at this point, and each subscription would therefore be reported as unhealthy.

### DIFF
--- a/internal/healthcheck/healthcheck.go
+++ b/internal/healthcheck/healthcheck.go
@@ -42,27 +42,13 @@ const (
 	NodeRoleWorker       = "node-role.kubernetes.io/worker"
 
 	SriovNetworkNodeStateNotPresentMsg = "no SriovNetworkNodeStates present"
-
-	// HealthCheckAnnotationX annotations allow the user to skip certain health checks by marking them as Disabled. Currently only available for the subscription check.
-	// HealthCheckAnnotationSubscriptions allows the user to skip the subscription check, for example if they explicitly want to ignore an available update.
-	// Only acceptable value is HealthCheckDisableValue. Any other value is treated as "Enabled".
-	HealthCheckAnnotationSubscriptions = "healthchecks.lca.openshift.io/subscriptions"
-	// HealthCheckDisableValue value that decides if a give healthcheck is disabled
-	HealthCheckDisableValue = "Disabled"
 )
 
 type HealthCheckOptions struct {
 	SkipSubscriptionCheck bool
 }
 
-func HealthCheckOptionsFromAnnotations(annotations map[string]string) (opts HealthCheckOptions) {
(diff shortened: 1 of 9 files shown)


internal/healthcheck/healthcheck.go
```diff
@@ -64,6 +85,13 @@ func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
 		failures = append(failures, err.Error())
 	}
 
+	if !opts.SkipSubscriptionCheck {
```
jc-rh, line 88: I thought subscription check were useful in general. The only bit that may need to be skipped is the upgradePending case?

jc-rh: Oh, never mind. I thought you were checking `state: UpgradePending` vs `state: AtLatestKnown`

donpenney: Subscription checks have been tightened to only consider the Failed and CatalogSourcesUnhealthy conditions

romfreiman:
/hold
again, what will it affect? when will the customers be blocked? why not to verify it only upon seed creation?

donpenney:
> /hold again, what will it affect? when will the customers be blocked? why not to verify it only upon seed creation?

@romfreiman as noted, this PR has been revised to only look for failure conditions. It no longer checks the conditions relating to available updates. It now just checks to ensure that the subscriptions are healthy - ie. catalog sources are healthy, no failed conditions.

internal/healthcheck/healthcheck.go
```diff
@@ -41,7 +44,11 @@ const (
 	SriovNetworkNodeStateNotPresentMsg = "no SriovNetworkNodeStates present"
 )
 
-func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
+type HealthCheckOptions struct {
+	SkipSubscriptionCheck bool
+}
+
+func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger, opts *HealthCheckOptions) error {
```
tsorya, line 51: I think opts is more class param, why not to provide boolean here?

PR #521, opened 2024-05-16
OCPBUGS-33275: extramanifest configmaps validation update

Missxiaoguo:
# Background / Context
LCA performs Kubernetes Dry-run to validate the manifests included in the extraManifests configmaps. Any invalid/webhook badRequest type of error returned from dry-run fails the Prep stage and blocks the upgrade.

Kubernetes Dry-run sends requests to API server, where the server processes the request as if it were real, but without acutally making any modifications to the actual resources. As all API requests are single-resource requests, dryrun doesn't work for resources that depends on eariler resources that are not persisted in the cluster. For example, in a multi-object manifest that includes namespace A and resource A created within the namespace A. Running a dry-run on resource returns the namespace not found error.

The extra manifests are configurations intented to be applied to the new cluster version. Runing dry-run on the old version aims to thoroughly check for possible manifest error before upgrading. However, in cases where extramanifests configmaps contain the dependent resources, the Dry-run failure results in failed Prep, but the failure may be not the case for the new cluster version.

# Solution / Feature Overview
Not only in the case of dependent resources, but there may also be other dry-run errors that occur only on the old cluster version.  The dry-run validation should be a soft validation. The errors such as invalid or webhook badRequest type of error from dry-run are considered warnings. This also includes cases where CRDs are missing from the old version and dependent namespace does not exist on the old version but is also not found in the configmaps. The warnings do not block upgrade. Instead, the Prep status condition could be updated with a warning message and the failure details can be provided in an IBU annotation. The other type of errors such as random chars, missing resource Kind, resource ApiVersion, resource name or the presence of disallowed resource types, are treated as critical errors which will fail the Prep.

# Implementation Details

# Other Information
- If user uses the ZTP/TALM to run IBU with extra manifests configmaps, the prep stage policy should include `message` check with "Prep stage completed successfully" which would prevent CGU to complete if there's validation warning

- Log error in the upgrade handler for the error that causes the upgrade to fail.

/cc @donpenney @pixelsoccupied @leo8a

### DIFF
--- a/controllers/upgrade_handlers.go
+++ b/controllers/upgrade_handlers.go
@@ -73,7 +73,6 @@ func (r *ImageBasedUpgradeReconciler) handleUpgrade(ctx context.Context, ibu *ib
 	origStaterootBooted, err := r.RebootClient.IsOrigStaterootBooted(ibu)
 
 	if err != nil {
-		//todo: abort handler? e.g delete desired stateroot
 		utils.SetUpgradeStatusFailed(ibu, err.Error())
 		return doNotRequeue(), nil
 	}
@@ -142,6 +141,7 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade)
 		if backuprestore.IsBRFailedValidationError(err) ||
 			backuprestore.IsBRFailedError(err) {
 
+			u.Log.Error(err, "Failed to handle backups")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			return doNotRequeue(), nil
 		}
@@ -169,6 +169,7 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade)
 
 	if err := u.exportOadpConfigurationAndRestore(ctx, ibu, staterootVarPath); err != nil {
 		if backuprestore.IsBRFailedError(err) || backuprestore.IsBRFailedValidationError(err) {
+			u.Log.Error(err, "Failed to export OADP configuration and restores")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			return doNotRequeue(), nil
 		}
@@ -183,6 +184,7 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade)
 	u.Log.Info("Writing extra-manifests into new stateroot")
 	if err := u.extractAndExportExtraManifests(ctx, ibu, staterootVarPath); err != nil {
 		if extramanifest.IsEMFailedError(err) {
+			u.Log.Error(err, "Failed to export manifests")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			return doNotRequeue(), nil
 		}
@@ -225,8 +227,7 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade)
 	u.Recorder.Event(ibu, v1.EventTypeNormal, "Reboot", "System will now reboot for upgrade")
 	err = u.RebootClient.RebootToNewStateRoot("upgrade")
 	if err != nil {
-		//todo: abort handler? e.g delete desired stateroot
-		u.Log.Error(err, "")
+		u.Log.Error(err, "Failed to reboot to new stateroot")
 		utils.SetUpgradeStatusFailed(ibu, err.Error())
 		return doNotRequeue(), nil
 	}
@@ -362,6 +363,7 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade
 	err := u.BackupRestore.EnsureOadpConfiguration(ctx)
 	if err != nil {
 		if backuprestore.IsBRStorageBackendUnavailableError(err) {
+			u.Log.Error(err, "Failed to ensure OADP configuration")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to missing DataProtectionApplication: %s", err))
 			return doNotRequeue(), nil
@@ -379,6 +381,7 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade
 	err = u.ExtraManifest.ApplyExtraManifests(ctx, common.PathOutsideChroot(extramanifest.PolicyManifestPath))
 	if err != nil {
 		if extramanifest.IsEMFailedError(err) {
+			u.Log.Error(err, "Failed to apply policy manifests")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to failure applying policy manifests: %s", err))
 			return doNotRequeue(), nil
@@ -392,9 +395,10 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade
 		u.Log.Error(updateErr, "failed to update IBU CR status")
 	}
 
-	err = u.ExtraManifest.ApplyExtraManifests(ctx, common.PathOutsideChroot(extramanifest.ExtraManifestPath))
+	err = u.ExtraManifest.ApplyExtraManifests(ctx, common.PathOutsideChroot(extramanifest.CmManifestPath))
 	if err != nil {
 		if extramanifest.IsEMFailedError(err) {
+			u.Log.Error(err, "Failed to apply config manifests")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to failure applying config manifests: %s", err))
 			return doNotRequeue(), nil
@@ -413,6 +417,7 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade
 	if err != nil {
 		// Restore failed
 		if backuprestore.IsBRFailedError(err) {
+			u.Log.Error(err, "Failed to handle restore")
 			utils.SetUpgradeStatusFailed(ibu, err.Error())
 			u.autoRollbackIfEnabled(ibu, fmt.Sprintf("Rollback due to restore failure: %s", err))
 			return doNotRequeue(), nil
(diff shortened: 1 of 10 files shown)


internal/extramanifest/extramanifest.go
```diff
@@ -179,79 +192,149 @@ func ApplyExtraManifest(ctx context.Context, dc dynamic.Interface, restMapper me
...
-func (h *EMHandler) ValidateExtraManifestConfigmaps(ctx context.Context, content []ibuv1.ConfigMapRef, ibu *ibuv1.ImageBasedUpgrade) error {
+// ValidateExtraManifestConfigmaps validates IBU extramanifest configmaps and returns validation warning and error.
+// It iterates over all manifests included in the configmaps and perform Kubernetes Dry-run.
+// Errors from Dry-run apply are considered warnings. This also includes cases where CRDs are missing from the current
+// stateroot and dependent namespace does not exist on the current stateroot but is also not found in the configmaps.
+// Other validation failures are treated as errors, such as random chars, missing resource Kind, resource ApiVersion,
+// resource name or the presence of disallowed resource types
+func (h *EMHandler) ValidateExtraManifestConfigmaps(ctx context.Context, content []ibuv1.ConfigMapRef) (string, error) {
+	var errs = []string{}     // validation errors
+	var warnings = []string{} // validation warnings
```
leo8a, line 203:
```suggestion
	var errs []string     // validation errors
	var warnings []string // validation warnings
```
safer empty slice declaration (according to [styleguide](https://google.github.io/styleguide/go/decisions#nil-slices))

internal/extramanifest/extramanifest.go
```diff
@@ -47,18 +47,26 @@ import (
...
 
+//+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;watch;create;
+
 const (
-	ExtraManifestPath           = "/opt/extra-manifests"
-	PolicyManifestPath          = "/opt/policy-manifests"
-	WarnAnnotationUnknownCRDKey = "lca.openshift.io/warn-extramanifest-cm-unknown-crd"
+	ExtraManifestPath                        = "/opt/extra-manifests"
+	PolicyManifestPath                       = "/opt/policy-manifests"
+	ExtraManifestValidationWarningAnnotation = "lca.openshift.io/extramanifest-cm-validation-warning"
```
leo8a, line 55:
```suggestion
	Path                        = "/opt/extra-manifests"
	PolicyManifestPath                       = "/opt/policy-manifests"
	ValidationWarningAnnotation = "lca.openshift.io/extramanifest-cm-validation-warning"
```
optional nit: exported constants should not start with package name (see [styleguide](https://google.github.io/styleguide/go/decisions#constant-names))

in this way we could call it like `extramanifest.ValidationWarningAnnotation` instead of `extramanifest.ExtraManifestValidationWarningAnnotation` (repeated)

totally get it if we don't want to go over this refact in this PR, just suggesting it

internal/extramanifest/extramanifest.go
```diff
@@ -179,79 +192,149 @@ func ApplyExtraManifest(ctx context.Context, dc dynamic.Interface, restMapper me
...
+						h.Log.Info(fmt.Sprintf("Created namespace %s for dryrun", ns.Name))
+						fakedNamespacesForDryrun[ns.Name] = true
+					} else {
+						return validationWarn, fmt.Errorf("failed to query namespace %s: %w", ns, err)
+					}
+				}
+			}
+
+			// Apply manifest with dryRun
+			if err := ApplyExtraManifest(ctx, h.DynamicClient, h.Client.RESTMapper(), manifest, dryRun); err != nil {
```
leo8a, line 300:
```suggestion
			if err := ApplyExtraManifest(ctx, h.DynamicClient, h.Client.RESTMapper(), manifest, true); err != nil {
```
`dryRun` is always true, maybe could be simplified (if so, there's a declaration above that could be removed)

leo8a (commented):
nice work Angie, thanks

dropped some minor comments below, mainly related to code style guidelines

jc-rh:
/cc @sudomakeinstall2 

> If user uses the ZTP/TALM to run IBU with extra manifests configmaps, the prep stage policy should include message check with "Prep stage completed successfully" which would prevent CGU to complete if there's validation warning

controllers/upgrade_handlers.go
```diff
@@ -226,7 +229,7 @@ func (u *UpgHandler) PrePivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade)
 	err = u.RebootClient.RebootToNewStateRoot("upgrade")
 	if err != nil {
 		//todo: abort handler? e.g delete desired stateroot
```
pixelsoccupied, line 231: i think this todo can be removed right?

docs/image-based-upgrade.md
```diff
@@ -291,9 +291,11 @@ oc patch imagebasedupgrade upgrade -n openshift-lifecycle-agent --type='json' -p
...
 
 - Perform the following validations:
-  - If the oadpContent is populated, validate that the specified configmap has been applied and is valid
-  - If the extraManifests is populated, validate that the specified configmap has been applied and is valid
-    - If a required CRD is missing from the current stateroot, a warning message will be included in the IBU CRs with annotation `lca.openshift.io/warn-extramanifest-cm-unknown-crd`.
+  - If the `oadpContent` is populated, validate that the specified configmap has been applied and is valid
+  - If the `extraManifests` is populated, validate that the specified configmap has been applied and is valid
+    - Validation errors from Dry-run apply are treated as warnings. This also includes cases where CRDs are missing from the current stateroot and dependent namespace does not exist on the current stateroot but is also not found in the configmaps.
+    LCA doesn't block the upgrade due to validation warnings. Instead, it updates the Prep status condition with a warning message and annotates IBU with the annotation `lca.openshift.io.extramanifest-cm-validation-warning`, providing detailed information about the failures
```
pixelsoccupied, line 297: What do you think about updating annotation to something like
```
extra-manifest.lca.openshift.io/warn-validation-error
```

Follows the recommended pattern `subsystem.lca.ocp.io/name`  and matches our other ones we have (autorollback and upcoming HC one too)

controllers/upgrade_handlers.go
```diff
@@ -428,7 +435,7 @@ func (u *UpgHandler) PostPivot(ctx context.Context, ibu *ibuv1.ImageBasedUpgrade
 
 	if err := u.RebootClient.DisableInitMonitor(); err != nil {
 		// Don't fail the upgrade on failure here, just log it
-		u.Log.Error(err, "unable to disable LCA init monitor")
+		u.Log.Error(err, "Unable to disable LCA init monitor")
```
pixelsoccupied, line 438: i think all errors are recommended to start with lowercase letter?

internal/extramanifest/extramanifest.go
```diff
@@ -179,79 +192,149 @@ func ApplyExtraManifest(ctx context.Context, dc dynamic.Interface, restMapper me
...
-			case errors.As(err, &groupDiscoveryErr) || meta.IsNoMatchError(err): // Capture unknown CRD issue
-				key := manifest.GroupVersionKind().String()
-				if _, exists := unknownCRDs[key]; !exists {
-					unknownCRDs[key] = manifest.GetName()
-					h.Log.Info("unknown CRD", "CRD", key)
+			// Verify disallowed resource
+			if kinds, exists := unSupportedResources[mApiVersion]; exists {
+				if lo.Contains(kinds, mKind) {
+					errs = append(errs, fmt.Sprintf("%s[%s] in extramanifests is not allowed", mKind, mName))
+					continue
```
pixelsoccupied, line 263: Ah nice..so we now collecting all the errors before returning yeah? (unlike prev we were returning early?)

internal/extramanifest/extramanifest.go
```diff
@@ -179,79 +192,149 @@ func ApplyExtraManifest(ctx context.Context, dc dynamic.Interface, restMapper me
...
+	if len(warnings) == 0 && len(errs) == 0 {
+		h.Log.Info("ExtraManifests configmaps are validated", "configmaps", content)
+		return "", nil
 	}
 
-	h.Log.Info("ExtraManifests configmaps are validated", "configmaps", content)
-	return nil
+	if len(warnings) > 0 {
+		validationWarn = strings.Join(warnings, "; ")
+		h.Log.Info(fmt.Sprintf("[WARNING] ExtraManifests configmaps validation: %s", validationWarn))
```
pixelsoccupied, line 321: I wish we had `.log.Warning` option!

internal/extramanifest/extramanifest.go
```diff
@@ -179,79 +192,149 @@ func ApplyExtraManifest(ctx context.Context, dc dynamic.Interface, restMapper me
...
+			// Add to namespace set when manifest is Namespace
+			if mKind == "Namespace" {
+				nsManifestsSet[mName] = true
+				continue
+			}
+
+			// Check if manifest's namespace exists on cluster before running dryrun
+			if mNamespace != "" && mNamespace != "default" {
+				ns := &corev1.Namespace{}
+				if err := h.Client.Get(ctx, types.NamespacedName{Name: mNamespace}, ns); err != nil {
```
pixelsoccupied, line 276: What do you think about moving the fakeNS creation + deletion moved out of ValidateExtraManifestConfigmaps? It was a bit hard to follow

internal/extramanifest/extramanifest.go
```diff
@@ -179,79 +192,149 @@ func ApplyExtraManifest(ctx context.Context, dc dynamic.Interface, restMapper me
 	return nil
 }
 
-func (h *EMHandler) ValidateExtraManifestConfigmaps(ctx context.Context, content []ibuv1.ConfigMapRef, ibu *ibuv1.ImageBasedUpgrade) error {
+// ValidateExtraManifestConfigmaps validates IBU extramanifest configmaps and returns validation warning and error.
+// It iterates over all manifests included in the configmaps and perform Kubernetes Dry-run.
+// Errors from Dry-run apply are considered warnings. This also includes cases where CRDs are missing from the current
```
pixelsoccupied, line 197: I think should say something like "Most errors from Dry-run apply..." ?

pixelsoccupied (commented):
Just tiny comments! But lgtm otherwise

controllers/upgrade_handlers.go
donpenney: This would be an error log, not an error that can be wrapped

internal/extramanifest/extramanifest.go
Missxiaoguo: Exactly, collect all errors and warnings.

internal/extramanifest/extramanifest.go
Missxiaoguo: yeah, I think I should make it more accurate

internal/extramanifest/extramanifest.go
Missxiaoguo: Certainly, that's good suggestion to simplify this validation function

Missxiaoguo: Unable to completely moving the creation part out but I hope this version is easier to follow

Issue #424, opened 2024-04-01
unrecognized import path "vbom.ml/util"


pamoedom:
Environment:
~~~
$ cat /etc/redhat-release 
Fedora release 38 (Thirty Eight)

$ rpm -qa | grep golang
golang-src-1.21.8-1.fc38.noarch
golang-1.21.8-1.fc38.x86_64
golang-bin-1.21.8-1.fc38.x86_64
~~~

Issue:
~~~
$ make cli-build 
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/controller-tools/cmd/controller-gen@v0.13.0
go: warning: "all" matched no packages
go: creating new go.mod: module tmp
Downloading sigs.k8s.io/kustomize/kustomize/v5@v5.1.1
go: warning: "all" matched no packages
go mod tidy
go: downloading golang.org/x/crypto v0.21.0
go: downloading golang.org/x/sys v0.18.0
go: downloading golang.org/x/net v0.21.0
go: downloading golang.org/x/term v0.18.0
go: github.com/openshift/machine-config-operator@v0.0.1-0.20201023110058-6c8bd9b2915c requires
	github.com/openshift/library-go@v0.0.0-20200831114015-2ab0c61c15de requires
	vbom.ml/util@v0.0.0-20180919145318-efcd4e0f9787: unrecognized import path "vbom.ml/util": https fetch: Get "https://vbom.ml/util?go-get=1": dial tcp: lookup vbom.ml on 127.0.0.53:53: no such host
make: *** [Makefile:107: common-deps-update] Error 1
~~~

Workaround:
~~~
$ echo "replace vbom.ml/util => github.com/fvbommel/util v0.0.0-20180919145318-efcd4e0f9787" >> go.mod
$ go mod vendor
$ make cli-build
~~~
Regards.

leo8a:
below is the current workaround I use in my env, FWIW

```
GOPROXY=https://proxy.golang.org/,direct make cli-build
```

PR #522, opened 2024-05-17
Update to golang 1.21 and k8s api 0.29.4

jc-rh:
# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


leo8a:
nice, wondering if this bump would fix https://github.com/openshift-kni/lifecycle-agent/issues/424

jc-rh:
> nice, wondering if this bump would fix #424

I am not sure. I use go 1.21 locally, never saw the issue.

rauhersu:
/hold

We will introduce the midstream changes prior to this MR

Cc: @fontivan

PR #523, opened 2024-05-19
MGMT-17652: Possibly improve etcd performance through defrag/compaction

tsorya:
# Background / Context
 When reconfiguring a seed image, we have to perform a lot of etcd delete operations. This leaves etcd is a fragmented state, which could possibly harm its performance when its later actually used for OCP and thus lengthen the IBX duration.
Running defrag command allows us to minimize defrag state in etcd db as example 

Without running defrag on seed i got 68% if difference between db size and size in use
`[root@seed core]# oc exec etcd-seed -c etcd -n openshift-etcd -- etcdctl endpoint status --cluster -w json | jq '.[].Status|"dbSize: " + (.dbSize|tostring) + ", dbSizeInUse: " + (.dbSizeInUse|tostring) + ", (dbSize-dbSizeInUse)/dbSize => " + ((.dbSize - .dbSizeInUse)/.dbSize*100|tostring)+"%"'
"dbSize: 107253760, dbSizeInUse: 33562624, (dbSize-dbSizeInUse)/dbSize => 68.70727515753295%"
`
With running defrag on seed creation , i got 0% difference
`[root@seed core]# oc exec etcd-seed -c etcd -n openshift-etcd -- etcdctl endpoint status --cluster -w json | jq '.[].Status|"dbSize: " + (.dbSize|tostring) + ", dbSizeInUse: " + (.dbSizeInUse|tostring) + ", (dbSize-dbSizeInUse)/dbSize => " + ((.dbSize - .dbSizeInUse)/.dbSize*100|tostring)+"%"'
"dbSize: 83079168, dbSizeInUse: 83079168, (dbSize-dbSizeInUse)/dbSize => 0%"
`

# Solution / Feature Overview
Running etcd defrag command after running recert on seed creation

# Implementation Details

# Other Information

### DIFF
--- a/lca-cli/ops/ops.go
+++ b/lca-cli/ops/ops.go
@@ -1,6 +1,7 @@
 package ops
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -12,10 +13,12 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/sirupsen/logrus"
+	etcdClient "go.etcd.io/etcd/client/v3"
+
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
 	"github.com/openshift-kni/lifecycle-agent/internal/recert"
 	"github.com/openshift-kni/lifecycle-agent/utils"
-	"github.com/sirupsen/logrus"
 )
 
 var podmanRecertArgs = []string{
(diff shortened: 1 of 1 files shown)


lca-cli/ops/ops.go
```diff
@@ -115,7 +118,29 @@ func (o *ops) ForceExpireSeedCrypto(recertContainerImage, authFile string, hasKu
 	if err := recert.CreateRecertConfigFileForSeedCreation(recertConfigFile, hasKubeAdminPassword); err != nil {
 		return fmt.Errorf("failed to create %s file: %w", recertConfigFile, err)
 	}
-	if err := o.RecertFullFlow(recertContainerImage, authFile, recertConfigFile, nil, nil); err != nil {
+
+	// Run post recert operation to defragment etcd after recert tool is run
+	// this should allow for a more efficient etcd work after installation
+	// should not fail seed creation for now
```
hexfusion, line 124: You will also need to run a compaction to have the defrag be useful. This is an example of using the latest revision. You could perform this during seed and it would improve I/O latency. Tradeoff being defrag is blocking but if you did it periodically during seed the duration would be minimal.

```
	resp, _ := client.KV.Get(ctx, "key")
	latestRevision := resp.Header.Revision

	_, err = client.Compact(ctx, latestRevision)
```

hexfusion: Note: Compaction is not required in OCP/K8s because it is done periodically every 5m by the ApiServer.

tsorya: do you think we need to run it here? I assume it will run 5 minutes after we will start the cluster, no? 
I mean that if in seed we run it every 5 minutes it is compacted already so we just need to defrag it. From what i see in doc it say compact and then defrag. IMO don't see a reason to add compact here in this case but i really can miss something here

hexfusion: If the goal is to improve seed timing, then consider it. Post seed apiserver and etcd-operator manage this as you noted.  Compacted db will read/write faster in general so this is the problem we are solving for.

hexfusion: Talking with @tsorya I misunderstood seed step as not having an apiserver. In which case this would be useful. If the cluster is otherwise running then compaction would not be necessary step.

hexfusion (approved):

donpenney (approved):

donpenney:
thanks @hexfusion @tsorya

PR #524, opened 2024-05-19
OCPBUGS-32553: Adding ChronyConfig param to seed reconfiguration

tsorya:
This param will allow to provide data to /etc/chrony.conf that will be used in order to oeverride seed data in the same file in order to allow setup of user ntp servers.

Currently i added it as full configuration taken from IBU cluster that will be written after upgrade, though maybe better to split configuration and provide it as additional ntp server as assisted installer uses it?
In assisted we have default chrony.conf and allow to provide additional ntp servers

Default :
`pool 0.rhel.pool.ntp.org iburst
driftfile /var/lib/chrony/drift
makestep 1.0 3
rtcsync
logdir /var/log/chrony`

Adding default ntp servers like :
`server <provided ntp server> iburst`

Adding logic to provide chrony.conf to recert
Restarting chronyd in post-pivot phase after running recert in order to
apply new chrony.conf

depends on https://github.com/rh-ecosystem-edge/recert/pull/145


donpenney (commented):
/lgtm

jc-rh:
@tsorya @omertuc do guys have a plan on when to propagate the recert change to the release branches?

tsorya:
> @tsorya @omertuc do guys have a plan on when to propagate the recert change to the release branches?

@jc-rh was pushed to v0 lets push this one, test and then push to 4.15 i think

donpenney:
> > @tsorya @omertuc do guys have a plan on when to propagate the recert change to the release branches?
> 
> @jc-rh was pushed to v0 lets push this one, test and then push to 4.15 i think

If there's a dependent recert change, we'll need it in the 4.14, 4.15, and 4.16 release branches before we can resync the LCA release branches to pick up this PR

tsorya:
> > > @tsorya @omertuc do guys have a plan on when to propagate the recert change to the release branches?
> > 
> > 
> > @jc-rh was pushed to v0 lets push this one, test and then push to 4.15 i think
> 
> If there's a dependent recert change, we'll need it in the 4.14, 4.15, and 4.16 release branches before we can resync the LCA release branches to pick up this PR

Sounds like, though i would like to test it before backporting everything. I tested it manually but i would like to have QE verify too. 
So we need to push to master, test it, get qe verified and than backport it in recert and lca

PR #422, opened 2024-03-27
MGMT-17364: Allow providing subnet for IBI installation

tsorya:
[MGMT-17364](https://issues.redhat.com//browse/MGMT-17364): Allow providing subnet for IBI installation
Adding new api field MachineNetwork it will be used to set nodeip hint that will point nodeip-configuration to the right ip

PR #525, opened 2024-05-21
MGMT-16515: in order to set machine network in node ip hint file in post pivot we should take it from target

tsorya:
[MGMT-16515](https://issues.redhat.com//browse/MGMT-16515): in order to set machine network in node ip hint file in post pivot we should take it from target

# Background / Context
[MGMT-16515](https://issues.redhat.com//browse/MGMT-16515): in order to set machine network in node ip hint file in post pivot we should take it from target

# Solution / Feature Overview
This pr adds logic to take machine network from install-config of upgraded sno and set it in seed reconfiguration api.
Post pivot will use this field and set it in nodeip hint file in order require from node-configuration service to set kubelet ip from the machine network we provided

# Implementation Details

# Other Information

### DIFF
--- a/utils/client_helper.go
+++ b/utils/client_helper.go
@@ -4,14 +4,18 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net"
 	"strings"
+
+	"github.com/samber/lo"
+	log "github.com/sirupsen/logrus"
 
 	"github.com/openshift-kni/lifecycle-agent/api/seedreconfig"
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
 	ocp_config_v1 "github.com/openshift/api/config/v1"
 	mcv1 "github.com/openshift/api/machineconfiguration/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
-	"github.com/samber/lo"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -77,6 +81,7 @@ type ClusterInfo struct {
 	MirrorRegistryConfigured bool
 	ClusterNetworks          []string
 	ServiceNetworks          []string
+	MachineNetwork           string
 }
 
 func GetClusterInfo(ctx context.Context, client runtimeclient.Client) (*ClusterInfo, error) {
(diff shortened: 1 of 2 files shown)


utils/client_helper.go
```diff
@@ -145,6 +153,28 @@ func getNodeInternalIP(node corev1.Node) (string, error) {
...
+		if _, _, err := net.ParseCIDR(mn.CIDR); err != nil {
+			return "", fmt.Errorf("MachineNetwork has an invalid CIDR: %s", mn.CIDR)
+		}
+		if _, ipnet, err := net.ParseCIDR(mn.CIDR); err == nil && ipnet.Contains(net.ParseIP(nodeIp)) {
+			return mn.CIDR, nil
+		}
+	}
+
+	// retuning empty string if no match found in order to be backward compatible as there is a possibility that
+	// that we upgrade cluster that was previously upgraded with lc-agent version that didn't set new machineNetwork
```
eranco74, line 172:
```suggestion
	// that we upgrade cluster that was previously upgraded with lca-agent version that didn't set new machineNetwork
```

eranco74, line 172: Add logging, with the reason we return empty (node IP doesn't match?, no machine network?, machine network is invalid...)

eranco74 (commented):
Will this play nice with https://github.com/openshift-kni/lifecycle-agent/pull/422?

tsorya:
> Will this play nice with #422?

this is exactly the idea

utils/client_helper.go
```diff
@@ -145,6 +156,31 @@ func getNodeInternalIP(node corev1.Node) (string, error) {
...
+	for _, mn := range installConfig.Networking.MachineNetwork {
+		if _, _, err := net.ParseCIDR(mn.CIDR); err != nil {
+			return "", fmt.Errorf("machineNetwork has an invalid CIDR: %s", mn.CIDR)
+		}
+		if _, ipnet, err := net.ParseCIDR(mn.CIDR); err == nil && ipnet.Contains(net.ParseIP(nodeIp)) {
+			return mn.CIDR, nil
+		}
+	}
+
+	log.Warnf("failed to find machine network in <%s> for node ip: %s. Returning empty string",
```
eranco74, line 174: The order is wrong

eranco74 (commented):
Just a nit

utils/client_helper.go
tsorya: oops

eranco74 (commented):
/lgtm

PR #526, opened 2024-05-21
OCPBUGS-34010: In case proxy status was set there is no reason to update proxy in post pivot phase (IBU flow)

tsorya:
OCPBUGS-34010: In case proxy status was set there is no reason to update proxy in post pivot phase (IBU flow)

# Background / Context
In IBU flow we take proxy and proxy status from upgraded cluster and there is no need to update those values in post-pivot

failure in proxy installation


tsorya:
tested by @mcornea

eranco74 (commented):
/approve

PR #527, opened 2024-05-21
docs: Update seedgen prereqs section

donpenney:
# Background / Context

Seed generation doc has been updated to expand the pre-requisites section with additional information about seed SNO criteria and alignment with target SNOs.

### DIFF
--- a/docs/seed-image-generation.md
+++ b/docs/seed-image-generation.md
@@ -28,14 +28,42 @@ The Lifecycle Agent provides orchestration of IBU Seed Image generation via the 
 
 The seed SNO configuration has some pre-requisites:
 
-- The CPU topology must align with the target SNO(s).
-  - Same number of cores.
-  - Same tuned performance configuration (ie. reserved CPUs).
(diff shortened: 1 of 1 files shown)


leo8a:
maybe this PR can also drop a note on the dnsmaq configuration section like below... or remove it since the minimum supported version of ACM should be +4.10

```
> [!NOTE]
> The dnsmasq configuration section is only relevant when the version of ACM used is below v4.9.2.
```

donpenney:
> maybe this PR can also drop a note on the dnsmaq configuration section like below... or remove it since the minimum supported version of ACM should be +4.10
> 
> ```
> > [!NOTE]
> > The dnsmasq configuration section is only relevant when the version of ACM used is below v2.9.2.
> ```

Thanks @leo8a . I added a note to the line that the required config is automatically included if deployed via ACM >= 2.9.2. We could consider removing the check from seedgen completely, but would likely need to include some analogous check of the dnsmasq config for the case where the node has not been deployed via ACM at all.

PR #539, opened 2024-05-24
Update OLM api to 0.23

jc-rh:
# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


jc-rh:
Replaces #538

PR #541, opened 2024-05-27
OCPBUGS-32493: improve prep stage error handling to allow for unexpected apiserver or network issues

pixelsoccupied:
# Background / Context

Cluster can expect to run into unexpected network issues and it's up to the controller to eventually reach desired state

We ran into network outage around the time when `Prep` was trying to `Get` a `deployment` CR for the first time. Because it was the first time...controller needed to reach out to the API server in order to init a new informer. At which point we see 
```
failed to get API group resources: unable to retrieve the complete list of server APIs: apps/v1: Get "https://[fd02::1]:443/apis/apps/v1": dial tcp [fd02::1]:443: connect: connection refuse
```

# Solution / Feature Overview

Controller should allow time for the API server outage to heal and retry until then

# Implementation Details

- We are now requeue with error for functions that needs to reach to API server. For `validateIBUSpec` we are being selective with the requeue since it's more likely from user rather than network
- Health checks are now moved out to be called at every reconcile. During normal operations reconcile still fast and when the cluster is healing it will simply requeue before checking the `Prep` stage state. 
- Increase the logs to make it easier to trace 

# Other Information

- As part of the increased logging...I added a way to update the logger name which should now allow us quickly identify stage specific logs (can push to it another PR but this really helped me with this bug) 

Tip

To force an API outage 
```
# bring down
mv /etc/kubernetes/manifests/kube-apiserver-pod.yaml /tmp/kube-apiserver-pod.yaml

# bring up
mv /tmp/kube-apiserver-pod.yaml /etc/kubernetes/manifests/kube-apiserver-pod.yaml
```

(At this you need to check with crictl for logs)

/cc @donpenney


pixelsoccupied:
/hold
validateIBUSpec also may also return network error

pixelsoccupied:
/unhold

Also requeue on when IBU spec is invalid. Requeue with error has the built-in rate limiter which eventually slows the requeue time so it’s not going to thrash the system if anything

There are too many different types of network error that we not generically handle and from from IBU's side we could potentially add a custom error type but that will be error prone (making sure every function returns a custom func when not making network calls) and devs in the future may forget to follow a custom pattern...

pixelsoccupied:
/cc @Missxiaoguo 

As discussed we are being a but more selective with  `validateIBUSpec` since the likely reason of error from it is cause of user error

But we are still requeueing with known network error (about 4 types)...and they of course cover the error we are focusing here i.e  `..connection refused` due to API server down

pixelsoccupied:
/hold

doing a bit of testing with the new change

PR #543, opened 2024-05-28
MGMT-17949: Proxy params for installation iso

tsorya:
[MGMT-17949](https://issues.redhat.com//browse/MGMT-17949): Proxy params for installation iso
Setting proxy setting as part of installation service in installation iso

# Background / Context
As part of installation flow on installation iso we need to pull seed image and precache seed images.  We need to add proxy settings to installation service in order for it to work 

# Solution / Feature Overview
Set proxy params as part of installation service in installation iso


eranco74 (commented):
/lgtm

PR #544, opened 2024-05-28
MGMT-17011: Allow providing additional trusted bundle to installation iso.

tsorya:
[MGMT-17011](https://issues.redhat.com//browse/MGMT-17011): Allow providing additional trusted bundle to installation iso.

# Background / Context
In order to install ibi with proxy and mirror registry we should be able to provide additional trusted bundle for installation iso. 

# Solution / Feature Overview
Provided additional trusted bundle file is read and added to installation ignition to pki/anchors same way as assisted installer is doing it in discovery ignition

# Implementation Details

# Other Information

### DIFF
--- a/api/ibiconfig/ibiconfig.go
+++ b/api/ibiconfig/ibiconfig.go
@@ -16,8 +16,11 @@ const (
 // IBIPrepareConfig or aka ImageBasedInstallConfig is the API for specifying configuration
 // for the image-based installer.
 type IBIPrepareConfig struct {
-	SSHPublicKeyFile string `json:"sshPublicKeyFile,omitempty"`
-	RHCOSLiveISO     string `json:"rhcosLiveIso,omitempty"`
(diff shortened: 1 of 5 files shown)


api/ibiconfig/ibiconfig.go
```diff
@@ -16,8 +16,11 @@ const (
 // IBIPrepareConfig or aka ImageBasedInstallConfig is the API for specifying configuration
 // for the image-based installer.
 type IBIPrepareConfig struct {
-	SSHPublicKeyFile string `json:"sshPublicKeyFile,omitempty"`
-	RHCOSLiveISO     string `json:"rhcosLiveIso,omitempty"`
+	// Special installation iso params
```
eranco74, line 19: what's special about ssh public key?

eranco74 (commented):
/approve
Just a nit

PR #546, opened 2024-05-29
MGMT-16859: Provide an option to configure mirror registry in installation iso

tsorya:
[MGMT-16859](https://issues.redhat.com//browse/MGMT-16859): Provide an option to configure mirror registry in installation iso

# Background / Context
We need to be able to pull images in installation iso, in case of mirror registry we must configure registry.conf file in order to be able to pull those images

# Solution / Feature Overview
Add mirror-registry cli param for ib-cli in order to get path to mirror registry config file

# Implementation Details

# Other Information

### DIFF
--- a/ib-cli/cmd/createiso.go
+++ b/ib-cli/cmd/createiso.go
@@ -49,6 +49,7 @@ var (
 	httpsProxy            string
 	noProxy               string
 	additionalTrustBundle string
+	mirrorRegistry        string
 )
 
 func addFlags(cmd *cobra.Command) {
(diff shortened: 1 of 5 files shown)


ib-cli/cmd/createiso.go
```diff
@@ -49,6 +49,7 @@ var (
 	httpsProxy            string
 	noProxy               string
 	additionalTrustBundle string
+	mirrorRegistry        string
```
eranco74, line 52: Consider renaming to imageContentSources

eranco74 (commented):
Looks good, just consider renaming it everywhere to imageContentSources

PR #548, opened 2024-05-29
OCPBUGS-32493: make all client calls retriable with middleware

pixelsoccupied:
# Background / Context

This builds on https://github.com/openshift-kni/lifecycle-agent/pull/541.  We now have middleware that will do the retries, making all the client call implicitly retriable.

As long as clients are configured to include it during initialization...the middleware will intercept all calls and retry exponentially for max 3 mins if any of the error conditions are met.

/cc @donpenney

### DIFF
--- a/internal/common/utils.go
+++ b/internal/common/utils.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+
 	"io"
 	"os"
 	"path/filepath"
@@ -32,6 +33,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/go-logr/logr"
+	ibuv1 "github.com/openshift-kni/lifecycle-agent/api/imagebasedupgrade/v1"
 	cp "github.com/otiai10/copy"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -40,8 +42,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	ibuv1 "github.com/openshift-kni/lifecycle-agent/api/imagebasedupgrade/v1"
 )
 
 // TODO: Need a better way to change this but will require relatively big refactoring
@@ -129,12 +129,6 @@ func GetStaterootOptOpenshift(staterootPath string) string {
 	return filepath.Join(staterootPath, "var", OptOpenshift)
 }
 
-// FuncTimer check execution time
-func FuncTimer(start time.Time, name string, r logr.Logger) {
-	elapsed := time.Since(start)
-	r.Info(fmt.Sprintf("%s took %s", name, elapsed))
-}
-
 func IsConflictOrRetriable(err error) bool {
 	return apierrors.IsConflict(err) || apierrors.IsInternalError(err) || apierrors.IsServiceUnavailable(err) || net.IsConnectionRefused(err)
 }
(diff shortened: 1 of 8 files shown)


pixelsoccupied:
/hold
will update the existing retry func instead

main/main.go
```diff
@@ -138,7 +136,9 @@ func main() {
 
 	mux := &sync.Mutex{}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	cfg := ctrl.GetConfigOrDie()
+	cfg.Wrap(common.RetryMiddleware(log)) // allow all client calls to be retriable
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
```
Missxiaoguo, line 141: I see that ctrl.options has an option to configure a custom HTTPclient, but I think that's just another way to configure it
```
		Client: client.Options{
			HTTPClient: &http.Client{
				Transport: <your-custom-transport>,
			},
		},
```

internal/common/utils.go
```diff
@@ -243,3 +239,50 @@ func GenerateDeleteOptions() *client.DeleteOptions {
...
+		resp, err = r.transport.RoundTrip(req)
+		if err != nil && IsRetriable(err) {
+			r.log.Info("WARNING: retrying", "req.url", req.URL.String(), "error", err)
+		}
+		return err //nolint:wrapcheck
+	}); errRetryOnError != nil {
+		r.log.Info("WARNING: retrying failed", "error", errRetryOnError)
+	}
+
+	return resp, err //nolint:wrapcheck
```
Missxiaoguo, line 287: Should this be 'return resp, errRetryOnError'?

main/main.go
```diff
@@ -138,7 +136,9 @@ func main() {
 
 	mux := &sync.Mutex{}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	cfg := ctrl.GetConfigOrDie()
+	cfg.Wrap(common.RetryMiddleware(log)) // allow all client calls to be retriable
```
Missxiaoguo, line 140: Should this be added to dynamicClient and clientset as well?

internal/common/utils.go
```diff
@@ -243,3 +239,50 @@ func GenerateDeleteOptions() *client.DeleteOptions {
...
+
+// RetryRoundTripper include anything here that's useful during middleware processing
+type RetryRoundTripper struct {
+	transport http.RoundTripper
+	backoff   wait.Backoff
+	log       logr.Logger
+}
+
+// RetryMiddleware pass this into your client to make it retriable with configured backoff
+func RetryMiddleware(logger logr.Logger) func(rt http.RoundTripper) http.RoundTripper {
```
Missxiaoguo, line 251: Maybe RetryTransportMiddleware is more clear?

main/main.go
pixelsoccupied: Yep! Also `config.Wrap()` seems to be a nice abstraction which allowed us to include the same middleware to our other clients (dynamic and clienset) pretty easily without worrying about any underlying issues/config like including the default Transporter at the end of the chain etc etc

main/main.go
pixelsoccupied: Added

internal/common/utils.go
pixelsoccupied: Yeah you're right errRetryOnError and err should be the exact same value (at least with the impl they have now)...but I have it like this intentionally in this case. 

I just didn't want to pass on the error from `retry.OnError` instead just logging it...just so that from caller's perspective as if nothing changed (with or without retry). I'm open to changing though no worries. Wdyt?

internal/common/utils.go
pixelsoccupied: Hmm intentionally didnt add transport as it's actually a WrapperFunc that gets returned and used by Wrap()...

How about `RetryWrapMiddleware` instead?

internal/common/utils.go
Missxiaoguo: Yeah, right, they should be the same value. Just at first glance, my initial reation is to return errRetryOnError since that's the error from the retry function.

pixelsoccupied:
/hold 
testing the changes e2e

pixelsoccupied:
/unhold

Will remove all the existing wrapped calls in a separate but everything is working well! Just a FYI the post-pivot dynamicclient middleware update will be picked up once a new seed is generated (but of course upgrade will go through regardless)

Missxiaoguo:
I also tried this on my env. It worked well! Thanks for your great work! @pixelsoccupied 
/lgtm

PR #549, opened 2024-05-30
MGMT-17989: Support static network configuration in installation iso

tsorya:
[MGMT-17989](https://issues.redhat.com//browse/MGMT-17989): Support static network configuration in installation iso
This change allows to provide nmstate configuration that will be applied on installation iso.
In case nmstate config was provided network configuration service will be added, this service run nmstatectl and applies provided config


tsorya:
was pushed as part of https://github.com/openshift-kni/lifecycle-agent/pull/551

PR #551, opened 2024-06-02
MGMT-16402: Move to new api for ibi create installation iso

tsorya:
# Background / Context
 [MGMT-16402](https://issues.redhat.com//browse/MGMT-16402): Move to new api for ibi create installation iso

As part of moving to openshift installer in 4.17 we want to have same api in 4.16 for ibi-cli

That's why we need to reformat current logic and to provide configuration file and not params. 

# Solution / Feature Overview
[MGMT-16402](https://issues.redhat.com//browse/MGMT-16402): Move to new api for ibi create installation iso
1. Adding new api struct ImageBasedInstallConfig
2. Remove butane and start using ignition package
3. Change mirror registry from string to ImageDigestSources

# Implementation Details

# Other Information


ib-cli/installationiso/mirror-registry.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal registries.conf data: %w", err)
+	}
+
+	return marshalled, nil
+}
+
+// MergedMirrorSets consolidates a list of ImageDigestSources so that each
+// source appears only once.
+func mergedMirrorSets(sources []ibiconfig.ImageDigestSource) []ibiconfig.ImageDigestSource {
```
eranco74, line 50: Can we import this code from somewhere (assisted)?

eranco74 (commented):
/lgtm

ib-cli/installationiso/mirror-registry.go
tsorya: not really, this is installer code. 
Importing installer and assisted is very painful.
This code will be removed once we move to installer so lets have it as temporary solution

eranco74:
The job is failing because the API changes require [these](https://github.com/rh-ecosystem-edge/ib-orchestrate-vm/pull/75/files) changes.
/lgtm
/approve

PR #553, opened 2024-06-05
MGMT-18049: Refactor ibi code to use assisted-image-service in order to embed ignition into iso

tsorya:
[MGMT-18049](https://issues.redhat.com//browse/MGMT-18049): Refactor ibi code to use assisted-image-service in order to embed ignition into iso

# Background / Context
Currently we embed ignition into iso by using coreos-install container. We want to remove usage of required containers in order to minimize requirements for mirroring

We should move to assisted-image-service library that allows to do the same 

# Solution / Feature Overview

# Implementation Details

# Other Information


eranco74 (commented):
/lgtm
Add UT for error cases

PR #554, opened 2024-06-05
NO-ISSUE: handle case for static ip configuration without provide node ip (IBI) 

tsorya:
NO-ISSUE: In order to handle case for static ip configuration and without provided nodeip (IBI flow) we need to restart newtork manager before running nodeip configuration service.

# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


leo8a:
```
error: failed to get network 'ib-orchestrate-1'
error: Network not found: no network with matching name 'ib-orchestrate-1' 
```
/test ibi-e2e-flow

javipolo (approved):

PR #555, opened 2024-06-05
OCPBUGS-32493: remove redundant retry wrappers

pixelsoccupied:
# Background / Context

This builds on https://github.com/openshift-kni/lifecycle-agent/pull/548 to clean up the existing explicitly wrapped client calls since they are redundant 

/cc @jc-rh  @Missxiaoguo

### DIFF
--- a/controllers/seedgen_controller.go
+++ b/controllers/seedgen_controller.go
@@ -27,8 +27,6 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/client-go/util/retry"
-
 	"github.com/openshift-kni/lifecycle-agent/controllers/utils"
 	"github.com/openshift-kni/lifecycle-agent/internal/common"
 	"github.com/openshift-kni/lifecycle-agent/internal/healthcheck"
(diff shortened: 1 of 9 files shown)


pixelsoccupied:
/hold

we tested and found that the way .Update was wrapped to retry was basically unused since we must retrieve the latest CR first before retrying in case of conflict e.g see [here](https://github.com/kubernetes-sigs/controller-runtime/blob/2e9781e9fc6054387cf0901c70db56f0b0a63083/pkg/envtest/crd.go#L266-L272) 

I'm also removing them as having them is really just delaying the error currently

controllers/seedgen_controller.go
```diff
@@ -597,18 +596,14 @@ func (r *SeedGeneratorReconciler) restoreSeedgenCRIfNeeded(ctx context.Context,
 	// Save status as the seedgen structure gets over-written by the create call
 	// with the result which has no status
 	status := seedgen.Status
-	if err := common.RetryOnConflictOrRetriable(retry.DefaultBackoff, func() error {
-		return client.IgnoreAlreadyExists(r.Client.Create(ctx, seedgen)) //nolint:wrapcheck
```
Missxiaoguo, line 601: Is it intended to remove the alreadyExists check?

pixelsoccupied: Nope! Deleted too much. Brought it back!

PR #558, opened 2024-06-06
CNF-12806: IBU: Inspect seed image before pulling

donpenney:
# Background / Context

LCA uses labels on the seed image for certain config information, which allows validation against the seed image to ensure it is compatible.

The image labels can be inspected prior to pulling the seed image, in order to do validation prior to the pull. Since the seed image is more than 1G, doing this ahead of the pull can save downloading the image in the case where the validation would fail.

# Implementation Details

Using `skopeo inspect` to remotely access the image labels, rather than `podman inspect`, we can validate the seed image using the labels before pulling the image.

### DIFF
--- a/controllers/prep_handlers.go
+++ b/controllers/prep_handlers.go
@@ -73,12 +73,18 @@ func GetSeedImage(c client.Client, ctx context.Context, ibu *ibuv1.ImageBasedUpg
 	}
 	log.Info("Successfully pulled seed image", "image", ibu.Spec.SeedImageRef.Image)
 
-	labels, err := getLabelsForSeedImage(ibu.Spec.SeedImageRef.Image, ops)
+	return nil
+}
+
+// validateSeedImageConfig retrieves the labels for the seed image without downloading the image itself, then validates
+// the config data from the labels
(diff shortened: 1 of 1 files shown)


pixelsoccupied:
@donpenney looks like the last update I pushed the wrong check 
```
2024-06-07T00:14:21.910582148Z 2024-06-07T00:14:21Z	INFO	controllers.ImageBasedUpgrade.ibu-manager-client	Setting up retry middleware
2024-06-07T00:14:21.926695060Z 2024-06-07T00:14:21Z	INFO	setup	Saved IBU CR found, restoring ...
2024-06-07T00:14:21.930108925Z 2024-06-07T00:14:21Z	ERROR	setup	unable to initialize IBU CR	{"error": "failed to delete IBU during restore: imagebasedupgrades.lca.openshift.io \"upgrade\" not found"}
2024-06-07T00:14:21.930108925Z main.main
2024-06-07T00:14:21.930108925Z 	/opt/app-root/src/main/main.go:185
2024-06-07T00:14:21.930108925Z runtime.main
2024-06-07T00:14:21.930108925Z 	/usr/lib/golang/src/runtime/proc.go:267
```

This should fix the CI https://github.com/openshift-kni/lifecycle-agent/pull/560

PR #560, opened 2024-06-07
OCPBUGS-32493: fix wrong error check

pixelsoccupied:
# Background / Context

Accidentally included the wrong error check during wrapped called cleanup.  (`client.IgnoreNotFound` should have been replaced with `k8serrors.IsNotFound` instead it was set to `k8serrors.IsAlreadyExists`)


donpenney:
Please update the PR description to add a little info to the template, and add a Jira to the title.

PR #561, opened 2024-06-07
OCPBUGS-34089: Add container storage mountpoint checks

donpenney:
This update adds checks to IBU and seedgen to ensure:
- container storage is setup shared, ie. is a mountpoint
- the mountpoint target for the seed image is the same as the mountpoint target on the running cluster

# Background / Context

IBU requires container storage setup such that it is shared between stateroots. The configuration of the shared storage is up to the user, but the mountpoint target (eg. `/dev/disk/by-partlabel/varlibcontainers`) must be the same on both seed and target SNO.

If the seed image is generated on an SNO with a different container storage mountpoint target, the upgraded SNO may fail to boot, as the mountpoint target would not exist.

# Solution / Feature Overview

In order to prevent such a configuration mismatch, LCA has been updated to include checks for the container storage setup.
- If container storage is not setup shared, ie. a mountpoint:
  - A seedgen request will be rejected
  - IBU Prep stage will fail
- If the container storage config in the seed image does not align with the running cluster, the IBU Prep stage will fail

# Implementation Details

LCA parses the var-lib-containers.mount service unit to determine the mountpoint target, as defined by the `What` config variable. This path is added to the seed config information, which is included in the seed image label. This label is checked during IBU Prep to ensure that it matches the configuration on the running cluster.

### DIFF
--- a/main/main.go
+++ b/main/main.go
@@ -202,6 +202,12 @@ func main() {
 	extraManifest := &extramanifest.EMHandler{
 		Client: mgr.GetClient(), DynamicClient: dynamicClient, Log: log.WithName("ExtraManifest")}
 
+	containerStorageMountpointTarget, err := op.GetContainerStorageTarget()
+	if err != nil {
+		setupLog.Error(err, "unable to get container storage mountpoint target")
+		os.Exit(1)
+	}
+
 	// a simple in-cluster client-go based client, useful for getting pod logs
 	// as runtime-controller client currently doesn't support pod sub resources
 	// https://github.com/kubernetes-sigs/controller-runtime/issues/452#issuecomment-792266582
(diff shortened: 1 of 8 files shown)


main/main.go
```diff
@@ -202,6 +202,13 @@ func main() {
 	extraManifest := &extramanifest.EMHandler{
 		Client: mgr.GetClient(), DynamicClient: dynamicClient, Log: log.WithName("ExtraManifest")}
 
+	containerStorageMountpointTarget, err := op.GetContainerStorageTarget()
+	if err != nil {
+		setupLog.Error(err, "unable to get container storage mountpoint target")
+		os.Exit(1)
+	}
+	setupLog.Info("DPENNEY", "containerStorageMountpointTarget", containerStorageMountpointTarget)
```
browsell, line 210: nit. cleanup

donpenney: thanks.. fixed

PR #564, opened 2024-06-10
NO-ISSUE: in case node ip  was not provided, dnsmasq condifuration should run after setting node ip

tsorya:
NO-ISSUE: in case node ip  was not provided, dnsmasq condifuration should run after setting node ip

# Background / Context
If node ip was not provided as part of IBI flow we run node-ip configuration service that should set node ip and only after that we should run dnsmasq configuration 

# Solution / Feature Overview

# Implementation Details

# Other Information


eranco74 (commented):
/lgtm
/approve

PR #592, opened 2024-07-03
[release-4.15] OCPEDGE-1089: chore: use rhel9 for the CI build root

jeff-roche:
Cherry-pick https://github.com/openshift-kni/lifecycle-agent/pull/590 to release-4.15

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.21-openshift-4.16
+  tag: rhel-9-release-golang-1.21-openshift-4.16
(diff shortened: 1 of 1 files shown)


.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.21-openshift-4.16
+  tag: rhel-9-release-golang-1.21-openshift-4.16
```
donpenney, line 4: If this is backported, this should probably be a 4.15 image ref?

jeff-roche: yeah potentially. It was 4.16 before so I didn't change it but don't mind doing the update

jeff-roche: There is no builder for `rhel9-golang-1.21-openshift-4.15`

donpenney (approved):

PR #619, opened 2024-08-06
[release-4.16] CNF-13064: IBU: Add automated container storage cleanup to Prep stage

donpenney:
This is a cherry-pick of https://github.com/openshift-kni/lifecycle-agent/pull/594

/cc @jc-rh


internal/imagemgmt/imagemgmt.go
```diff
@@ -0,0 +1,278 @@
+package imagemgmt
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+	"syscall"
+
+	"github.com/go-logr/logr"
+	"github.com/openshift-kni/lifecycle-agent/lca-cli/ops"
```
tsorya, line 10: lets move it to next block

internal/imagemgmt/imagemgmt.go
```diff
@@ -0,0 +1,278 @@
...
+		if lo.Contains(inUse, ref) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// GetRemovalCandidates gets the list of images sorted by creation timestamp, filtering out in-use and pinned images
+func (c *ImageMgmtClient) GetRemovalCandidates() (removalCandidates []imageMgmtImageInfo, rc error) {
```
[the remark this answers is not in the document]
tsorya: Can't we get list of images from ocp and not touch them? Not sure though how to handle application jobs

internal/imagemgmt/imagemgmt.go
```diff
@@ -0,0 +1,278 @@
...
+// CheckDiskUsageAgainstThreshold gets the disk usage for the specified path and checks whether it exceeds the specified threshold
+func (c *ImageMgmtClient) CheckDiskUsageAgainstThreshold(thresholdPercent int) (bool, error) {
+	var stat syscall.Statfs_t
+	if err := syscallStatfs(c.StoragePath, &stat); err != nil {
+		return false, fmt.Errorf("statfs failed for %s: %w", c.StoragePath, err)
+	}
+
+	// Get disk usage percentage
+	usedpct := int(((stat.Blocks - stat.Bavail) * 100) / stat.Blocks)
+	c.Log.Info(fmt.Sprintf("DEBUG: Disk usage of %s is currently %d%%", c.StoragePath, usedpct))
```
tsorya, line 90: C.Log.Debug?

internal/imagemgmt/imagemgmt.go
```diff
@@ -0,0 +1,278 @@
...
+		rc = fmt.Errorf("unable to parse crictl ps command output: %w", err)
+		return
+	}
+
+	for _, container := range crictlContainers.Containers {
+		inUse = lcautils.AppendToListIfNotExists(inUse, container.Image.Image)
+	}
+
+	// Get the list of containers via podman
+	output, err = c.Executor.Execute("podman", "ps", "-a", "--format", "json")
```
tsorya, line 114: why podman? There should not be images running with podman in ocp

internal/imagemgmt/imagemgmt.go
```diff
@@ -0,0 +1,278 @@
...
+		if image.Pinned {
+			pinned = lcautils.AppendToListIfNotExists(pinned, image.Id)
+		}
+	}
+
+	return
+}
+
+// isImageUsed compares the specified image info against the list of in-use images to check whether the image ID, tag, or digest is in the in-use list
+func isImageUsed(inUse []string, image imageMgmtImageInfo) bool {
```
tsorya, line 158: I wonder if we can just try to remove image and maximum it will not succeed if they are in use?

internal/imagemgmt/imagemgmt.go
```diff
@@ -0,0 +1,278 @@
...
+
+		removalCandidates = append(removalCandidates, image)
+	}
+
+	return
+}
+
+// CleanupUnusedImages iterates through the image removal candidates,
+// deleting in sets of 5 until the container storage disk usage threshold is met
+func (c *ImageMgmtClient) CleanupUnusedImages(thresholdPercent int) error {
```
tsorya, line 238: I wonder not easier to set threshold to 70% and just run 
`crictl rmi --prune` ? I mean that in 10 minutes those images will be not relevant at all as this node will be rebooted in new ocp and this will require much less logic

donpenney: The images would potentially be relevant if a rollback occurs. The goal is to delete as few images automatically as possible, just to get under the threshold.

internal/imagemgmt/imagemgmt.go
donpenney: In which case this is an empty list. But it's possible for there to be podman containers, so why not check?

internal/imagemgmt/imagemgmt.go
donpenney: This is a backport.

internal/imagemgmt/imagemgmt.go
donpenney: Only Info and Error are supported, I believe

internal/imagemgmt/imagemgmt.go
donpenney: We're also deleting corresponding tags to ensure everything is cleaned up. Ensuring that the image doesn't seem to be in use avoids deleting tags for those images. It also avoids potential unnecessary failures and logs

PR #651, opened 2024-09-06
build env update for linux on arm

jc-rh:
# Background / Context
Several LCA developers are on M3 Macbooks running linux VMs now. This update enables building x64 images locally with the toolchain for arm64/aarch64 linux.

# Solution / Feature Overview

# Implementation Details

# Other Information

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,8 +25,8 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a
 
 #####################################################################################################
 # Build the operator image
-# note: update origin-cli-artifacts from `latest` to an appropriate OCP verison during release e.g `4.18`
-FROM quay.io/openshift/origin-cli-artifacts:latest AS origincli
+# note: update origin-cli-artifacts from `latest` to an appropriate OCP verison during release e.g `4.17`
+FROM --platform=linux/amd64 quay.io/openshift/origin-cli-artifacts:latest AS origincli
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 
 RUN if [[ ! -f /bin/nsenter ]]; then \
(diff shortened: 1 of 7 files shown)


pixelsoccupied:
build time went down from minutes to seconds! thanks for update!!

/lgtm

jc-rh:
/cc @rauhersu @fontivan 
This is not a real dockfile change. You can decide whether to take it, unless it's possible that d/s infra may use arm machines someday

fontivan:
> /cc @rauhersu @fontivan This is not a real dockfile change. You can decide whether to take it, unless it's possible that d/s infra may use arm machines someday

I like the idea, though I think it would be even better if we could specify a target architecture with `GOARCH` like I had done here for NTO: https://github.com/openshift/cluster-node-tuning-operator/pull/1135

jc-rh:
> > /cc @rauhersu @fontivan This is not a real dockfile change. You can decide whether to take it, unless it's possible that d/s infra may use arm machines someday
> 
> I like the idea, though I think it would be even better if we could specify a target architecture with `GOARCH` like I had done here for NTO: [openshift/cluster-node-tuning-operator#1135](https://github.com/openshift/cluster-node-tuning-operator/pull/1135)

Building arm images is beyond the scope of this PR. I am happy to make that change if I can verify the output. Does NTO get CI running on arm hardware now?

fontivan:
> > > /cc @rauhersu @fontivan This is not a real dockfile change. You can decide whether to take it, unless it's possible that d/s infra may use arm machines someday
> > 
> > 
> > I like the idea, though I think it would be even better if we could specify a target architecture with `GOARCH` like I had done here for NTO: [openshift/cluster-node-tuning-operator#1135](https://github.com/openshift/cluster-node-tuning-operator/pull/1135)
> 
> Building arm images is beyond the scope of this PR. I am happy to make that change if I can verify the output. Does NTO get CI running on arm hardware now?

I'm not sure what architecture the CI runs on but the change was very useful for testing since by using GOARCH you can build any arch from any arch (within the supported golang limits)

PR #654, opened 2024-09-15
MGMT-18910: Run once  apply static network in order not to fail in case installation service was restarted

tsorya:
[MGMT-18910](https://issues.redhat.com//browse/MGMT-18910): Run once  apply static network in order not to fail in case installation service was restarted

# Background / Context
In case of restart or failure of installation service we will try to reapply static network configuration though it was successfully applied already and it causes failure even though we can succeed on second run

# Solution / Feature Overview

# Implementation Details

# Other Information


eranco74 (commented):
/lgtm

eranco74 (commented):
/approve

PR #656, opened 2024-09-17
NO-ISSUE: cleanup etcd member update from lca as it was moved to recert

tsorya:
# Background / Context
We want all the etcd operations to be part of recert. After implementing https://github.com/rh-ecosystem-edge/recert/pull/161 in recert , cleaning up lca code

# Solution / Feature Overview

# Implementation Details

# Other Information

### DIFF
--- a/lca-cli/postpivot/postpivot.go
+++ b/lca-cli/postpivot/postpivot.go
@@ -38,7 +38,6 @@ import (
 
 	cp "github.com/otiai10/copy"
 	"github.com/sirupsen/logrus"
-	etcdClient "go.etcd.io/etcd/client/v3"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


lca-cli/postpivot/postpivot.go
```diff
@@ -38,7 +38,6 @@ import (
 
 	cp "github.com/otiai10/copy"
 	"github.com/sirupsen/logrus"
-	etcdClient "go.etcd.io/etcd/client/v3"
```
omertuc, line 41: If we move https://github.com/openshift-kni/lifecycle-agent/blob/1365cbe57129b75b964006fc17b4f34ed5a84f65/lca-cli/ops/ops.go#L123-L142 to recert as well we can get rid of the etcd lca dependency altogether, if we care about that

tsorya: Actually why not , though here was the idea to run it only on seed in order not to block installation. I believe we can add flag to recert and run it on demand in this case

tsorya: but this will be next pr

PR #657, opened 2024-09-17
MGMT-18928: take first image from .repoDigests

tsorya:
# Background / Context

  [MGMT-18928](https://issues.redhat.com//browse/MGMT-18928): take first image from .repoDigests
  While precaching images we pull all the images in parallel,
  some of them are pulled twice cause they have more than one repoDigest, though it is totally same image.
  Pulling them in parallel can cause podman to corrupt it though it finish ok

# Solution / Feature Overview

# Implementation Details

# Other Information


eranco74 (commented):
/lgtm

PR #659, opened 2024-09-18
Revert "MGMT-18928: take first image from .repoDigests"

tsorya:
This reverts commit 279f0c6a5514a796bceb5d8a7b5565ca16764459.
failed to reproduce issue locally, no need to change the logic if we can't reproduce it anymore

# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


omertuc:
It's still a good change IMO, why revert?

tsorya:
@omertuc to say the truth i don't think it will break something and is sort of good but till now it didn't bother and who knows what will be broken after all. 
Pretty difficult to validate it as we don't have CI for this specific case and QE runs it once in a version

tsorya:
and it looks like podman is handling it well by pulling layer only once and skipping all other tries

omertuc:
I don't have a strong opinion either way

PR #664, opened 2024-10-01
OCPBUGS-33605: Skip retry on unknown manifest and detect digests made up by CRIO ima…

jc-rh:
# Background / Context
Image service in CRIO can create phantom digests with this implementation:
https://github.com/openshift/cri-o/blob/1bcadab7e200e767394c7d19709eccba3462db9f/internal/storage/image.go#L216-L217
In the case of the bug reported, OCP uses quay.io digest for the same kube-rbac-proxy image while operators uses the multi-arch manifest list digest from registry.redhat.io. After pulling quay.io@sha:A and registry.redhat.io@sha:B, we get 4 digests on the image from a mix match of the repos and the sha values

This becomes issue in a disconnected environment for IBU, as the image mirroring only covers two digests from registry.redhat.io and one from quay.io. The manifest list sha from quay is not part of the single arch OCP release, hence not mirrored. IBU precaching requires all digests captured in the seed to be present.

# Solution / Feature Overview

- Double check failed image list at the end of precaching and consider it a success if they exist locally
- Skip retry on the not found case

# Implementation Details

# Other Information

### DIFF
--- a/internal/precache/workload/pullImages.go
+++ b/internal/precache/workload/pullImages.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
@@ -64,6 +65,15 @@ func podmanImgPull(image, authFile string) error {
 	return nil
 }
 
+func podmanImgExists(image string) bool {
+	args := []string{"image", "exists", image}
+	_, err := Executor.Execute("podman", args...)
+	if err != nil {
+		log.Errorf("failed podman image check with args %s: %v", args, err)
+	}
+	return err == nil
+}
+
 // pullImage attempts to pull an image via podman CLI
 func pullImage(image, authFile string, progress *precache.Progress) error {
 
@@ -74,7 +84,12 @@ func pullImage(image, authFile string, progress *precache.Progress) error {
 			log.Infof("Successfully pulled image: %s", image)
 			break
 		} else {
-			log.Infof("Attempt %d/%d: Failed to pull %s: %v", i+1, MaxRetries, image, err)
+			message := fmt.Sprintf("%v", err)
+			log.Infof("Attempt %d/%d: Failed to pull %s: %s", i+1, MaxRetries, image, message)
+			if strings.Contains(message, "manifest unknown") {
+				// no point to retry if we can reach the registry and the digest doesn't exist
+				break
+			}
 		}
 	}
 	// update precache progress tracker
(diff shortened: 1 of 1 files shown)


internal/precache/workload/pullImages.go
```diff
@@ -74,7 +84,11 @@ func pullImage(image, authFile string, progress *precache.Progress) error {
 			log.Infof("Successfully pulled image: %s", image)
 			break
 		} else {
-			log.Infof("Attempt %d/%d: Failed to pull %s: %v", i+1, MaxRetries, image, err)
+			message := fmt.Sprintf("%v", err)
+			log.Infof("Attempt %d/%d: Failed to pull %s: %s", i+1, MaxRetries, image, message)
+			if strings.Contains(message, "manifest unknown") {
+				break
```
donpenney, line 90: Might be good to have a comment here for a future dev, to explain why we're breaking vs a retry

internal/precache/workload/pullImages.go
```diff
@@ -158,9 +172,18 @@ func PullImages(precacheSpec []string, authFile string) *precache.Progress {
...
 		for _, image := range status.FailedPullList {
-			log.Info(image)
+			if strings.Contains(image, "@sha:") && podmanImgExists(image) {
+				log.Infof("%s, but found locally after downloading other images", image)
+				imagesFound = append(imagesFound, image)
+			} else {
+				log.Info(image)
+			}
+		}
+		if len(imagesFound) == status.Failed {
```
donpenney, line 185: I assume `status.Failed` is a count of the failed images? Not obvious from the field name, so a comment here would be good, too

internal/precache/workload/pullImages.go
```diff
@@ -158,9 +172,18 @@ func PullImages(precacheSpec []string, authFile string) *precache.Progress {
 func ValidatePrecache(status *precache.Progress, bestEffort bool) error {
 	// Check pre-caching execution status
 	if status.Failed != 0 {
+		var imagesFound []string
 		log.Info("Failed to pre-cache the following images:")
 		for _, image := range status.FailedPullList {
-			log.Info(image)
+			if strings.Contains(image, "@sha:") && podmanImgExists(image) {
```
donpenney, line 178: nit: We probably don't need the sha check here. That would be more useful if we were determining a failure action when pulling. Since we're just double-checking failed images, we could just do an existence check for any failed.

donpenney (commented):
/lgtm

internal/precache/workload/pullImages.go
```diff
@@ -64,6 +65,15 @@ func podmanImgPull(image, authFile string) error {
 	return nil
 }
 
+func podmanImgExists(image string) bool {
+	args := []string{"image exists", image}
```
donpenney, line 69:
```suggestion
	args := []string{"image", "exists", image}
```

javipolo (approved):

PR #670, opened 2024-10-09
OCPBUGS-42966: Readding node labels from target on post pivot

tsorya:
# Background / Context
User can add node labels and don't want to loose them after upgrade, this pr will save target node labels and they will be reapplied on post pivot.
In order not to reapply default labels we are filtering them from the list

https://issues.redhat.com/browse/OCPBUGS-42966

# Solution / Feature Overview

# Implementation Details

# Other Information

### DIFF
--- a/lca-cli/postpivot/postpivot_test.go
+++ b/lca-cli/postpivot/postpivot_test.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


lca-cli/postpivot/postpivot_test.go
```diff
@@ -876,3 +879,120 @@ func TestProxyAndProxyStatus(t *testing.T) {
...
+					Labels: map[string]string{
+						"node-role.kubernetes.io/master": "",
+					},
+				},
+			},
+			expectedError: false,
+		},
+		{
+			name:          "No labels provided",
+			nodeLabels:    map[string]string{},
```
donpenney, line 942: This would be equivalent to using a seed image generated without this update, thanks to the omitempty? Such that a seed image generated with a previous LCA can be handled by this change? The code looks like it handles this case, just wanted to confirm

tsorya: That what i meant in this test case

tsorya: let me think how i can validate it

tsorya: It is more like , post pivot should be able to run even if target doesn't set the field. 
Ill check manually if there is any issue if target sets the field but seed has old version of post-pivot

donpenney (approved):

Issue #676, opened 2024-10-17
need more fine-grained control of the creation of the "extraPartition" used by varlibcontainers


xphyr:
When applying "ImageBasedInstallationConfig" to a server, we need additional control over the creation of the varlibcontainers partition. 

Currently functionality only allows for defining an "ExtraPartitionStart" and then consume the remainder of the disk, or you can set "ExtraPartionStart" to a negative number and it works backwards from the end of the disk consuming a fixed amount of space, and leaving the rest for the root partition. 

The problem with this is that it does not allow for reserving any extra space on the device for other uses (such as LVM for local storage).

By adding an extra configuration option "ExtraPartitionSize" we can add additional controls around the creation of the varlibcontainers partition, and allow for additional partitions to be defined later via MachineConfigs.

NOTE: adding the additional partition will need to be configured with some form of post config script, or manually via the console until the MCO is updated to allow for changes to the partition configuration (https://github.com/openshift/machine-config-operator/blob/master/docs/FAQ.md#q-can-i-use-the-mco-to-re-partition-or-re-install).

I will be submitting a PR with these changes shortly.

PR #677, opened 2024-10-17
Add support for ExtraPartitionSize to address issue #676

xphyr:
# Background / Context

This PR addresses the inability to reserve some space on the install device for other partitions.  By allowing for additional partitions on the install disk we can create partitions for use by other processes, such as the LVM Operator.

This PR addresses issue #676 

# Solution / Feature Overview

I have added one extra configuration option "ExtraPartitionSize" which allows you to define the size of the varlibcontainers partition that is created. It is a optional configuration. If you do not define this option, the lca-cli works exactly the same as it did prior to this code change. This means that all existing configuration files will work with no changes.

# Implementation Details

This change takes advantage of the existing "sgdisk" code and appends the required additional value of ExtraPartitionSize to the "new" command, allowing us to specify a total size. By defaulting "ExtraPartitionSize" to "0" the any previous config files that did not specify this value will work without any modification. 

# Other Information

Given that MCO can not be used to configure additional partitions the lca-cli utility should be updated to support the creation of additional partitions as part of the image application process. In theory this can be handled by adding a "post.sh" with the additional commands, but it would be easier if it was built into the utility itself.

### DIFF
--- a/lca-cli/ibi-preparation/ibipreparation_test.go
+++ b/lca-cli/ibi-preparation/ibipreparation_test.go
@@ -2,10 +2,11 @@ package ibi_preparation
 
 import (
 	"fmt"
-	"github.com/openshift-kni/lifecycle-agent/api/ibiconfig"
 	"os"
 	"path"
 	"testing"
+
+	"github.com/openshift-kni/lifecycle-agent/api/ibiconfig"
 
 	preinstallUtils "github.com/rh-ecosystem-edge/preinstall-utils/pkg"
 	"github.com/sirupsen/logrus"
@@ -19,6 +20,7 @@ func TestDiskPreparation(t *testing.T) {
 	installationDisk := "/dev/sda"
 	extraPartitionLabel := "label"
 	extraPartitionStart := "-40"
+	extraPartitionEnd := "0"
 	extraPartitionNumber := uint(5)
 
 	testcases := []struct {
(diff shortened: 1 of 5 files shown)


lca-cli/ibi-preparation/ibipreparation_test.go
```diff
@@ -2,11 +2,12 @@ package ibi_preparation
 
 import (
 	"fmt"
-	"github.com/openshift-kni/lifecycle-agent/api/ibiconfig"
 	"os"
 	"path"
 	"testing"
 
+	"github.com/openshift-kni/lifecycle-agent/api/ibiconfig"
```
mresvanis, line 9: There seems to be an issue with this change in the CI, specifically I noticed the following [here](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-kni_lifecycle-agent/677/pull-ci-openshift-kni-lifecycle-agent-main-ci-job/1854191739665911808):

```shell
go vet ./...
# github.com/openshift-kni/lifecycle-agent/lca-cli/ibi-preparation
vet: lca-cli/ibi-preparation/ibipreparation_test.go:17:2: ibiconfig redeclared in this block
```

I didn't look into the cause of this, but I think if we moved this line to line 16 in the same file (i.e. before the other lifecycle-agent import) , it should resolved the CI issue. @xphyr could you perhaps give it a try?

tsorya:
we found another workaround for it @xphyr maybe we can close it?

xphyr:
> we found another workaround for it @xphyr maybe we can close it?

Yes, I think that makes sense. Thanks for addressing this request.

PR #678, opened 2024-10-22
MGMT-19151: Fix ingress crt not signed by target ingress CA key

mresvanis:
# Background / Context

The ingress certificate's Subject.CN that LCA provides to recert, i.e. `ingress-operator.key`, along with the ingress CA private key of the target cluster, in order for HTTP clients to keep trusting the ingress certificate is incorrect, i.e. it is not contained in the ingress CA certificate. Recert regenerates the ingress CA certificate and its private key, instead of using the provided ingress private key of the target cluster. The latter leads in HTTP clients requesting an API resource via ingress using the cluster's CA certificates in their TLS transport to fail because they can't verify the ingress certificate.

After an IBU we should have the ingress CA certificates properly regenerated and signed by the provided target cluster's ingress private key, using the correct CN in the recert config. That way we will leave the target cluster's ingress healthy instead of broken.

# Solution / Feature Overview

Instead of using the Subject.CN `ingress-operator.key`, we should use the [correct](https://github.com/openshift/cluster-ingress-operator/blob/release-4.17/pkg/operator/controller/certificate/ca.go#L63) one, which is formatted as `ingress-operator@<unix-timestamp>` and is contained in the `secrets/openshift-ingress-operator/router-ca` `tls.crt`. 

We should also replace the seed cluster's ingress CA Subject.CN with the target cluster's ingress CA Subject.CN, as it contains a unix timestamp.

# Implementation Details

# Other Information

You can verify this issue via the following steps:

1. Run an IBU

2. Download the cluster's CA bundle:
    ```sh
    oc config view --raw -o jsonpath='{.clusters[0].cluster.certificate-authority-data}' | base64 --decode > ca.crt
    ```

3. Download the ingress certificate:
    ```sh
    openssl s_client -connect oauth.apps.<cluster-name>.<base-domain>:443 -showcerts </dev/null </dev/null 2>/dev/null | awk '/BEGIN CERTIFICAT/,/END CERTIFICATE/ {print}' > ingress.crt
    ```

4. Try to verify the ingress certificate with the cluster's CA bundle: 
    ```sh
    openssl verify -CAfile ca.crt ingress.crt
    ```

### DIFF
--- a/api/seedreconfig/seedreconfig.go
+++ b/api/seedreconfig/seedreconfig.go
@@ -173,7 +173,15 @@ type ClientAuthCrypto struct {
 }
 
 type IngresssCrypto struct {
-	IngressCA PEM `json:"ingress_ca,omitempty"`
+	IngressCAPrivateKey PEM `json:"ingress_ca,omitempty"`
+
+	// IngressCertificateCN is the common name of the ingress CA used to sign the
+	// ingress certificate. The ingress CA is contained in the cluster's CA bundle,
+	// which is in turn contained in the kubeconfig CA. We need to reconfigure it
+	// during an IBU, in order for the ingress certificate to be regenerated and
+	// trusted with the target cluster's CA bundle.
+	// Its format is `ingress-operator@<unix-timestamp>`.
+	IngressCertificateCN string `json:"ingress_certificate_cn,omitempty"`
 }
 
 // Proxy defines the proxy settings for the cluster.
--- a/internal/recert/recert.go
+++ b/internal/recert/recert.go
@@ -137,14 +137,6 @@ func CreateRecertConfigFile(seedReconfig *seedreconfig.SeedReconfiguration, seed
 	seedFullDomain := fmt.Sprintf("%s.%s", seedClusterInfo.ClusterName, seedClusterInfo.BaseDomain)
 	clusterFullDomain := fmt.Sprintf("%s.%s", seedReconfig.ClusterName, seedReconfig.BaseDomain)
 	config.ExtendExpiration = true
-	config.CNSanReplaceRules = []string{
-		fmt.Sprintf("system:node:%s,system:node:%s", seedClusterInfo.SNOHostname, seedReconfig.Hostname),
-		fmt.Sprintf("%s,%s", seedClusterInfo.SNOHostname, seedReconfig.Hostname),
-		fmt.Sprintf("%s,%s", seedClusterInfo.NodeIP, seedReconfig.NodeIP),
-		fmt.Sprintf("api.%s,api.%s", seedFullDomain, clusterFullDomain),
-		fmt.Sprintf("api-int.%s,api-int.%s", seedFullDomain, clusterFullDomain),
-		fmt.Sprintf("*.apps.%s,*.apps.%s", seedFullDomain, clusterFullDomain),
-	}
 	config.KubeadminPasswordHash = seedReconfig.KubeadminPasswordHash
 	config.PullSecret = seedReconfig.PullSecret
 	if seedReconfig.ChronyConfig != "" {
@@ -160,7 +152,7 @@ func CreateRecertConfigFile(seedReconfig *seedreconfig.SeedReconfiguration, seed
 	}
 
 	if _, err := os.Stat(cryptoDir); err == nil {
-		ingressFile, ingressCN, err := getIngressCNAndFile(cryptoDir)
+		ingressKeyFile, err := getIngressKeyPath(cryptoDir)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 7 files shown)


api/seedreconfig/seedreconfig.go
```diff
@@ -174,6 +174,14 @@ type ClientAuthCrypto struct {
...
 type IngresssCrypto struct {
 	IngressCA PEM `json:"ingress_ca,omitempty"`
+
+	// IngressCertificateCN is the common name of the ingress CA used to sign the
+	// ingress certificate. The ingress CA is contained in the cluster's CA bundle,
+	// which is in turn contained in the kubeconfig CA. We need to reconfigure it
+	// during an IBU, in order for the ingress certificate to be regenerated and
+	// trusted with the target cluster's CA bundle.
+	// Its format is `ingress-operator@<unix-timestamp>`.
+	IngressCertificateCN string `json:"ingress_certificate_cn,omitempty"`
```
omertuc, line 184: Is this CN not already inside the PEM of the IngressCA field above?

internal/recert/recert.go
```diff
@@ -145,6 +145,11 @@ func CreateRecertConfigFile(seedReconfig *seedreconfig.SeedReconfiguration, seed
 		fmt.Sprintf("api-int.%s,api-int.%s", seedFullDomain, clusterFullDomain),
 		fmt.Sprintf("*.apps.%s,*.apps.%s", seedFullDomain, clusterFullDomain),
 	}
+	// check if there is an ingress CN provided for backwards compatibility
```
omertuc, line 148: Can we move the entire `config.CNSanReplaceRules = []string{` assignment to be *after* the `config.UseKeyRules = []string{` assignment to emphasize that internally in recert the former happens only after the latter? As in, the use-key-rule needs to use the name before the replacement, not after

internal/recert/recert.go
```diff
@@ -160,15 +165,15 @@ func CreateRecertConfigFile(seedReconfig *seedreconfig.SeedReconfiguration, seed
 	}
 
 	if _, err := os.Stat(cryptoDir); err == nil {
-		ingressFile, ingressCN, err := getIngressCNAndFile(cryptoDir)
+		ingressKeyFile, err := getIngressKeyFile(cryptoDir)
```
omertuc, line 168: If we're already renaming this function can you change "File" to "Path"

api/seedreconfig/seedreconfig.go
mresvanis: The IngressCA field is the private key, so no details in there, just crypto data. Its name is a bit misleading though.

utils/client_helper.go
```diff
@@ -137,6 +140,16 @@ func GetClusterInfo(ctx context.Context, client runtimeclient.Client) (*ClusterI
...
 	}
+
+	ingressOperatorCrt, err := GetSecretData(ctx, "router-ca", "openshift-ingress-operator", "tls.crt", client)
+	if err != nil {
+		return nil, err
+	}
+	ingressCN, err := GetCommonNameFromCertificate([]byte(ingressOperatorCrt))
+	if err != nil {
+		return nil, err
+	}
```
omertuc, line 151: since `ingressOperatorCrt` is only used for `GetCommonNameFromCertificate` we should probably put both calls under a single function to make it clear it's not used for `GetClusterInfo` in general and that they're connected

utils/crypto_dir.go
```diff
@@ -76,12 +76,21 @@ func SeedReconfigurationKubeconfigRetentionFromCluster(ctx context.Context, clie
...
 		return nil, err
 	}
+	ingressOperatorCrt, err := GetSecretData(ctx, "router-ca", "openshift-ingress-operator", "tls.crt", client)
+	if err != nil {
+		return nil, err
+	}
+	ingressCN, err := GetCommonNameFromCertificate([]byte(ingressOperatorCrt))
+	if err != nil {
+		return nil, err
+	}
```
omertuc, line 86: Same here

api/seedreconfig/seedreconfig.go
omertuc: That's very unfortunate, my bad.

api/seedreconfig/seedreconfig.go
```diff
@@ -174,6 +174,14 @@ type ClientAuthCrypto struct {
 
 type IngresssCrypto struct {
 	IngressCA PEM `json:"ingress_ca,omitempty"`
```
omertuc, line 176:
```suggestion
        // THIS FIELD WAS POORLY NAMED. It should be "IngressCAPrivateKey"
	IngressCA PEM `json:"ingress_ca,omitempty"`
```

utils/client_helper.go
mresvanis: This makes sense. Updated. PTAL

api/seedreconfig/seedreconfig.go
```diff
@@ -173,7 +173,16 @@ type ClientAuthCrypto struct {
 }
 
 type IngresssCrypto struct {
+	// THIS FIELD WAS POORLY NAMED. It should be "IngressCAPrivateKey" instead.
 	IngressCA PEM `json:"ingress_ca,omitempty"`
```
omertuc, line 177:
```suggestion
	IngressCAPrivateKey PEM `json:"ingress_ca,omitempty"`
```

Actually we can just do this and keep the `json` tag the same

PR #693, opened 2024-11-20
MGMT-19292: align /var/lib/containers partition label with var-lib-containers

tsorya:
[MGMT-19292](https://issues.redhat.com//browse/MGMT-19292): align /var/lib/containers partition label with var-lib-containers

In all ztp examples and docs we have var-lib-containers as label for containers partition, that's why we need to align with them

# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


eranco74 (commented):
/lgtm
/approve

PR #709, opened 2024-12-16
MGMT-19344: adding retry to static network apply as it it will fail in ibi case for some period of time and log and error

tsorya:
MGMT-19344: adding retry to static network apply as it it will fail in ibi case for some period of time and log and error

# Background / Context
In case we apply static network in ibi flow , it will fail for first time cause of disabled dbus of network manager, service will be restarted and everything will work though it creates confusing error , this change removes it

# Solution / Feature Overview

# Implementation Details

# Other Information


eranco74:
I think a better a fix is to update the systemd service with `After: whatever we require`

PR #727, opened 2025-01-23
MGMT-19545: Replace seed cluster mirror registries in IBI preparation step

mresvanis:
# Background / Context
During the installation step of IBI, after setting up the new stateroot we precache the seed cluster's container images. Although there is [support for mirror registries](https://github.com/openshift/installer/blob/main/pkg/asset/imagebased/image/registriesconf.go) when creating the installation ISO, we currently don't account for the seed image containing pull-specs from mirror registries (i.e. the seed image was generated in a cluster with mirror registries configured). This results in the IBI preparation/installation step to fail when trying to pull images from the release mirror registries of the seed cluster, instead of the target cluster's release mirror registries (or the original release registries if no mirror registries are configured on the target cluster).

We want to support different mirror registry configuration between the seed and the target cluster, in order to enable IBI with seed images generated from disconnected SNO clusters. 

# Solution / Feature Overview

During the IBI preparation step before precaching the seed cluster's images, we can replace the seed release mirror registry with the target release (mirror) registry. That's exactly what we do in IBU [here](https://github.com/openshift-kni/lifecycle-agent/blob/main/controllers/prep_handlers.go#L377-L398).

# Implementation Details

In order to replace the seed release mirror registry with the target release (mirror) registry, we need the seed's release registry and the target's release registry. We can read the first one from the `manifest.json` that's included in the seed image.We can read the second one from the new IBI preparation configuration field `ReleaseRegistry`, which we will add during the creation of the installation ISO.

# Other Information


internal/prep/prep.go
```diff
@@ -123,7 +123,7 @@ func getDeploymentFromDeploymentID(deploymentID string) (string, error) {
 }
 
 func SetupStateroot(log logr.Logger, ops ops.Ops, ostreeClient ostreeclient.IClient,
-	rpmOstreeClient rpmostreeclient.IClient, seedImage, expectedVersion, imageListFile string, ibi bool) error {
+	rpmOstreeClient rpmostreeclient.IClient, seedImage, expectedVersion, tmpPath string, ibi bool) error {
```
mresvanis, line 126: We should probably rename `tmpPath` to something indicating that this is for IBI. @eranco74 WDYT?

eranco74: Why not keep `imageListFile` ?

api/ibiconfig/ibiconfig.go
```diff
@@ -79,6 +79,9 @@ type IBIPrepareConfig struct {
 	// Provide the device id like /dev/by-id/ata-xxxxx
 	InstallationDisk string `json:"installationDisk"`
 
+	// ReleaseRegistry is the container image registry that hosts the OpenShift release-image content.
+	ReleaseRegistry string `json:"releaseRegistry,omitempty"`
```
eranco74, line 83: Please mention that this registry is used during pre-caching (while performing the preparation/installation step) and can be different than the registry to be used at the edge site deployment

internal/prep/prep.go
mresvanis: Because we in addition to the `imageListFile`, we would also need the `manifests.json`: 

https://github.com/openshift-kni/lifecycle-agent/pull/727/files#diff-8200c824bb53076107616f57fa19be0cb56e51514a600a70c65f0c8a94116c8fR249-R256

eranco74: Or just move ibi_preparation.ibi_preparation to common and use it in both places

lca-cli/ibi-preparation/ibipreparation.go
```diff
@@ -81,15 +92,28 @@ func (i *IBIPrepare) Run() error {
...
+		return fmt.Errorf("failed to read seed info: %s, %w", common.PathOutsideChroot(seedInfoFile), err)
+	}
+	i.log.Info("Collected seed info for precache: ", fmt.Sprintf("%+v", seedInfo))
+
+	i.log.Info("Checking whether to override seed registry")
+	shouldOverrideSeedRegistry, err := i.shouldOverrideSeedRegistry(seedInfo.MirrorRegistryConfigured, seedInfo.ReleaseRegistry)
+	if err != nil {
+		return fmt.Errorf("failed to check ShouldOverrideSeedRegistry %w", err)
+	}
+	i.log.Info("Should override seed registry", "shouldOverrideSeedRegistry", shouldOverrideSeedRegistry, "releaseRegistry", i.config.ReleaseRegistry)
```
eranco74, line 113: Can we share the code with https://github.com/openshift-kni/lifecycle-agent/blob/main/controllers/prep_handlers.go#L377-L398 ?

lca-cli/ibi-preparation/ibipreparation.go
```diff
@@ -211,3 +235,45 @@ func (i *IBIPrepare) cleanupRhcosSysroot() error {
 	}
 	return nil
 }
+
+func (i *IBIPrepare) shouldOverrideSeedRegistry(seedMirrorRegistryConfigured bool, seedReleaseRegistry string) (bool, error) {
```
eranco74, line 239: Its already implemented at https://github.com/openshift-kni/lifecycle-agent/blob/e8a8907961f834444fc215a9ea19e5015179f0c5/utils/client_helper.go#L414

lca-cli/ibi-preparation/ibipreparation.go
```diff
@@ -211,3 +235,45 @@ func (i *IBIPrepare) cleanupRhcosSysroot() error {
...
+	//   - and seed SNO has mirror registry configured then we should try to override the registry
+	//   - and seed SNO has no mirror registry configured then we shouldn't try to override the registry
+	if !targetMirrorRegistryConfigured {
+		return seedMirrorRegistryConfigured, nil
+	}
+
+	return !lo.Contains(targetMirroredRegistries, seedReleaseRegistry), nil
+}
+
+func mirrorRegistrySourceRegistries() ([]string, error) {
```
eranco74, line 259: https://github.com/openshift-kni/lifecycle-agent/blob/e8a8907961f834444fc215a9ea19e5015179f0c5/utils/client_helper.go#L291

lca-cli/ibi-preparation/ibipreparation.go
mresvanis: That's what I first thought, but the main difference is where we fetch the mirror registry locations from.

- IBU - we're fetching the mirror registry locations from the [cluster ](https://github.com/openshift-kni/lifecycle-agent/blob/main/utils/client_helper.go#L291-L314)
- IBI (what's implemented here) we're fetching the mirror registry locations from `/etc/containers/registries.conf` - https://github.com/openshift-kni/lifecycle-agent/pull/727/files#diff-2a78eaecaa85ed29707e26c8abe068e48b486f8c3f2b816e7f9135a22f890441R260

Perhaps I could try refactoring the code to share some logic. WDYT?

api/ibiconfig/ibiconfig.go
mresvanis: Great point, I've updated the doc to reflect this.

lca-cli/ibi-preparation/ibipreparation.go
eranco74: If it's a lot of trouble in order to share 3 lines of code you can resolve this comment

lca-cli/ibi-preparation/ibipreparation.go
mresvanis: Same as https://github.com/openshift-kni/lifecycle-agent/pull/727/files#r1935308620.

lca-cli/ibi-preparation/ibipreparation.go
mresvanis: Same as https://github.com/openshift-kni/lifecycle-agent/pull/727/files#r1935308620.

internal/prep/prep.go
```diff
@@ -246,10 +246,14 @@ func SetupStateroot(log logr.Logger, ops ops.Ops, ostreeClient ostreeclient.ICli
 		return fmt.Errorf("failed to process etc.deletions: %w", err)
 	}
 
-	if err := common.CopyOutsideChroot(filepath.Join(mountpoint, "containers.list"), imageListFile); err != nil {
+	if err := common.CopyOutsideChroot(filepath.Join(mountpoint, "containers.list"), filepath.Join(tmpPath, "containers.list")); err != nil {
```
tsorya, line 249: don't we have var in common for containers.list? Lets add it in this case

lca-cli/ibi-preparation/ibipreparation.go
tsorya: Hmm, still will be nice to try to share it, seems like it can get targetMirroredRegistries from outside and in this case logic is pretty the same or i missed something?

lca-cli/ibi-preparation/ibipreparation.go
tsorya: Probably we can change name to show that it is read from registry conf file?

mresvanis: Absolutely, I've updated the function to make it clearer that it reads the registry conf file. PTAL

lca-cli/ibi-preparation/ibipreparation.go
mresvanis: I think both your suggestions look a lot better, it's now less confusing :+1:  

@tsorya I don't think you're missing anything, the difference is where we get the target's mirror registry sources from, either via the cluster's kube API or the `registries.conf` file.

lca-cli/ibi-preparation/ibipreparation.go
mresvanis: I think it's just 4 lines of codes we're sharing, but it makes sense to share them anyway https://github.com/openshift-kni/lifecycle-agent/pull/727/files#diff-89d0ea845042a03c3156e8b2e5d8e1980f1b97379fd4a26d02d5c271cc18195fR414-R422.

Thanks for the suggestion :pray:

internal/prep/prep.go
mresvanis: I have updated the changes to better handle common file paths between IBU/IBI and removed this arg from this function. PTAL

internal/prep/prep.go
mresvanis: Took your suggestion and did a small refactoring into how we manage common filepaths between IBu/IBI. PTAL :pray:

PR #733, opened 2025-01-29
NO-ISSUE: retry image pull  in all cases

tsorya:
NO-ISSUE: in some registries (brew for example) we can get a redirect to another registry that doesn't have this specific image manifest and we will get manifest-unknown error and will not retry to pull it, removing this pretty small optimisation will allow us to retry in this case and make our code more error resilient

# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


omertuc:
Why would retrying suddenly work? Is the redirect temporary?

PR #734, opened 2025-01-31
OCPBUGS-49692: fix seedgen container being stuck as stopped

mresvanis:
# Background / Context

On podman 5.3.1 and systemd 252-49 (OpenShift 4.19) the seed generation container, run with systemd-run via podman, while exiting with code 0 (successfully) it's stuck with status `stopped` instead of `exited`. This means that [conmon](https://github.com/containers/conmon), run as a separate process, did not cleanup after the container exited. 

In OCP 4.19 although seed generation completes successfully, the container `exited` status is never reached, leading to the SeedGenerator CR being stuck in a failed state. Podman 5.3 improved the container exit wait time. Since podman now exits a lot faster and systemd kills all processes of the unit's control group (e.g. conmon) right after the main process of the unit (i.e. podman) exits, conmon doesn't have a chance to cleanup after the container exits. The container never reaches status `exited`.

# Solution / Feature Overview

We can set systemd's [ExitType](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExitType=) property to `cgroup`, to indicate that it should wait for all processes in the unit's control group to exit.

/cc @tsorya

### DIFF
--- a/controllers/seedgen_controller.go
+++ b/controllers/seedgen_controller.go
@@ -493,6 +493,22 @@ func (r *SeedGeneratorReconciler) launchImager(seedgen *seedgenv1.SeedGenerator)
 	systemdRunOpts := []string{
 		"--collect",
 		"--wait",
+		// Conmon is the process that podman uses to cleanup after a container exits.
+		// After cleanup the container status moves from stopped to exited. The following:
+		//
+		//   - Podman 5.3 improved the container exit wait time.
+		//   - Systemd by default kills all processes in the control group after the
+		//     main process (i.e. podman) exits.
+		//
+		// lead to the container never reaching the exited status (i.e. no container
+		// cleanup is executed), because systemd kills the conmon process before it
+		// has a chance to cleanup as podman exits faster than before.
+		//
+		// This flag indicates to systemd that it should wait for all the processes
+		// in the control group to exit, instead of waiting only on the main process.
+		//
+		// https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExitType=
+		"--property", "ExitType=cgroup",
 		"--unit", "lca-generate-seed-image",
 		// Ensure the proxy environment variables of the LCA container are
 		// passed through systemd-run to the podman process (which will pass it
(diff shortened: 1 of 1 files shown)


controllers/seedgen_controller.go
```diff
@@ -521,7 +522,7 @@ func (r *SeedGeneratorReconciler) checkImagerStatus() error {
 		State ContainerState `json:"State"`
 	}
 
-	expectedStatus := "exited"
+	expectedStatuses := []string{"exited", "stopped"}
```
tsorya, line 525: @donpenney do you remember why we check exited only?
Why we can't create some file with status and read it?

eranco74: Because it's the standard way to check if a container you ran completed successfully.
In  any case, we can just ignore the status and check the exit code (as we already do in the next line).

tsorya: Seems like it doesn't really matter which of this status to change

mresvanis: IIUC from what I've gathered so far, these 2 statuses from our point of view are the same. 

Specifically, this is what I managed to find out about `stopped` vs `exited`:

> stopped is general only a short state, that will be transitioned to exited very quickly, it is the state between we notice the container process died/exited and we cleanup all container resources (i.e. mounts, network, etc...). So in general I think exited is expected there.

I'm also in discussion with podman folks to see if we can track this different behavior between podman version 5.2.2 and 5.3.1 (since when using the same systemd version, the results are the same).

mresvanis:
/hold

It seems that systemd kills podman's 5.3.1 cleanup process faster than it did in the previous version of podman 5.2.2, so perhaps we should let the cleanup process complete (e.g. by using sleep). Investigating potential solutions (by also discussing with the podman team).

controllers/seedgen_controller.go
mresvanis: Updated the PR's [description](https://github.com/openshift-kni/lifecycle-agent/pull/734#issue-2823737928) to reflect the root cause and the solution proposed.

controllers/seedgen_controller.go
```diff
@@ -493,6 +493,22 @@ func (r *SeedGeneratorReconciler) launchImager(seedgen *seedgenv1.SeedGenerator)
...
+		//
+		// lead to the container never reaching the exited status (i.e. no container
+		// cleanup is executed), because systemd kills the conmon process before it
+		// has a chance to cleanup as podman exits faster than before.
+		//
+		// This flag indicates to systemd that it should wait for all the processes
+		// in the control group to exit, instead of waiting only on the main process.
+		//
+		// https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExitType=
+		"--property=ExitType", "cgroup",
```
tsorya, line 511: Great finding, thank you

PR #762, opened 2025-03-25
Pre-Konflux Alignment for LCA Repo

fontivan:
# Background / Context
Some project alignment is necessary alongside Konflux onboarding

https://issues.redhat.com/browse/CNF-15329

# Solution / Feature Overview
- Update project to golang 1.23.0
- Prep Dockerfile for Konflux build
- Update golangci-lint to a version compatible with golang 1.23
- Ignore new warnings exposed by linter update as they are out of scope for this

# Implementation Details
These changes are necessary to support Konflux onboarding: https://github.com/openshift-kni/lifecycle-agent/pull/763

# Other Information
n/a

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@
 # Build arguments
 ARG BUILDER_IMAGE=quay.io/projectquay/golang:1.23
 ARG RUNTIME_IMAGE=registry.access.redhat.com/ubi9-minimal:9.4
+# note: update origin-cli-artifacts from `latest` to an appropriate OCP version during release e.g `4.18`
 ARG ORIGIN_CLI_IMAGE=quay.io/openshift/origin-cli-artifacts:latest
 
 # Assume x86 unless otherwise specified
@@ -37,7 +38,6 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GO111MODULE=on go build -mod=vendo
 
 #####################################################################################################
 # Build the operator image
-# note: update origin-cli-artifacts from `latest` to an appropriate OCP verison during release e.g `4.18`
 FROM ${ORIGIN_CLI_IMAGE} AS origincli
 FROM ${RUNTIME_IMAGE}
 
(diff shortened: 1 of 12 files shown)


fontivan:
Depends on https://github.com/openshift/release/pull/63230 due to golangci-lint adding a cache directory in a place that is not writable in our default CI configuration

Dockerfile
```diff
@@ -1,6 +1,21 @@
...
+ARG BUILDER_IMAGE=quay.io/projectquay/golang:1.23
+ARG RUNTIME_IMAGE=registry.access.redhat.com/ubi9-minimal:9.4
+# note: update origin-cli-artifacts from `latest` to an appropriate OCP version during release e.g `4.18`
+ARG ORIGIN_CLI_IMAGE=quay.io/openshift/origin-cli-artifacts:latest
+
+# Assume x86 unless otherwise specified
+ARG GOARCH="amd64"
+
 # Build the binaries
-FROM registry.access.redhat.com/ubi9/go-toolset:1.22.5-1730550521 as builder
```
jc-rh, line 3: @mresvanis I can't recall why we started using the go-toolset image. It seems slow on releasing newer golang versions. Can you double check if Steven's change is good?

fontivan: I found the change here in an older PR: https://github.com/openshift-kni/lifecycle-agent/pull/65

and yeah it seems like go1.23 is not available yet in ubi.

Also note that for Konflux we will be using the openshift golang builder but that will be in a separate PR: https://github.com/openshift-kni/lifecycle-agent/pull/763/files#diff-e5b5fa6deab8f9cda1e5725b06bd8d766482779861b02c95a47669a7ff8a844e

Dockerfile
```diff
@@ -18,14 +33,19 @@ COPY main main
...
+FROM ${RUNTIME_IMAGE}
+
+# Pass GOARCH into runtime
+ARG GOARCH
+
+# Explicitly set the working directory
+WORKDIR /
 
 RUN if [[ ! -f /bin/nsenter ]]; then \
         microdnf -y install util-linux-core && \
```
rauhersu, line 51: todo: we need to start working on the hermetic build + subscription manager to install those packages on the konflux side here.

fontivan: I think I will do that in a separate PR once I go through the details

PR #797, opened 2025-04-25
[release-4.19] Update Konflux target branch to release-4.19

fontivan:
- Update the branch Konflux is checking against

### DIFF
--- a/.tekton/lifecycle-agent-operator-bundle-4-19-push.yaml
+++ b/.tekton/lifecycle-agent-operator-bundle-4-19-push.yaml
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "false"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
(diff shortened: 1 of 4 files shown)


.tekton/lifecycle-agent-operator-bundle-4-19-push.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "false"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && target_branch == "main" && files.all != ["konflux_clusterserviceversion_overlay.data"]
```
rauhersu, line 11: talm does not have the same config here for the bundle. I am assuming we do want to trigger a bundle build if that '.data' file is modified. Is that correctly coded up ? 

talm config: https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/release-4.19/.tekton/topology-aware-lifecycle-manager-bundle-4-19-push.yaml#L11

fontivan: we do want the bundle to build when the data file is modified, good catch

PR #798, opened 2025-04-25
Update Konflux overlay to mimic oran implementation

fontivan:
- Cleans up the repo a bit by keeping konflux things contained in .konflux/
- Uses more readable yaml for defining the overlay data
- Update tekton pipelines and renovate configuration to match new paths


.konflux/overlay/overlay.bash
```diff
@@ -0,0 +1,356 @@
...
+    local name_version="$name.v$version"
+    local manager="lifecycle-agent-operator"
+    local skip_range=">=4.9.0 <4.20.0"
+    local replaces="lifecycle-agent-operator.v4.20.0"
+    # min_kube_version should match ocp
+    export min_kube_version="1.32.0"
+
+    yq e -i ".metadata.annotations[\"containerImage\"] = \"${IMAGE_TO_TARGET[$MANAGER_KEY]}\"" $ARG_CSV_FILE
+    yq e -i ".spec.displayName = \"$display_name\"" $ARG_CSV_FILE
+    yq e -i ".spec.description = \"$description\""  $ARG_CSV_FILE
```
yanirq, line 284: I did not see this getting injected in the old render template method. is this needed here ? do we want to keep the csv format as close as we can to the previous ones ?

.konflux/overlay/overlay.bash
```diff
@@ -0,0 +1,356 @@
...
+overlay_release()
+{
+    echo "Overlaying relase..."
+
+    local display_name="lifecycle-agent-operator"
+    local description="lifecycle-agent-operator"
+    local version="4.20.0"
+    local name="lifecycle-agent-operator"
+    local name_version="$name.v$version"
+    local manager="lifecycle-agent-operator"
```
yanirq, line 276: were is `manager` used here?

fontivan: I could not identify it being used anywhere, but I wanted to keep the script as in-tact as possible versus the oran implementation from @rauhersu

.konflux/overlay/overlay.bash
fontivan: I believe this is needed to set various fields that we currently set in the existing midstream and was missing from the original konflux upstream implementation

jc-rh (approved):

PR #817, opened 2025-04-28
Update Konflux References

fontivan:
- Include commit from Konflux bot
- Account for new mandatory parameters
- Update to latest task versions for good measure

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,9 +11,6 @@
                 "\\.yaml$",
                 "\\.yml$"
             ],
-            "includePaths": [
-                ".konflux/**"
-            ],
             "matchStrings": [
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -32,6 +29,9 @@
             "matchManagers": [
                 "custom.regex"
             ],
+            "includePaths": [
```
rauhersu, line 32: Is this the most up to date renovate config we have ? I believe I still have the old (changes to the left) present on ims.

fontivan: Yes, I have this exact configuration in TALM now and it is working again. The config was invalid with the includePaths in the top section

PR #818, opened 2025-04-28
[release-4.19] Konflux Backports from main

fontivan:
Includes 3 PRs:
- https://github.com/openshift-kni/lifecycle-agent/pull/817
- https://github.com/openshift-kni/lifecycle-agent/pull/798
- https://github.com/openshift-kni/lifecycle-agent/pull/816


.tekton/lifecycle-agent-4-19-push.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/cancel-in-progress: "false"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && target_branch == "release-4.19" && files.all != ["konflux_clusterserviceversion_overlay.data"]
+    pipelinesascode.tekton.dev/on-cel-expression: event == "push" && target_branch == "main" && files.all != [".konflux/overlay/pin_images.in.yaml"]
```
rauhersu, line 11: I believe this PR is intended for a release-4.19 branch, do we need to replace 'main' by it ? 

We might want to add further logic here: we don't want lca manager to rebuild if we change the 'map_images.in.yaml' or 'overlay.bash' -> basically, everything under /overlay. Ditto for all operators.

fontivan: ahh no this is a good catch, I copy pasted the entire line instead of just the file section

I am going to evaluate a better on-cel-expression for the repos soon, but separately

rauhersu: Let us know if you change expression. I also started a discussion related to this, but from the bundle side instead: https://redhat-internal.slack.com/archives/C07UVNGQD2T/p1746011417019879

PR #866, opened 2025-05-06
Update to more acceptable openshift cli image

fontivan:
# Background / Context
quay.io is not a good source for oc cli as it is not an official build

conforma policy does not allow us to use quay.io/openshift/* as a source image

# Solution / Feature Overview
- Use the production oc cli image to fetch the oc binary

# Implementation Details
- when 4.19 is released we should come back and update the tag here from v4.18 to v4.19

# Other Information


jc-rh (approved):

fontivan:
/override ci/prow/ibu-e2e-flow

ci is still broken (https://github.com/openshift/release/pull/64480)

PR #902, opened 2025-05-12
[release-4.18] Backport Konflux changes to 4.18

fontivan:
# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


.konflux/overlay/map_images.in.yaml
```diff
@@ -0,0 +1,10 @@
+# Manager
+- key: manager
+  staging: registry.stage.redhat.io/openshift4/lifecycle-agent-rhel9-operator
+  production: registry.redhat.io/openshift4/lifecycle-agent-rhel9-operator
+#
+# Recert
```
rauhersu, line 6: Consider kube-rbac-proxy for mapping + pinning in every < 4.19 release (I believe lca has been included in >=4.16 or >=4.14 btw, we will need lesser y-streams than talm, for example),

fontivan: Interesting I did not see a conforma failure on LCA for the proxy but I definitely forgot about it here, was thinking it was only on TALM for some reason

rauhersu: And one more reminder: we need to decide what to do with the overlay for recert yet.

fontivan: I have an idea for that but I will do it later once the LCA backports are completed

PR #1504, opened 2025-07-11
Enable multi-arch building for lifecycle-agent

fontivan:
- Include arm64 as a build target in Konflux
- Update rpm lock configuration to include both architectures packages
- Update Dockerfile/Makefile to fix cross compiling to work both in and outside of Konflux

### DIFF
--- a/.konflux/lock-runtime/redhat.repo
+++ b/.konflux/lock-runtime/redhat.repo
@@ -9,3474 +9,156 @@
 # a "yum repolist" to refresh available repos
 #
 
-[satellite-client-6-for-rhel-9-x86_64-eus-debug-rpms]
-name = Red Hat Satellite Client 6 for RHEL 9 x86_64 - Extended Update Support (Debug RPMs)
-baseurl = https://cdn.redhat.com/content/eus/rhel9/9.4/x86_64/sat-client/6/debug
+[rhel-9-for-$basearch-appstream-rhui-source-rpms]
+name = Red Hat Enterprise Linux 9 for $basearch - AppStream from RHUI (Source RPMs)
+baseurl = https://cdn.redhat.com/content/dist/rhel9/rhui/9.4/$basearch/appstream/source/SRPMS
 enabled = 0
 gpgcheck = 1
 gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
 sslverify = 1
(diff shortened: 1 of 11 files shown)


.konflux/lock-runtime/redhat.repo
```diff
@@ -9,8808 +9,9228 @@
 # a "yum repolist" to refresh available repos
 #
 
-[satellite-client-6-for-rhel-9-x86_64-eus-debug-rpms]
-name = Red Hat Satellite Client 6 for RHEL 9 x86_64 - Extended Update Support (Debug RPMs)
```
shajmakh, line 13: having such a large repo file is redundant and makes it hard to review the important changes. let's please include only the enabled repos and further filter the used repos according to the generated lockfile.

fontivan: This is straight copied out of the container used to generate the lockfile, is there any simple way to filter it down?

.konflux/lock-runtime/rpms.in.yaml
```diff
@@ -10,6 +10,7 @@ packages:
 
 arches:
   - x86_64
+  - aarch64
```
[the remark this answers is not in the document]
fontivan: Have updated the commit message

Dockerfile
```diff
@@ -38,21 +32,18 @@ COPY utils utils
 # Otherwise compile normally
 RUN if [[ "${KONFLUX}" == "true" ]]; then \
         echo "Compiling with fips" && \
-        GOEXPERIMENT=strictfipsruntime CGO_ENABLED=1 GOOS=linux GOARCH=${GOARCH} GO111MODULE=on go build -mod=vendor -tags strictfipsruntime -o build/manager main/main.go && \
-        GOEXPERIMENT=strictfipsruntime CGO_ENABLED=1 GOOS=linux GOARCH=${GOARCH} GO111MODULE=on go build -mod=vendor -tags strictfipsruntime -a -o build/lca-cli main/lca-cli/main.go; \
+        GOEXPERIMENT=strictfipsruntime CGO_ENABLED=1 GOOS=linux GO111MODULE=on go build -mod=vendor -tags strictfipsruntime -o build/manager main/main.go && \
```
rauhersu, line 35: Should we compile the binaries for arm64 as well somehow (toolchain option) ?  Don't know how we match the respective konflux builds with the binaries included for the archs we are considering.

PR #1611, opened 2025-07-16
Update project to use new shared script library

fontivan:
# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


.konflux/Dockerfile.catalog
```diff
@@ -29,8 +27,9 @@ RUN echo "run secrets" && ls -lra /run/secrets/ && echo "docker dir" && ls -lra
 
 ENV REGISTRY_AUTH_FILE=$HOME/.docker/config.json
 # The fbc build is not hermetic, so make will download yq and opm
-RUN make konflux-generate-catalog-production && \
-    rm $HOME/.docker/config.json
+RUN export PATH="/app/bin:/app/telco5g-konflux/scripts/catalog/bin:$PATH" && \
```
rauhersu, line 30: Apparently, there is content under '/app/bin' but I can't figure out where that dir has been populated.

fontivan: It is still required since the tools aren't installed in the default $PATH inside the container

rauhersu: Great ! It looks way better now

.konflux/overlay/release.in.yaml
```diff
@@ -9,9 +9,9 @@ variables:
 
     You can find additional guidance in the upstream repository: https://github.com/openshift-kni/lifecycle-agent
   display_name: "Lifecycle Agent"
-  manager_version: "lifecycle-agent.v4.20.0"
+  manager_version: "lifecycle-agent.v4.19.0"
```
rauhersu, line 12: Do we need this 4.19 in main ?

fontivan: no, this commit is on the wrong branch

PR #2990, opened 2025-10-24
CNF-18836: Reconfigure repo to allow mintmaker rpm updates to work

fontivan:
- We can't use complicated arg logic for the containers or rpm configuration for this to work
- The Dockerfile and rpm lockfile also need to be in the same directory
- The manual script update is still supported, and will run in a temporary `lock-runtime` folder with copied in files

AI-attribution: AIA,Entirely human-created,v1.0
For more information on AI attribution statements, see: https://aiattribution.github.io/


Dockerfile
```diff
@@ -1,14 +1,5 @@
-#####################################################################################################
```
shajmakh: IIUC you intend to split the konflux dockerfile. Unless the non-konflux dockerfile is used somewhere else (prow CI maybe?) then it can be deleted to avoid unnecessary maintenance for an additional file.

shajmakh (commented):
Thanks, I left some comments.
from what I observe, I see this repo is maintaining a single lockfile, why not just have the input yaml and lockfile in the root (adjust the manual script to tolerate the change) and let the minmaker file updates for it since the beginning? is it a matter of repo layout?

Dockerfile
fontivan: The developers use it as part of their workflow for testing changes

Makefile
```diff
@@ -107,8 +107,8 @@ SHELL = /usr/bin/env GOFLAGS=$(GOFLAGS) bash -o pipefail
 .SHELLFLAGS = -ec
 
 # RHEL9_RELEASE defines the RHEL9 release version to update the rpm lock file for the runtime
-# This is automatically extracted from the RUNTIME_IMAGE in `.konflux/container_build_args.conf`
-RHEL9_RELEASE ?= $(shell awk -F'=' '/^RUNTIME_IMAGE=/ {split($$2, parts, /[:|@]/); print parts[2]}' $(PROJECT_DIR)/.konflux/container_build_args.conf)
+# This is automatically extracted from the runtime image in `.konflux/Dockerfile`
+RHEL9_RELEASE ?= $(shell awk '/^FROM.*rhel.*as runtime-image/ {split($$2, parts, /[:|@]/); print parts[2]}' $(PROJECT_DIR)/.konflux/Dockerfile)
```
fontivan: Looking forward to CNF-20367, this will need adjustment

PR #4058, opened 2025-12-29
MGMT-21789: Fix Idle condition handling and IBU gating in ipc

danmanor:
This PR fixes the following issues:

1. IPC won't start when IBU is nor Idle - so it won't interfere. (In the future we will do the same for IBU)
2. Fixes IPC Idle conditions handling - Stops running Idle handler when the CR is already Idle

### DIFF
--- a/controllers/ibu_controller.go
+++ b/controllers/ibu_controller.go
@@ -54,7 +54,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	ibuv1 "github.com/openshift-kni/lifecycle-agent/api/imagebasedupgrade/v1"
-	ipcv1 "github.com/openshift-kni/lifecycle-agent/api/ipconfig/v1"
 	lcautils "github.com/openshift-kni/lifecycle-agent/utils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -182,21 +181,6 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	// Make sure "next stage" list is initialized
 	ibu.Status.ValidNextStages = getValidNextStageList(ibu, isAfterPivot)
-
-	if ibu.Spec.Stage != ibuv1.Stages.Idle && isTransitionRequested(ibu) {
-		ipc := &ipcv1.IPConfig{}
-		if getErr := r.NoncachedClient.Get(ctx, client.ObjectKey{Name: common.IPConfigName}, ipc); getErr != nil {
-			r.Log.Error(getErr, "failed to get IPConfig for gating")
-			return requeueWithShortInterval(), nil
-		}
-
-		if !(ipc.Spec.Stage == ipcv1.IPStages.Idle && utils.IsIdleConditionTrue(ipc.Status.Conditions)) {
-			r.Log.Info(
-				"Blocking IBU start: IPConfig not Idle",
-				"ipcStage", ipc.Spec.Stage)
-			return doNotRequeue(), nil
-		}
-	}
 
 	if isTransitionRequested(ibu) {
 		if validateStageTransition(ibu, isAfterPivot) {
--- a/controllers/ipc_controller.go
+++ b/controllers/ipc_controller.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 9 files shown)


controllers/ipc_controller.go
```diff
@@ -177,6 +183,62 @@ func (r *IPConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+func (r *IPConfigReconciler) gateIPConfigByIBU(
+	ctx context.Context,
+	ipc *ipcv1.IPConfig,
+) (ctrl.Result, error) {
+	ibu := &ibuv1.ImageBasedUpgrade{}
+	if getErr := r.NoncachedClient.Get(ctx, client.ObjectKey{Name: controllerutils.IBUName}, ibu); getErr != nil {
+		if !k8serrors.IsNotFound(getErr) {
+			return requeueWithError(fmt.Errorf("failed to get ImageBasedUpgrade for gating: %w", getErr))
+		}
+		return requeueImmediately(), nil
```
eranco74: Why `requeueImmediately`? if the CR doesn't exist, this will create a hotloop.

controllers/ipc_controller.go
```diff
@@ -177,6 +183,62 @@ func (r *IPConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+
+	if len(ibu.Status.Conditions) == 0 {
+		controllerutils.SetIPStatusBlocked(ipc, "Blocked by gating: IBU is not initialized")
+		if err := r.Client.Status().Update(ctx, ipc); err != nil {
+			return requeueWithError(fmt.Errorf("failed to update ipconfig status: %w", err))
+		}
+		return requeueImmediately(), nil
+	}
+
+	if ibu.Spec.Stage == ibuv1.Stages.Idle &&
```
eranco74: Add some logs here, e.g.
IBU is idle + IBU idle status condition
Removing the blocked condition

controllers/ipc_controller.go
```diff
@@ -177,6 +183,62 @@ func (r *IPConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+		if inProgressCondition != nil &&
+			inProgressCondition.Reason == string(controllerutils.ConditionReasons.Blocked) {
+			meta.RemoveStatusCondition(&ipc.Status.Conditions, inProgressCondition.Type)
+		}
+		if err := r.Client.Status().Update(ctx, ipc); err != nil {
+			return requeueWithError(fmt.Errorf("failed to update ipconfig status: %w", err))
+		}
+		return doNotRequeue(), nil
+	}
+
```
eranco74: If it's already blocked, we can return early and avoid the update

controllers/ibu_controller.go
```diff
@@ -183,21 +182,6 @@ func (r *ImageBasedUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	// Make sure "next stage" list is initialized
 	ibu.Status.ValidNextStages = getValidNextStageList(ibu, isAfterPivot)
 
-	if ibu.Spec.Stage != ibuv1.Stages.Idle && isTransitionRequested(ibu) {
```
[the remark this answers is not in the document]
danmanor: Because it doesn't really solve the issue, and I don't want to impact IBU at the moment. We discussed changing IBU code later to be blocked when ipc is on when we have time to test it.

controllers/ipc_controller.go
```diff
@@ -177,6 +183,62 @@ func (r *IPConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+// Notice: this is not mutual exclusion. For delivery reasons, IBU can still start while IPC is
+// not Idle. We should add mutual exclusion between IPC and IBU controllers to avoid this
+// in the future.
+func (r *IPConfigReconciler) gateIPConfigByIBU(
+	ctx context.Context,
+	ipc *ipcv1.IPConfig,
+) (ctrl.Result, error) {
+	ibu := &ibuv1.ImageBasedUpgrade{}
+	if getErr := r.NoncachedClient.Get(ctx, client.ObjectKey{Name: controllerutils.IBUName}, ibu); getErr != nil {
+		if !k8serrors.IsNotFound(getErr) {
```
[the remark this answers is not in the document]
danmanor: Unless the user deletes it, then it will be recreated soon after

controllers/ipc_controller.go
danmanor: Because the only scenario I know without IBU is very short when the user deletes it but then it should be recreated soon after

PR #4061, opened 2025-12-31
MGMT-21789: Implement new IP Configuration flow

danmanor:
IP Configuration is a new flow introduced to provide solution for 2 main
use cases -

  1. **Rehoming** - Changing SNO and its node IP address as well other basic networking features in a planned connected manner.
  2. **Disaster** Recovery - Changing SNO and its node IP address as well as other basic networking features when unplanned disaster occur and uplink is gone.

**Notice** - This PR depends on - https://github.com/rh-ecosystem-edge/recert/pull/975


danmanor:
Merging due to lack delivery urgency alongside lack of reviewers. Also this code was reviewed in `main`.

PR #4126, opened 2026-01-07
chore(deps): update lifecycle-agent version to 4.20.2

by2waysprojects:
# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information

### DIFF
--- a/.konflux/overlay/release.in.yaml
+++ b/.konflux/overlay/release.in.yaml
@@ -9,9 +9,9 @@ variables:
 
       You can find additional guidance in the upstream repository: https://github.com/openshift-kni/lifecycle-agent
   display_name: "Lifecycle Agent"
-  manager_version: "lifecycle-agent.v4.20.1"
+  manager_version: "lifecycle-agent.v4.20.2"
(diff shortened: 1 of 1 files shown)


.konflux/overlay/release.in.yaml
```diff
@@ -9,9 +9,9 @@ variables:
 
       You can find additional guidance in the upstream repository: https://github.com/openshift-kni/lifecycle-agent
   display_name: "Lifecycle Agent"
-  manager_version: "lifecycle-agent.v4.20.1"
+  manager_version: "lifecycle-agent.v4.20.2"
   # min_kube_version should match ocp
   # https://access.redhat.com/solutions/4870701
   min_kube_version: "1.32.0"
   recert_image: PLACEHOLDER_RECERT_IMAGE
-  version: "4.20.1"
```
fontivan: The catalog template input file will also need to be updated, with both the final 4.20.1 build and the placeholder for 4.20.2

fontivan: this will be done in a second, follow up pr

PR #4197, opened 2026-01-14
CNF-21209: Add missing parameters to rpm lockfile

fontivan:
- We need to pass the appropriate ssl parameters for Mintmaker to work correctly
- Also update local make target to replace them as necessary
- These changes were made in other repos when this was implemented, but seem to be missing from the original implementation here: https://github.com/openshift-kni/lifecycle-agent/pull/3289/changes


fontivan:
cc. @by2waysprojects @rauhersu

fontivan:
Close in favor of #4200

PR #6241, opened 2026-04-28
Fix/remove automerge renovate

by2waysprojects:
# Background / Context

# Solution / Feature Overview

# Implementation Details

# Other Information


renovate.json
```diff
@@ -88,8 +88,6 @@
         "addLabels": [
             "allow-automatic-merge"
         ],
-        "autoApprove": true,
-        "automerge": true,
```
fontivan, line 92: I think you actually want the ones on lines 60/61, though I think we should probably remove all the instances of `autoapprove` and `automerge` from the file to be on the safe side

by2waysprojects: ok, I removed autoApproved and automerge from whole file,

PR #7786, opened 2026-07-10
CNF-22168: Implement tiered health checks for PostPivot

jmontesi:
# Background / Context

During an IBU (Image-Based Upgrade) on SNO clusters, the LCA's `PostPivot()` function runs after the node reboots into the new stateroot. It performs health checks to validate the cluster is stable, then triggers OADP restore to bring back workload configurations, and finally declares the upgrade complete.

Currently, all health checks run as a single block before OADP restore begins. This means checks that are not strictly required for a successful restore (Node readiness, CSVs, CSRs) still gate the restore operation, delaying workload recovery unnecessarily.

The health check set currently includes: ClusterOperators, MachineConfigPools, Node readiness, ClusterServiceVersions, CertificateSigningRequests, ClusterVersion, SRIOV, DPA, and BSL.

[CNF-22168](https://redhat.atlassian.net/browse/CNF-22168) requires implementing a two-tier health check model where only the checks that are genuinely required for a successful OADP restore gate the restore operation, while remaining checks run after restore completes before declaring the upgrade finished.

The mandatory pre-restore checks (as specified in the Jira) are: Control plane (ClusterOperators), MCP, SRIOV, and OADP readiness. Everything else can be deferred.

Additionally, [CNF-22164](https://redhat.atlassian.net/browse/CNF-22164) requests removing the `IsClusterVersionReady` check entirely, as it is redundant with `AreClusterOperatorsReady`. Empirical testing (9 clean runs) confirmed it has no measurable impact on any downtime metric.

# Solution / Feature Overview

Split the health checks into two tiers:

1. `MandatoryHealthChecks` (pre-restore gate): ClusterOperators, MachineConfigPools, SRIOV (gated on COs), and OADP (DPA + BSL, gated on COs). These must pass before OADP restore can begin.

2. `DeferredHealthChecks` (post-restore): Node readiness, CSVs, and CSRs. These run only after OADP restore completes successfully, before declaring the upgrade finished.

The existing `HealthChecks()` function is preserved as a wrapper calling both (used by PrePivot and other callers that need the full set in a single call).

`IsClusterVersionReady` is removed entirely ([CNF-22164](https://redhat.atlassian.net/browse/CNF-22164)).

# Implementation Details

- `internal/healthcheck/healthcheck.go`: Added `MandatoryHealthChecks()` and `DeferredHealthChecks()` as new exported functions. Removed `IsClusterVersionReady()` and its RBAC marker. `HealthChecks()` now calls Mandatory then Deferred sequentially.

- `controllers/upgrade_handlers.go`: Added `CheckMandatoryHealth` and `CheckDeferredHealth` package-level vars (for test mockability). Modified `PostPivot()` to call `CheckMandatoryHealth` before restore, and `CheckDeferredHealth` after restore completes successfully.

- SRIOV and OADP checks within `MandatoryHealthChecks` are gated on ClusterOperators passing first (same pattern as the original code, which gated them on both COs and CSVs). Since CSVs are now deferred, gating on COs alone is sufficient -- if the SRIOV/OADP operators are healthy as ClusterOperators, their CSVs will be in a ready state.

# Other Information

- Lab tested with 15 clean upgrade cycles (16 total, 1 excluded for hardware boot hang). Results show a statistically significant workload downtime improvement of -1m21s (p<0.001) compared to the custom baseline.
- No check runs in both tiers. The split is clean with no overlap.

# Checklist

- [x] I performed a rough self-review of my changes
- [x] I explained non-trivial motivation for my code using code-comments
- [x] I made sure my code passes linting, tests, and builds correctly
- [x] I have ran the code and made sure it works as intended, and doesn't introduce any obvious regressions
- [x] I have not committed any irrelevant changes
- [x] I added tests

### DIFF
--- a/internal/healthcheck/healthcheck.go
+++ b/internal/healthcheck/healthcheck.go
@@ -26,7 +26,6 @@ import (
 // +kubebuilder:rbac:groups=machineconfiguration.openshift.io,resources=machineconfigpools,verbs=list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators,verbs=list;watch
 // +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=list;watch
-// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=list;watch
 // +kubebuilder:rbac:groups=config.openshift.io,resources=infrastructures,verbs=get;list;watch
 // +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
 // +kubebuilder:rbac:groups=sriovnetwork.openshift.io,resources=sriovnetworknodestates,verbs=get;list;watch
@@ -41,11 +40,12 @@ const (
 	SriovNetworkNodeStateNotPresentMsg = "no SriovNetworkNodeStates present"
 )
 
-func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
+// MandatoryHealthChecks runs the checks required before OADP restore can begin:
+// ClusterOperators, MachineConfigPools, SRIOV, and OADP (DPA + BSL).
+func MandatoryHealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
 	var failures []string
 
 	clusterOperatorsReady := false
-	clusterServiceVersionsReady := false
 
 	if err := AreClusterOperatorsReady(ctx, c, l); err != nil {
 		l.Info("co health check failure", "error", err.Error())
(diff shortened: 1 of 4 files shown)


internal/healthcheck/healthcheck.go
```diff
@@ -41,11 +40,12 @@ const (
...
-func HealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
+// MandatoryHealthChecks runs the checks required before OADP restore can begin:
+// ClusterOperators, MachineConfigPools, SRIOV, and OADP (DPA + BSL).
+func MandatoryHealthChecks(ctx context.Context, c client.Reader, l logr.Logger) error {
 	var failures []string
 
 	clusterOperatorsReady := false
-	clusterServiceVersionsReady := false
 
 	if err := AreClusterOperatorsReady(ctx, c, l); err != nil {
```
jc-rh, line 50: @browsell Do we need all cluster operators to be ready or a subset?

jc-rh (approved):

